PR #113, opened 2019-02-04
Feature/h5store context manager reentry

csadorf:
Enable reentry of the H5Store context manager.

## Description

This essentially enables:
```python
with job.data:
    with job.data:
```
## Motivation and Context

This is especially useful if you use the store in different functions and just want to make sure that the file is open within that specific context without caring about whether the file is already opened in the outer context.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

This pull request is based on

## Checklist:

- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/misc/contributor-flow/CONTRIBUTING.md#code-style) of this project.

If necessary:

### DIFF
--- a/signac/core/h5store.py
+++ b/signac/core/h5store.py
@@ -55,9 +55,8 @@ def _requires_tables():
 logger = logging.getLogger(__name__)
 
 
-class ClosedH5StoreError(RuntimeError):
-    "Raised when trying to access a closed group."
-    pass
+class H5StoreClosedError(RuntimeError):
+    "Raised when trying to access a closed store."
 
 
 class H5StoreAlreadyOpenError(OSError):
@@ -307,7 +306,7 @@ class H5Store(MutableMapping):
     @property
     def file(self):
         if self._file is None:
-            raise ClosedH5StoreError(self._filename)
+            raise H5StoreClosedError(self._filename)
         else:
             return self._file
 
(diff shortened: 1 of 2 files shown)


signac/core/h5store.py
```diff
@@ -247,7 +247,11 @@ def __del__(self):
         self.close()
 
     def __enter__(self):
-        self.open()
+        try:
+            self.open()
+        except OSError as error:
+            if 'is already open' not in str(error):
```
bdice: I'm not sure that I like tying this implementation to the specific error text. I would imagine that non-English versions of Python might not raise an error with this string. (Versuch es bitte?) I tried to find what raises this error (h5py? CPython? Something else?) but couldn't locate the source. Is there another way?

csadorf: Unfortunately the h5py package overall relies on generic RuntimeErrors for almost all error conditions. I don't know of a better way to limit the scope here. I don't thing that localized versions of the h5py package or the HDF5 library exist and therefore do not anticipate this to be a problem.

csadorf:
I just noticed the tests failing on PyPy2 and 3. I'll fix that now.

csadorf:
@bdice I've addressed all comments and fixed all issues.

PR #115, opened 2019-02-05
Feature/h5store support storing of hdf5 data sets and mappings

csadorf:
Explicitly support `h5py` datasets and mappings as valid types for the `H5Store` class.

## Description

This patch special-case handles both h5py data sets and mappings during assignments and equality checks. An exception is raised when a user attempts to set an existing group to the same key.

## Motivation and Context

Changes introduced here are aimed at increasing the versatility and robustness of the `H5Store`.
The `H5Store` should look-and-feel as much as *reasonable* like a `dict` without compromising the inherent advantages of the underlying file format.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

This pull request is based on

## Checklist:

If necessary:

### DIFF
--- a/signac/core/h5store.py
+++ b/signac/core/h5store.py
@@ -67,6 +67,9 @@ def _h5set(file, grp, key, value, path=None):
     import numpy    # h5py depends on numpy, so this is safe.
     path = path + '/' + key if path else key
 
+    # Guard against assigning a group to itself, e.g., `h5s[key] = h5s[key]`,
+    # where h5s[key] is a mapping. This is necessary, because the original
+    # mapping would be deleted prior to assignment.
     if isinstance(value, H5Group) and key in grp:
         if grp[key] == value._group:
             return  # Groups are identical, do nothing.
--- a/signac/core/h5store.py
+++ b/signac/core/h5store.py
@@ -58,8 +58,12 @@ logger = logging.getLogger(__name__)
 def _h5set(file, grp, key, value, path=None):
     """Set a key in an h5py container, recursively converting Mappings to h5py
     groups and transparently handling None."""
+    import h5py
     import numpy    # h5py depends on numpy, so this is safe.
     path = path + '/' + key if path else key
+
+    if isinstance(value, Mapping) and path in file and file[path] == value:
+        raise RuntimeError("Cannot assign mapping to the same key!")
 
     # Delete any existing data
     if key in grp:
@@ -82,6 +86,9 @@ def _h5set(file, grp, key, value, path=None):
     # NumPy types
     elif type(value).__module__ == numpy.__name__:
         grp[key] = value
+
+    elif isinstance(value, h5py._hl.dataset.Dataset):
+        grp[key] = value[()]    # Create a copy, not a hard link!
 
     # Other types
     else:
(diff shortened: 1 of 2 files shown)


csadorf:
I'm going to address the failing unit tests now.

csadorf:
Ok, this appears to be ready for review now.

signac/core/h5store.py
```diff
@@ -55,12 +55,22 @@ def _requires_tables():
 logger = logging.getLogger(__name__)
 
 
+class ClosedFileError(RuntimeError):
```
bdice: Since this is being implemented specifically for `H5Store` and not for signac as a whole, would the name `ClosedH5StoreError` be more appropriate?

signac/core/h5store.py
```diff
@@ -55,12 +55,22 @@ def _requires_tables():
...
+
 def _h5set(file, grp, key, value, path=None):
     """Set a key in an h5py container, recursively converting Mappings to h5py
     groups and transparently handling None."""
+    import h5py
     import numpy    # h5py depends on numpy, so this is safe.
     path = path + '/' + key if path else key
 
+    if isinstance(value, H5Group) and key in grp:
+        if grp[key] == value._group:
```
bdice: Is this check for performance or correctness? If it's correct without it, I'm not sure if we should do this check.

signac/core/h5store.py
```diff
@@ -83,6 +93,9 @@ def _h5set(file, grp, key, value, path=None):
     elif type(value).__module__ == numpy.__name__:
         grp[key] = value
 
+    elif isinstance(value, h5py._hl.dataset.Dataset):
```
bdice: It would be good to have a corresponding comment above this line, for consistency with other types.

signac/core/h5store.py
csadorf: This is required for correctness. Otherwise you delete the key and then try to assign the value that was just deleted to the same key.

signac/core/h5store.py
```diff
@@ -55,12 +55,22 @@ def _requires_tables():
...
+    "Raised when trying to access a closed group."
+    pass
+
+
 def _h5set(file, grp, key, value, path=None):
     """Set a key in an h5py container, recursively converting Mappings to h5py
     groups and transparently handling None."""
+    import h5py
     import numpy    # h5py depends on numpy, so this is safe.
     path = path + '/' + key if path else key
```
mikemhenry: I just had a quick question, I've seen this construct pop up a few times in the h5store code, is this path refer to a file path? Should `os.path` be used to ensure the right direction of slash is used?

signac/core/h5store.py
bdice: A comment might be helpful. e.g.
`# Guards against setting a key to itself, which can result in deletion before assignment`

signac/core/h5store.py
bdice: @mikemhenry It's not a platform-dependent implementation in my understanding, the HDF5 standard uses a forward slash for separations. Source: https://support.hdfgroup.org/HDF5/Tutor/crtgrpar.html

mikemhenry: Cool, I've only used HDF5 with pandas and I couldn't quite work out if this was a literal file system path or a path in the HDF5 api..

PR #116, opened 2019-02-05
Expand the H5Store doc-string documentation.

csadorf:
Expand the H5Store docstrings.

### DIFF
--- a/signac/core/h5store.py
+++ b/signac/core/h5store.py
@@ -237,8 +237,20 @@ class H5Group(MutableMapping):
 
 
 class H5Store(MutableMapping):
-    """An HDF5 backend for storing arbitrary array-like and dictionary-like data.
-    Both attribute-based and index-based operations are supported.
+    """An HDF5-backed container for storing array-like and dictionary-like data.
+
+    The H5Store API implements the MutableMapping API and behaves similar
+    to a dict, but all data is persistently stored in the associated HDF5-file.
+
+    Supported types include:
+
+      * built-in types (int, float, str, bool, NoneType, array)
+      * numpy arrays
+      * pandas data frames (requires pandas and pytables),
+
+    as well as mappings of values of these types.
+
+    Values can be accessed as attributes (h5s.foo) or via key index (h5s['foo']).
 
     Example:
 
@@ -246,8 +258,9 @@ class H5Store(MutableMapping):
 
         with H5Store('file.h5') as h5s:
             h5s['foo'] = 'bar'
+            assert 'foo' in h5s
             assert h5s.foo == 'bar'
-
+            assert h5s['foo'] == 'bar'
     """
     __slots__ = ['_filename', '_file']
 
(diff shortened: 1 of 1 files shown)


signac/core/h5store.py
```diff
@@ -215,17 +215,30 @@ def __eq__(self, other):
 
 
 class H5Store(MutableMapping):
-    """An HDF5 backend for storing arbitrary array-like and dictionary-like data.
-    Both attribute-based and index-based operations are supported.
+    """An HDF5-backed container for storing array-like and dictionary-like data.
+
+    The H5Store API implements the MutableMapping API and behaves similar
```
bdice: We include the Python docs in signac via intersphinx, so it might be nice to have an intersphinx link to `collections.abc.MutableMapping` here.

signac/core/h5store.py
```diff
@@ -215,17 +215,30 @@ def __eq__(self, other):
...
+
+    Supported types include:
+
+      * built-in types (int, float, str, bool, NoneType, array)
+      * numpy arrays
+      * pandas data frames (requires pandas and pytables),
+
+    as well as mappings of values of these types.
+
+    Values can be accessed as attributes (h5s.foo) or via key index (h5s['foo']).
```
bdice: Use code font for code.

csadorf:
I am tentatively building the documentation here: https://docs.signac.io/projects/core/en/misc-h5store-docs/
However, I realized that like `JSONDict`, the API for this module is not online, because it is not considered core API.

I suggest we revisit a possible restructuring of the root namespace as part of a separate discussion.

csadorf:
> LGTM, pending minor suggestions.

@bdice I've addressed all suggestions.

signac/core/h5store.py
```diff
@@ -215,17 +215,30 @@ def __eq__(self, other):
...
+    the associated HDF5-file on disk.
+
+    Supported types include:
+
+      * built-in types (int, float, str, bool, NoneType, array)
+      * numpy arrays
+      * pandas data frames (requires pandas and pytables),
+
+    as well as mappings of values of these types. Values can be accessed as
+    attributes (`h5s.foo`) or via key index (`h5s['foo']`).
```
bdice: @csadorf Did you render the "code font" change? I think that using one backtick doesn't work, I believe you must use one of the following:
```
:code:`h5s.foo`
```
or
```
``h5s.foo``
```

signac/core/h5store.py
```diff
@@ -215,17 +215,30 @@ def __eq__(self, other):
 
 
 class H5Store(MutableMapping):
-    """An HDF5 backend for storing arbitrary array-like and dictionary-like data.
-    Both attribute-based and index-based operations are supported.
+    """An HDF5-backed container for storing array-like and dictionary-like data.
+
+    The H5Store API is a :class:`collections.abc.MutableMapping` and therefore
+    behaves similar to a :class:`dict`, but all data is stored persisently in
```
bdice: persisently ➡️ persistently

PR #117, opened 2019-02-08
Expand and forward constructor arguments of H5Store to h5py.File.

csadorf:
## Description

 * Provide users with more options to control how the underlying h5py.File
   is constructed.
 * Implemented unit tests that test opening in read- and write-only mode.

## Motivation and Context
It was not possible to control the open mode prior to this patch.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

This pull request is based on

## Checklist:

### DIFF
--- a/signac/core/h5store.py
+++ b/signac/core/h5store.py
@@ -263,10 +263,9 @@ class H5Store(MutableMapping):
             assert h5s['foo'] == 'bar'
 
     :param filename:
-        The filename of the underlying HDF5-file. This file will be create if it
-        does not exist yet with the standard open mode ('a').
+        The filename of the underlying HDF5-file.
     :param mode:
-        The file open mode to use, defaults to 'a'.
+        The file open mode to use. Defaults to 'a' (append).
     :param kwargs:
         Additional keyword arguments to be forward to the h5py.File constructor
         See documentation for :class:`h5py.File` for more information.
--- a/signac/core/h5store.py
+++ b/signac/core/h5store.py
@@ -311,7 +311,7 @@ class H5Store(MutableMapping):
         import h5py
         self._thread_lock.acquire()
         try:
-            self._file = h5py.File(self._filename, mode=self.mode, **self.kwargs)
+            self._file = h5py.File(self._filename, mode=self.mode, **self._kwargs)
         except:  # noqa We need to release under **all** circumstances upon error!
             self._thread_lock.release()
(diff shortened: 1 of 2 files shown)


signac/core/h5store.py
```diff
@@ -261,16 +261,29 @@ class H5Store(MutableMapping):
             assert 'foo' in h5s
             assert h5s.foo == 'bar'
             assert h5s['foo'] == 'bar'
+
+    :param filename:
+        The filename of the underlying HDF5-file. This file will be create if it
```
bdice: create ➡️ created

signac/core/h5store.py
```diff
@@ -261,16 +261,29 @@ class H5Store(MutableMapping):
...
+        does not exist yet with the standard open mode ('a').
```
bdice: Is this part about the standard open mode redundant? Or am I missing something?

signac/core/h5store.py
```diff
@@ -298,7 +311,7 @@ def open(self):
         import h5py
         self._thread_lock.acquire()
         try:
-            self._file = h5py.File(self._filename)
+            self._file = h5py.File(self._filename, mode=self.mode, **self._kwargs)
```
bdice: Should this be `self._mode`? edit: nevermind, I see the property, but why do we expose that property in the first place?

signac/core/h5store.py
csadorf: I wanted to make sure that users understand they don't *have* to create the file beforehand. The comment that 'a' is default may be redundant, but is that a problem?

signac/core/h5store.py
csadorf: It should be `_mode`. The mode is a property of the `H5Store`, why not expose it?

bdice: I see it is exposed for file objects. We can expose it. https://docs.python.org/3/library/io.html#io.FileIO.mode

signac/core/h5store.py
bdice: To me, this implies that if you specify `mode='r'` and the file does not yet exist, it overrides your provided `'r'` and uses `'a'` instead. I'm not sure if that is true.

csadorf: Got it, will revise.

csadorf:
@bdice Is this better?

PR #118, opened 2019-02-09
Add get_job function

### LINKED ISSUE #108: Implement a `signac.get_job()` function
**[Original report](https://bitbucket.org/glotzer/signac/issue/108) by Carl Simon Adorf (Bitbucket: [csadorf](https://bitbucket.org/csadorf), GitHub: [csadorf](https://github.com/csadorf)).**

The idea is to be able to "get" a job from the local directory. Use cases for example when opening a `signac shell` from within a signac workspace directory.

The main challenge is that the current data model does not associate a job workspace directory with one specific project.

See also related pull request: https://bitbucket.org/glotzer/signac/pull-requests/62

bdice:
## Description
When the current working directory is a signac job directory, it would be nice to be able to quickly identify that job in a similar way to `signac.get_project()`. This implementation

## Motivation and Context
Resolves #108. Sometimes I navigate the filesystem to a job directory (e.g. `cd workspace/abc123`) so I can access files directly. However, it is tough to open a Python REPL and get to this job without copy-pasting the job id.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

This pull request is based on

## Checklist:

If necessary:


signac/contrib/project.py
```diff
@@ -1524,6 +1524,36 @@ def get_project(cls, root=None):
                     os.getcwd() if root is None else os.path.abspath(root)))
         return cls(config=config)
 
+    @classmethod
+    def get_job(cls, root=None):
+        """From a job workspace directory or subdirectory, get the job.
```
csadorf: I think this needs to be clarified.

signac/contrib/project.py
```diff
@@ -1524,6 +1524,36 @@ def get_project(cls, root=None):
...
+        :param root: The job root directory.
+            If no root directory is given, the current working directory is
+            assumed to be the job directory.
+        :type root: str
+        :returns: The job handle.
+        :raises LookupError: If this job cannot be found."""
+        if root is None:
+            root = os.getcwd()
+        results = re.findall(JOB_ID_REGEX, root)
+        if len(results) > 0:
```
csadorf: Shouldn't the length of `results` be exactly one? What does it mean if it is larger?

signac/contrib/project.py
```diff
@@ -1827,3 +1857,16 @@ def get_project(root=None):
     :rtype: :py:class:`~.Project`
     :raises LookupError: If no project configuration can be found."""
     return Project.get_project(root=root)
+
+
+def get_job(root=None):
+    """From a job workspace directory, locate the closest parent project
```
csadorf: I think this is supposed to express "when the current working directory is a workspace directory" or so, but I don't think this is clear with the current description.

signac/contrib/project.py
```diff
@@ -1827,3 +1857,16 @@ def get_project(root=None):
...
+    and return this job.
+
```
csadorf: Can we maybe add a simple example on how this function is to be used here?

tests/test_project.py
```diff
@@ -1827,6 +1827,58 @@ def check_root(root=None):
...
+            with self.assertRaises(KeyError):
+                project.get_job()
+            with self.assertRaises(KeyError):
+                signac.get_job()
+
+    def test_get_job_subdir_of_project(self):
+        # Test case: The job workspace dir is a sub-directory of the project root dir.
+        pass
+
+    def test_get_job_not_subdir_of_project(self):
```
csadorf: I think I understand this unit test that still needs to implemented, but I forgot what the one above is supposed to test. @bdice Do you remember?

signac/contrib/project.py
bdice: We're doing a regex match on the full path. If you're in a directory like `my_project/workspace/JOB_ID/my_subproject/workspace/SUBJOB_ID`, then we take `SUBJOB_ID` (the last match). Then this will be a job belonging to the project `my_subproject` found from that same directory.

signac/contrib/project.py
bdice: I changed this.

signac/contrib/project.py
bdice: I changed this.

signac/contrib/project.py
bdice: Added an example.

tests/test_project.py
bdice: Two (three?) tests are unimplemented:
```python
    def test_get_job_subdir_of_project(self):
        # Test case: The job workspace dir is a sub-directory of the project root dir.
        pass
```
This description doesn't make sense to me. I would like to test the case where we have a project with a job, and the current working directory is a subdirectory of the job workspace.
```python
    def test_get_job_not_subdir_of_project(self):
        # Test case: The job workspace dir is not a sub-directory of project root dir.
        # Test case: The job is not actually part of the project that we identified for this job.
        pass
```
I think the first case would be like making a symlink to `my_project/workspace/abc123` and placing that symlink outside of the `my_project` directory. I expect this to fail.

I think the second case is similar to the first, but placing that symlink in an unrelated project `other_project` directory. This should also fail unless that job symlink is placed in the workspace, `other_project/workspace`, in which case it is actually "in the `other_project`." Not sure what to test there.

If you want me to implement these, ping me.

bdice:
@csadorf Ready for additional review.

csadorf:
I'm going to rebase this branch on develop.

csadorf:
> I'm going to rebase this branch on develop.

Just kidding, I merged it.

bdice:
@csadorf I have implemented three additional tests for various cases and added more comments to the logic of `get_job`. This addresses all of the cases I can think of. I'm ready to call this done when tests pass.

signac/contrib/project.py
```diff
@@ -1531,6 +1531,49 @@ def get_project(cls, root=None, search=True):
                 "Unable to determine project id for path '{}'.".format(os.path.abspath(root)))
         return cls(config=config)
 
+    @classmethod
+    def get_job(cls, root=None):
+        """Find a Job in or above the current working directory (or
+        provided path).
```
csadorf: The first paragraph in a doc-string should only span one line: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

signac/contrib/project.py
```diff
@@ -1531,6 +1531,49 @@ def get_project(cls, root=None, search=True):
...
+        results = list(re.finditer(JOB_ID_REGEX, root))
+        if len(results) == 0:
+            raise LookupError("Could not find a job id in path '{}'.".format(root))
+        match = results[-1]
+        job_id = match.group(0)
+        job_root = root[:match.end()]
+
+        # Find a project *above* the root directory (if the provided root
+        # contains a project, we must search from its parent)
+        project = cls.get_project(os.path.join(job_root, os.pardir))
```
csadorf: I don't quite understand the comment here. It seems like we are *always* searching the parent directory, but the comment implies that we do this only sometimes.

signac/contrib/project.py
```diff
@@ -1531,6 +1531,49 @@ def get_project(cls, root=None, search=True):
...
+
+        # Search for the matched job id in the found project
+        try:
+            job = project.open_job(id=job_id)
+        except KeyError:
+            raise
```
csadorf: Since we are just re-raising we can completely omit the above two lines.

signac/contrib/project.py
```diff
@@ -1531,6 +1531,49 @@ def get_project(cls, root=None, search=True):
...
+        else:
+            if job in project:
+                return job
+            else:
+                raise LookupError("Unable to find job id {} in project '{}'.".format(
```
csadorf: I think we can just rely on the exception raised by the `open_job()` function in line 1567, can we not?

signac/contrib/project.py
```diff
@@ -1836,6 +1879,30 @@ def get_project(root=None, search=True):
...
+    :rtype: :py:class:`~.Project`
     :raises LookupError: If no project configuration can be found.
     """
     return Project.get_project(root=root, search=search)
+
+
+def get_job(root=None):
+    """When the current working directory or provided root directory is a job
+    workspace directory (or subdirectory thereof), locate the closest parent
+    project and return this job.
```
csadorf: Same comment as above, the first paragraph in a multi-line doc string should span only one line.

tests/test_project.py
```diff
@@ -1827,19 +1841,90 @@ def check_root(root=None):
...
+    def test_get_job_subdir(self):
+        # Test case: Get a job from a sub-directory of the job workspace dir.
+        root = self._tmp_dir.name
+        project = signac.init_project(name='testproject', root=root)
+        job = project.open_job({'a': 1})
+        job.init()
+        with job:
+            os.mkdir('test_subdir')
+            self.assertEqual(project.get_job('test_subdir'), job)
+            self.assertEqual(signac.get_job('test_subdir'), job)
```
csadorf: Can we also test this without switching into the job>

tests/test_project.py
```diff
@@ -1827,19 +1841,90 @@ def check_root(root=None):
...
+        root = self._tmp_dir.name
+        project = signac.init_project(name='testproject', root=root)
+        job = project.open_job({'a': 1})
+        job.init()
+        with job:
+            nestedproject = signac.init_project('nestedproject')
+            nestedproject.open_job({'b': 2}).init()
+            os.mkdir('test_subdir')
+            self.assertEqual(project.get_job('test_subdir'), job)
+            self.assertEqual(signac.get_job('test_subdir'), job)
```
csadorf: Same as above, this should be tested also without switching into the job context.

bdice:
@csadorf I addressed your comments. Should be ready to go with your approval.

bdice:
@csadorf There are H5Store / numpy related failures on pypy2/3. Doesn't seem related to this PR.

csadorf:
> @csadorf There are H5Store / numpy related failures on pypy2/3. Doesn't seem related to this PR.

Not sure how this is possible... 😕  I'll have a look.

csadorf:
@bdice I've added an entry to the changelog. I'm ok with updating framework documentation as part of some general overhaul.

PR #122, opened 2019-02-14
Create a hard-link within files when assigning a dataset.

csadorf:
Instead of creating an automatic copy.

There are many good reasons for this change in behavior:

  1. Creating an automatic copy is the much more dangerous option,
  potentially leading to an unexpected copy of terra-bytes of data.
  2. Creating an automatic copy is not what a user familiar with h5py
  would expect to happen.
  3. Creating a hard-link is actually closer to the behavior of standard
  in-memory dicts, which would also create a reference instead of a copy
  when assigning the same array to multiple keys.
  4. With this behavior, the user still has the option to create a copy
  if desired, the reverse would not be immediately possible.

The only viable alternative behavior that I see is to not allow assigning
arrays at all, which I find much less appealing.

### DIFF
--- a/signac/core/h5store.py
+++ b/signac/core/h5store.py
@@ -73,12 +73,15 @@ def _h5set(file, grp, key, value, path=None):
     # Guard against assigning a group to itself, e.g., `h5s[key] = h5s[key]`,
     # where h5s[key] is a mapping. This is necessary, because the original
     # mapping would be deleted prior to assignment.
-    if isinstance(value, H5Group) and key in grp:
-        if grp[key] == value._group:
-            return  # Groups are identical, do nothing.
-
-    # Delete any existing data
     if key in grp:
+        if isinstance(value, H5Group):
+            if grp[key] == value._group:
+                return  # Groups are identical, do nothing.
+        elif isinstance(value, h5py._hl.dataset.Dataset):
+            if grp == value.parent:
+                return  # Dataset is identical, do nothing.
+
+        # Delete any existing data
         del grp[key]
 
     # Mapping-types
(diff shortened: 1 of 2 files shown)


bdice:
@csadorf
1. I think it's sensible to copy automatically if it's an inter-file assignment (e.g. from one job's datastore to another job's datastore).
2. I don't know what the expected deletion behavior is for hard links - if I assign `h5s['b'] = h5s['a']` and then delete `h5s['a']`, does it also delete `h5s['b']`? In normal dictionaries, it seems like the key `b`'s value is still there, as if it's being ref-counted. I'm not sure if the same mechanism applies to hard links via HDF5. I looked briefly but didn't see a test for this. Maybe add a test codifying the expected behavior here.

csadorf:
> @csadorf
> 
> 1. I think it's sensible to copy automatically if it's an inter-file assignment (e.g. from one job's datastore to another job's datastore).
It's safer to expect the user to always create an explicit copy because it may not always be immediately visible wether two references are operating on the same underlying file. Furthermore, changing the behavior based on context is more likely to lead to hard-to-detect bugs.

> 2. I don't know what the expected deletion behavior is for hard links - if I assign `h5s['b'] = h5s['a']` and then delete `h5s['a']`, does it also delete `h5s['b']`? In normal dictionaries, it seems like the key `b`'s value is still there, as if it's being ref-counted. I'm not sure if the same mechanism applies to hard links via HDF5. I looked briefly but didn't see a test for this. Maybe add a test codifying the expected behavior here.

The expected behavior is that the deleting a reference will only delete the value if the reference count drops to zero. That's the behavior you would expect on the file system, HDF5-files behave in the same way. I don't think it's strictly necessary, but I can add a unit test to explicitly guarantee this behavior.

csadorf:
As an additional argument against automatic copies between files: Users will get used to the fact that assigning a dataset creates a hard-link and the need to explicitly create copies when desired. Switching this behavior just because we are assigning between files will be unexpected and is therefore a bad idea.

We should follow *explicit is better than implicit* here.

PR #123, opened 2019-02-15
Feature/h5store multiprocessing

csadorf:
This patch adds unit tests for multi-processing environments.

Specifically I test how we can interact with one store by opening multiple writers and readers under varying conditions, for example within the same instance, across multiple processes, but within the same instance, and across different instances.

I also added a simple test how one could enable the SWMR mode manually, however there are too many caveats with this mode to make this any more official in my opinion, at least at this stage.

Finally, I enabled the option to internally override certain file parameters, such as the opening-mode, which allows us to execute specific functions in read-only mode, such as the `len` or the `in` checks.

### DIFF
--- a/signac/core/h5store.py
+++ b/signac/core/h5store.py
@@ -400,7 +400,7 @@ class H5Store(MutableMapping):
         try:
             with _ensure_open(self, mode='r'):
                 return len(self._file)
-        except OSError as error:
+        except (OSError, IOError) as error:
             if 'errno = 2' in str(error):
                 return 0
             else:
@@ -410,7 +410,7 @@ class H5Store(MutableMapping):
         try:
             with _ensure_open(self, mode='r'):
                 return key in self._file
-        except OSError as error:
+        except (OSError, IOError) as error:
             if 'errno = 2' in str(error):
                 return False
             else:
--- a/signac/core/h5store.py
+++ b/signac/core/h5store.py
@@ -4,6 +4,7 @@
 "Data store implementation with backend HDF5 file."
 import logging
 import os
+import errno
 import warnings
 import array
 from threading import RLock
@@ -401,8 +402,8 @@ class H5Store(MutableMapping):
             with _ensure_open(self, mode='r'):
                 return len(self._file)
         except (OSError, IOError) as error:
-            if 'errno = 2' in str(error):
-                return 0
+            if 'errno = {}'.format(errno.ENOENT) in str(error):
+                return 0     # file does not exist
             else:
                 raise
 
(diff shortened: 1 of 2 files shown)


csadorf:
@bdice I'm ready to call the `H5Store` *good enough for release* after this patch.

tests/test_h5store.py
```diff
@@ -623,6 +636,88 @@ def set_x(x):
...
 
+def _read_from_h5store(filename, **kwargs):
+    from signac.core.h5store import H5Store
+    with H5Store(filename, **kwargs) as h5s:
+        list(h5s)
+
+
+@unittest.skipIf(
+    six.PY2 or (six.PY3 and sys.version_info.minor < 4),
+    'tests only implemented for Python > 3.4')
```
bdice: Do you mean `Python >= 3.4`?

tests/test_h5store.py
```diff
@@ -623,6 +636,88 @@ def set_x(x):
...
+class H5StoreMultiProcessingTest(BaseH5StoreTest):
+
+    def test_single_writer_multiple_reader_same_process(self):
+        with self.open_h5store() as w1:
```
bdice: Please write out `writer1`, `reader1`, etc. It took me a bit to understand what was going on there.

tests/test_h5store.py
```diff
@@ -623,6 +636,88 @@ def set_x(x):
...
+            p.join()
+
+        with self.open_h5store() as w1:
+            read()
+            w1['test'] = True
+            read()
+
+    def test_multiple_reader_different_process_no_swmr(self):
+
+        read_cmd = "python -c 'from signac.core.h5store import H5Store; "
```
bdice: Maybe use a triple-quoted string instead?

signac/core/h5store.py
```diff
@@ -385,12 +397,24 @@ def __iter__(self):
                 yield key
 
     def __len__(self):
-        with _ensure_open(self):
-            return len(self._file)
+        try:
+            with _ensure_open(self, mode='r'):
+                return len(self._file)
+        except (OSError, IOError) as error:
+            if 'errno = 2' in str(error):
```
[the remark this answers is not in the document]
csadorf: It means that the file does not exist. I'll add a comment to make that obvious.

tests/test_h5store.py
csadorf: Yes, but also I think it should run on 3.4 now, so I'll fix that.

tests/test_h5store.py
csadorf: That would preserve new-lines, which is not what we want.

csadorf:
@bdice All comments addressed, let me know if there is anything else, otherwise I'll merge.

bdice:
@csadorf merge it up! ⤴️

Issue #120, opened 2019-02-13
Chain calls to job methods, specifically `.init()`


bdice:
### Feature description

I would like to be able to call `job = project.open_job(statepoint).init()`, which would be possible if `init()` returned `self`. This is related to #71. Other methods that could be chained if desired: `reset_statepoint`, `update_statepoint`, `clear`, `reset`, `move`, `sync`.

### Proposed solution

Make `init()` return `self`.

csadorf:
This is amazing. I was just about to create this issue **today**. And yes, that is a perfectly reasonable solution!

csadorf:
Resolved with merge of #125 .

PR #125, opened 2019-02-16
Resolve issue #120 by allowing job methods to be chained.

bdice:
See issue #120 for description.

I also looked at the possibility of chaining calls to `Project`, but there were no candidate methods that I thought would be good to chain.

### DIFF
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -347,12 +347,14 @@ class Job(object):
         Returns the calling job.
 
         :param force:
-                Overwrite any existing state point's manifest
-                files, e.g., to repair them when they got corrupted.
+            Overwrite any existing state point's manifest
+            files, e.g., to repair them when they got corrupted.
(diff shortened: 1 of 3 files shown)


bdice:
@csadorf This should be ready for review.

signac/contrib/job.py
```diff
@@ -344,18 +344,23 @@ def init(self, force=False):
...
 
+        Returns the calling job.
+
         :param force:
                 Overwrite any existing state point's manifest
                 files, e.g., to repair them when they got corrupted.
         :type force:
                 bool
+        :return: The job handle.
+        :rtype: :class:`~.Job`
```
csadorf: Could you adjust the formatting here to match lines 349 to 353?

bdice: I indented those and also de-indented 349-353 because they were double-indented and didn't match the rest of the file.

PR #127, opened 2019-02-16
Enable IPython display

### LINKED ISSUE #38: ipython notebook html display for jobs
**[Original report](https://bitbucket.org/glotzer/signac/issue/38) by Anonymous.**

Since ipython/jupyter notebook can display html output, I thought it might be nice to display jobs in a table (the same way pandas dataframes are displayed). So I copied an [online example](http://calebmadrigal.com/display-list-as-table-in-ipython-notebook/):

```
#!python

class JobTable(list):
    """ Overridden list class which takes a list of signac jobs 
        and renders an HTML table in IPython Notebook with the id
        and statepoint attributes. """
    
    def _repr_html_(self):
        """Return html table representing myself, the list of jobs.

        Operations
        ----------
        1. Determine what the attributes are.
        2. Add header row with name of each attribute.
        3. Add row to table for each job, with job id in first column.
        """
        attribs = set(sum((job.statepoint().keys() for job in self), []))
        html = ["<table>"]
        html.append("<td><b>{0}</b></td>".format('id'))
        for attrib in attribs:
            html.append("<td><b>{0}</b></td>".format(attrib))
            
        for job in self:
            html.append("<tr>")
            html.append("<td><b>{0}</b></td>".format(job.get_id()))
            
            for attrib in attribs:
                try:
                    html.append("<td>{0}</td>".format(job.statepoint()[attrib]))
                except KeyError:
                    html.append("<td></td>")
            
            html.append("</tr>")
        html.append("</table>")
        return ''.join(html)
```

This can then be used:

```
#!python
from IPython.display import display
display(JobTable(project.find_jobs({'type':'pmf'})))
```

And an HTML table describing the jobs appears in the results cell (image attached). If this were implemented as part of signac, it would be cleaner to change `find_jobs` from a generator function to instead return an iterator class with the `_repr_html_` method. Do you the pretty notebook display is worth embedding into signac?

bdice:
## Description
Resolves #38 by adding IPython display capabilities for the `JobsCursor` class.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

This pull request is based on

## Checklist:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -29,7 +29,7 @@ Added
  - Add automatic cast of numpy arrays to lists when storing them within a `JSONDict`, e.g., a `job.statepoint` or `job.document`.
  - Enable `Collection` class to manage collections stored in gzip files.
  - Enable deleting of `JSONDict` keys through the attribute interface, e.g., `del job.doc.foo`.
- - Added HTML representation of JobsCursor objects for Jupyter Notebooks (uses pandas).
+ - Added HTML representation of Project and JobsCursor objects for Jupyter Notebooks (uses pandas).
 
 Fixed
 +++++
--- a/changelog.txt
+++ b/changelog.txt
@@ -29,6 +29,7 @@ Added
  - Add automatic cast of numpy arrays to lists when storing them within a `JSONDict`, e.g., a `job.statepoint` or `job.document`.
  - Enable `Collection` class to manage collections stored in gzip files.
  - Enable deleting of `JSONDict` keys through the attribute interface, e.g., `del job.doc.foo`.
+ - Added HTML representation of JobsCursor objects for Jupyter Notebooks (uses pandas).
 
 Fixed
 +++++
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -133,6 +133,8 @@ class Project(object):
     FN_CACHE = '.signac_sp_cache.json.gz'
     "The default filename for the state point cache file."
 
+    _use_pandas_for_html_repr = True  # toggle use of pandas for html repr
+
     def __init__(self, config=None):
         if config is None:
             config = load_config()
@@ -168,20 +170,7 @@ class Project(object):
                    id=self.get_id(), rd=self.root_directory(), wd=self.workspace())
 
     def _repr_html_(self):
-        html = repr(self)
-        len_self = len(self)
-        if len_self <= 100:
-            try:
-                # Attempt to use pandas representation
-                import pandas
-                with pandas.option_context("display.max_rows", 20):
-                    html += self.to_dataframe()._repr_html_()
-            except ImportError:
-                warnings.warn('Install pandas for a pretty representation of JobsCursor.')
-                html += '<br/><strong>{}</strong> result(s) found'.format(len_self)
-        else:
-            html += '<br/><strong>{}</strong> result(s) found'.format(len_self)
-        return html
+        return repr(self) + self.find_jobs()._repr_html_jobs()
 
     def __eq__(self, other):
         return repr(self) == repr(other)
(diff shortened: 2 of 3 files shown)


bdice:
@glotzerlab/signac-developers I'd like some input on the desired appearance of this PR, specifically wrt nested keys. I was going to show them as dot-separated. The current code probably doesn't work for nested keys. Also how should this be tested? 

image

Note: job 99914b932bd37a50b983c5e7c90ae93b has statepoint `{}`.

csadorf:
I'd flatten all keys, so treat all keys separately and join nested keys with dots. 

There should be some fixtures for various valid statepoint (basic and nested, valid types) in the job unit tests. I'd make sure that this function doesn't fail for any of those and compare the raw output for a few select cases.

bdice:
Another question: how many rows should we show? It's probably good to clip it.

Pandas has configurable package-wide settings for this. Default is 60 rows and 8 columns. An alternative would be to just return the pandas dataframe and allow its `_repr_html_` method to be called. That introduces a requirement for pandas to have pretty-printed HTML, but it would give users the ability to configure the display the same way they would already do for pandas dataframes. It also simplifies our display and testing logic by a lot.

https://pandas.pydata.org/pandas-docs/stable/user_guide/options.html

```python
pd.options.display.chop_threshold      pd.options.display.max_colwidth
pd.options.display.colheader_justify   pd.options.display.max_info_columns
pd.options.display.column_space        pd.options.display.max_info_rows
pd.options.display.date_dayfirst       pd.options.display.max_rows
pd.options.display.date_yearfirst      pd.options.display.max_seq_items
pd.options.display.encoding            pd.options.display.memory_usage
pd.options.display.expand_frame_repr   pd.options.display.multi_sparse
pd.options.display.float_format        pd.options.display.notebook_repr_html
pd.options.display.html                pd.options.display.pprint_nest_depth
pd.options.display.large_repr          pd.options.display.precision
pd.options.display.latex               pd.options.display.show_dimensions
pd.options.display.max_categories      pd.options.display.unicode
pd.options.display.max_columns         pd.options.display.width
```

csadorf:
I like the idea of delgating to pandas. How would we deal with pandas not being installed?

bdice:
> 
> 
> I like the idea of delgating to pandas. How would we deal with pandas not being installed?

I would just return `str(self)` or so, similar to the existing default behavior.

bdice:
## With 🐼:
image

## Without 🐼:
image


(commit 6b5c5e7cf9a064a148a5bcfb77b893051e89e7c0)

bdice:
Should this also be the default Jupyter Notebook representation for `Project` or `Job`, possibly with some alteration?

csadorf:
> Should this also be the default Jupyter Notebook representation for `Project` or `Job`, possibly with some alteration?

For the `Project` it would be important to also show the directories I think. What would the `Job` HTML view look like?

signac/contrib/project.py
```diff
@@ -1827,6 +1827,15 @@ def _export_sp_and_doc(job):
             data={job._id: dict(_export_sp_and_doc(job)) for job in self},
             orient='index').infer_objects()
 
+    def _repr_html_(self):
+        """Represent a JobsCursor in IPython Notebooks."""
+        try:
+            # Attempt to use pandas representation
+            return self.to_dataframe()._repr_html_()
+        except ImportError:
+            import warnings
```
csadorf: Can you move that import to the top of the module?

changelog.txt
```diff
@@ -29,6 +29,7 @@ Added
  - Add automatic cast of numpy arrays to lists when storing them within a `JSONDict`, e.g., a `job.statepoint` or `job.document`.
  - Enable `Collection` class to manage collections stored in gzip files.
  - Enable deleting of `JSONDict` keys through the attribute interface, e.g., `del job.doc.foo`.
+ - Added HTML representation of JobsCursor objects for Jupyter Notebooks (uses pandas).
```
csadorf: Please use present tense and code-markup for `JobsCursor`.

PR #129, opened 2019-02-19
Implement the H5StoreManager class.

csadorf:
## Description

Introduces the `DictManager` and the `H5StoreManager` classes that are exposed as `job.stores` and `project.stores`.

## Motivation and Context

It would be highly advantageous to enable users to easily manage multiple HDF5-files instead of just one. There are plenty of reasons why that is a good idea:

1. While the HDF5-format is generally mutable, it is fundamentally designed to be used as an immutable data container. In fact, even the latest version is internally restricted to append-only operations and a file will continue to grow with repeated addition and edits. The only way to reduce the file size is to create a new file (e.g. with `h5repack`).
2. A single large file is difficult to synchronize. Even if we only change a few MBs or even KBs of a multiple TBs-large file, there is a large chance that synchronization tools like `rsync` or GLOBUS will have to transfer the complete file. Furthermore, all data stored within one large file share one modification time stamp, making fine-grained synchronization even more difficult. Finally, synchronization of parts of a file are impossible.
3. A single large file, especially when it is repeatedly mutated poses a data corruption risk. The HDF5 library uses file locks to reduce the risk of data corruption, but there is no journaling as of yet and the corruption of parts of the file, e.g., through a crashed application (think job killed due to walltime stop etc.) might corrupt the whole file leading to complete data loss.
4. The HDF5-format does not play well with concurrent/parallel access. While the single-writer-multiple-reader (SWMR) mode is implemented for some libraries, it comes with severe limitations and requires a special access style that cannot be easily made transparent to the user. Existence checks etc would be much easier to execute concurrently if they are kept to the file system level without needing to open the file.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

This pull request is based on

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -24,7 +24,7 @@ Highlights
 Added
 +++++
 
- - Adds an ``H5Store`` class, useful for storing array-like data with an HDF5 backend. Accessible via ``with job.data:`` or ``with project.data:``.
+ - Adds an ``H5Store`` and a ``H5StoreManager`` class, useful for storing array-like data with an HDF5 backend. Accessible via ``with job.data:`` or ``with project.data:``.
  - Adds the ``signac.get_job()`` and the ``signac.Project.get_job()`` functions which allow users to get a job handle by switching into or providing the job's workspace directory.
  - Add automatic cast of numpy arrays to lists when storing them within a `JSONDict`, e.g., a `job.statepoint` or `job.document`.
  - Enable `Collection` class to manage collections stored in gzip files.
--- a/changelog.txt
+++ b/changelog.txt
@@ -24,7 +24,7 @@ Highlights
 Added
 +++++
 
- - Adds an ``H5Store`` and a ``H5StoreManager`` class, useful for storing array-like data with an HDF5 backend. Accessible via ``with job.data:`` or ``with project.data:``.
+ - Adds an ``H5Store`` and a ``H5StoreManager`` class, useful for storing array-like data with an HDF5-backend. Those classes are exposed as part of the ``job.data``, ``job.stores``, ``project.data``, and ``project.stores``, properties.
  - Adds the ``signac.get_job()`` and the ``signac.Project.get_job()`` functions which allow users to get a job handle by switching into or providing the job's workspace directory.
  - Add automatic cast of numpy arrays to lists when storing them within a `JSONDict`, e.g., a `job.statepoint` or `job.document`.
  - Enable `Collection` class to manage collections stored in gzip files.
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -303,6 +303,28 @@ class Project(object):
 
     @property
     def stores(self):
+        """Access HDF5-stores associated wit this project.
+
+        Use this property to access an HDF5-file within the project's root
+        directory using the H5Store dict-like interface.
+
+        This is an example for accessing an HDF5-file called 'my_data.h5' within
+        the project's root directory:
+
+            project.stores['my_data']['array'] = np.random((32, 4))
+
+        This is equivalent to:
+
+            H5Store(project.fn('my_data.h5'))['array'] = np.random((32, 4))
+
+        Both the `project.stores` and the `H5Store` itself support attribute
+        access. The above example could therefore also be expressed as
+
+            project.stores.my_data.array = np.random((32, 4))
+
+        :return: The HDF5-Store manager for this project.
+        :rtype: :class:`~..core.h5store.H5StoreManager
+        """
         return H5StoreManager(self._rd)
 
     @property
@@ -314,7 +336,7 @@ class Project(object):
             return project.store['signac_data']
 
         :return: An HDF5-backed datastore.
-        :rtype: :class:`~signac.core.h5store.H5DictManager`
+        :rtype: :class:`~..core.h5store.H5Store`
         """
         return self.stores[self.KEY_DATA]
 
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -18,7 +18,7 @@ from multiprocessing.pool import ThreadPool
 from .. import syncutil
 from ..core.json import json
 from ..core.jsondict import JSONDict
-from ..core.h5store import H5Store
+from ..core.h5store import H5StoreManager
 from .collection import Collection
 from ..common import six
 from ..common.config import load_config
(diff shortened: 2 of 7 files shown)


csadorf:
@glotzerlab/signac-maintainers This is ready for review.

csadorf:
> LGTM, I don't work with the H5Store enough to approve the PR but looking over it I don't see anything obviously wrong. We should make sure to update [glotzerlab/signac-docs#6](https://github.com/glotzerlab/signac-docs/pull/6) in light of the changes made here.

Thank you, that's a good point and I will go through the docs before merging.

vyasr:
I will review this tomorrow, trying to finish up another task before I start going back through signac PRs.

changelog.txt
```diff
@@ -24,7 +24,7 @@ Highlights
 Added
 +++++
 
- - Adds an ``H5Store`` class, useful for storing array-like data with an HDF5 backend. Accessible via ``with job.data:`` or ``with project.data:``.
+ - Adds an ``H5Store`` and a ``H5StoreManager`` class, useful for storing array-like data with an HDF5 backend. Accessible via ``with job.data:`` or ``with project.data:``.
```
vyasr: Mention `job.stores` as well.

signac/contrib/project.py
```diff
@@ -305,21 +301,26 @@ def doc(self):
...
     def data(self):
         """The data associated with this project.
 
+        Equivalent to:
+
+            return project.store['signac_data']
+
         :return: An HDF5-backed datastore.
-        :rtype: :class:`~signac.core.h5store.H5Store`
+        :rtype: :class:`~signac.core.h5store.H5DictManager`
```
vyasr: This is still returning an H5Store instance, right? In any case, H5DictManager should just be H5StoreManager.

signac/core/dict_manager.py
```diff
@@ -0,0 +1,115 @@
...
+class DictManager(object):
+
+    cls = None
+    suffix = None
+
+    __slots__ = ['_prefix', '_dict_registry']
+
+    def __init__(self, prefix):
+        assert self.cls is not None
+        assert self.suffix is not None
```
vyasr: In the interest of supporting future developers who find alternative uses for subclassing DictManager I think it's worth adding explicit error messages here indicating that the subclasses need to define both of these class attributes.

signac/core/dict_manager.py
```diff
@@ -0,0 +1,115 @@
...
+            else:
+                raise error
+        except Exception as error:
+            try:
+                del self[tmp_key]
+            except KeyError:
+                pass
+            raise error
+        else:
+            del self._dict_registry[key]
```
vyasr: I don't follow why this is necessary. The file corresponding to `key` was successfully created if we reach this point, why does it need to be removed from the tracked list?

signac/core/dict_manager.py
```diff
@@ -0,0 +1,115 @@
...
+        if name.startswith('__') or name in self.__slots__:
+            super(DictManager, self).__delattr__(name)
+        else:
+            self.__delitem__(name)
+
+    def __iter__(self):
+        for fn in os.listdir(self.prefix):
+            m = re.match('(.*){}'.format(self.suffix), fn)
+            if m:
+                yield m.groups()[0]
```
vyasr: Using `re` seems like overkill here, I'd switch this to using `glob.glob`: `yield from glob.glob("*.{}".format(self.suffix))` (without using `yield from` for py2 compatibility).

signac/core/dict_manager.py
csadorf: It's not really necessary, I figured it might be a good idea to replace any references of "old" H5Store under the same key, but there is virtually no difference. I'll remove that.

signac/core/dict_manager.py
csadorf: I can't directly extract the key name with `glob`.

vyasr: I suppose you'd still have to do something like `...replace(self.suffix, '')` to get just the key without the extension, that's all that you're referring to right? I personally think that's still clearer than using a regex since you're really just parsing out an extension from a set of filenames, but I'm OK with leaving it as is if you prefer since it doesn't simplify the code that much.

tests/test_job.py
```diff
@@ -1384,5 +1384,368 @@ def open_data(job):
...
+            self.assertFalse(bool(job.stores.test))
+            self.assertEqual(len(job.stores.test), 0)
+            self.assertNotIn(key, job.stores.test)
+            job.stores.test[key] = d
+            self.assertTrue(bool(job.stores.test))
+            self.assertEqual(len(job.stores.test), 1)
+            self.assertIn(key, job.stores.test)
+            self.assertEqual(job.stores.test[key], d)
+            self.assertEqual(job.stores.test.get(key), d)
+            self.assertEqual(job.stores.test.get('bs', d), d)
```
bdice: I would suggest another word like `'nonexistent_key'` so it's more clear.

tests/test_job.py
```diff
@@ -1384,5 +1384,368 @@ def open_data(job):
...
+        with self.open_data(job):
+            job.stores.test[key] = d
+            self.assertIn(key, job.stores.test)
+            self.assertEqual(len(job.stores.test), 1)
+            job.stores.test.clear()
+            self.assertNotIn(key, job.stores.test)
+            self.assertEqual(len(job.stores.test), 0)
+
+    def test_reopen(self):
+        key = 'clear'
```
bdice: Should this be `key = 'reopen'`?

csadorf: Yes.

tests/test_job.py
csadorf: Ummm, yes. Oops. :D

signac/core/dict_manager.py
csadorf: Turns out this is actually necessary. I tried to delete this line and the `test_assign` and `test_assign_data` unit tests started to fail.

signac/core/dict_manager.py
csadorf: I just tried it. I would also need to strip off the directory name, so I'm sticking to the original implementation.

However, should we explicitly skip hidden h5-files?

tests/test_job.py
csadorf: I'm going to fix that in a separate commit, because there are a bunch of other occurrences of the same pattern in other test modules.

csadorf:
@vyasr I've addressed all of your comments.

signac/core/dict_manager.py
vyasr: I don't think that's necessary. I view creating and editing hidden hd5 files as a feature, not a bug. If users explicitly access a hidden file I think it's safe to expect that they know what they're doing.

vyasr:
Approving pending the change that Bradley requested.

PR #132, opened 2019-02-20
Fix/document view order

vyasr:
Document required order for `signac view` CLI arguments.


signac/__main__.py
```diff
@@ -1236,14 +1236,18 @@ def main():
         type=str,
```
csadorf: Can you add an example to the `parser_view` instance directly? Similar to line 1203.

vyasr:
You're right, having it in the parser docs directly is better. In the process of making this change, I expanded the view documentation in general.

PR #140, opened 2019-02-25
Elevate all dots-in-keys deprecation warnings to errors.

csadorf:
The use of an invalid key should lead to the raise of an
InvalidKeyError, which is newly defined in the errors module.

### DIFF
--- a/signac/contrib/collection.py
+++ b/signac/contrib/collection.py
@@ -16,7 +16,6 @@ import sys
 import io
 import re
 import logging
-import warnings
 import argparse
 import operator
 from itertools import islice
@@ -201,9 +200,11 @@ def _build_index(docs, key, primary_key):
             except KeyError:
                 pass
             else:
-                warnings.warn(
-                    "Using keys with dots ('.') is pending deprecation in the future!",
-                    PendingDeprecationWarning)
+                from ..errors import InvalidKeyError
+                raise InvalidKeyError(
+                    "\nThe use of '.' (dots) in keys is invalid.\n\n"
+                    "See http://www.signac.io/document-wide-migration/ "
+                    "for a recipe on how to replace dots in all keys.")
                 # inlined for performance
                 if type(v) is list:     # performance
                     index[_to_tuples(v)].add(doc[primary_key])
--- a/signac/core/synceddict.py
+++ b/signac/core/synceddict.py
@@ -3,7 +3,6 @@
 # This software is licensed under the BSD 3-Clause License.
 "Synchronized dictionary."
 import logging
-import warnings
 from contextlib import contextmanager
 from functools import wraps
 from copy import deepcopy
@@ -89,12 +88,11 @@ class _SyncedDict(MutableMapping):
     def _validate_key(key):
         "Emit a warning or raise an exception if key is invalid. Returns key."
         if '.' in key:
-            from ..warnings import SignacDeprecationWarning
-            warnings.warn(
+            from ..errors import InvalidKeyError
+            raise InvalidKeyError(
                 "\nThe use of '.' (dots) in keys is deprecated and may lead to "
                 "unexpected behavior!\nSee http://www.signac.io/document-wide-migration/ "
-                "for a recipe on how to replace dots in all keys.",
-                SignacDeprecationWarning)
+                "for a recipe on how to replace dots in all keys.")
         return key
 
     def _dfs_convert(self, root):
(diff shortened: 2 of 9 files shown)


vyasr:
I'm a little confused by the organization with respect to the `validate_key` method. I think it's probably safe to prevent dots in keys for an arbitrary `dict_manager`, but if that's not the case it seems unnecessary to introduce the function at that level. A reasonable alternative would be to make it a `classmethod` and just include `cls` in the error message.

csadorf:
> I'm a little confused by the organization with respect to the `validate_key` method. I think it's probably safe to prevent dots in keys for an arbitrary `dict_manager`, but if that's not the case it seems unnecessary to introduce the function at that level. A reasonable alternative would be to make it a `classmethod` and just include `cls` in the error message.

That's exact reason I did not do this. I don't think there is any reason to assume that *any* `DictManager` has this restriction.

Using a `staticmethod` provides a sensible path forward to overriding default behavior, why would you prefer a `classmethod` if there is no dependence on the `cls`?

signac/contrib/collection.py
```diff
@@ -201,9 +200,11 @@ def _build_index(docs, key, primary_key):
             except KeyError:
                 pass
             else:
-                warnings.warn(
-                    "Using keys with dots ('.') is pending deprecation in the future!",
-                    PendingDeprecationWarning)
+                from ..errors import InvalidKeyError
+                raise InvalidKeyError(
+                    "\nThe use of '.' (dots) in keys is invalid.\n\n"
+                    "See http://www.signac.io/document-wide-migration/ "
```
bdice: Replace with https://signac.io/document-wide-migration

signac/contrib/collection.py
```diff
@@ -201,9 +200,11 @@ def _build_index(docs, key, primary_key):
...
+                    "for a recipe on how to replace dots in all keys.")
```
bdice: Should we say _existing_ keys?

vyasr:
Sorry, I misread the diff. I didn't realize that the `__setitem__` use of `validate_key` was in the parent class, so I was suggesting that you could remove it from the `DictManager` entirely.

signac/core/synceddict.py
```diff
@@ -89,12 +88,11 @@ def __init__(self, initialdata=None, parent=None):
...
         if '.' in key:
-            from ..warnings import SignacDeprecationWarning
-            warnings.warn(
+            from ..errors import InvalidKeyError
+            raise InvalidKeyError(
                 "\nThe use of '.' (dots) in keys is deprecated and may lead to "
                 "unexpected behavior!\nSee http://www.signac.io/document-wide-migration/ "
-                "for a recipe on how to replace dots in all keys.",
-                SignacDeprecationWarning)
+                "for a recipe on how to replace dots in all keys.")
```
bdice: Make this message match the other one I commented on.

tests/test_collection.py
```diff
@@ -293,8 +293,7 @@ def test_find_int_float(self):
         self.assertEqual(self.c.find_one({'a': {'$type': 'float'}})['_id'], id_float)
 
     def test_find_with_dots(self):
-        with warnings.catch_warnings(record=True) as w:
-            warnings.simplefilter('once')
+        with self.assertRaises(InvalidKeyError):
```
bdice: Should this `with` surround each of the individual line(s) that are expected to fail? I think this test passes if "at least one" error is raised, but this should raise multiple errors if I understand it correctly.

tests/test_h5store.py
```diff
@@ -406,15 +405,12 @@ class Foo(object):
             self.assertEqual(h5s[key], d)
 
     def test_keys_with_dots(self):
-        with warnings.catch_warnings(record=True) as w:
-            warnings.simplefilter('always')
+        with self.assertRaises(InvalidKeyError):
```
bdice: Same as above - I'd move this around and wrap each of the lines that are expected to trigger failure in the `assertRaises` block.

csadorf:
> I made some change recommendations - I guess I didn't realize it was already merged?

The PR was approved by @vyasr , so I merged it. I'm going to address your comments and potentially open a separate PR.

tests/test_collection.py
csadorf: I tried to touch the unit test as little as possible, simply switching the warnings catch with the assertion. You are right that only one particular line is expected to trigger the exception. We can make that more explicit.

tests/test_h5store.py
csadorf: See comment above.

bdice:
@csadorf Yeah, sorry about that - I think we were reviewing in parallel and he finished first. 😋

PR #144, opened 2019-02-25
Misc/validate collections documents

csadorf:


signac/contrib/collection.py
```diff
@@ -493,6 +494,24 @@ def __getitem__(self, _id):
             self._assert_open()
             raise
 
+    @staticmethod
+    def _validate_key(key):
+        "Emit a warning or raise an exception if key is invalid. Returns key."
+        if '.' in key:
+            from ..errors import InvalidKeyError
```
vyasr: We normally don't put imports nested anywhere unless it's a `try...except` block, is there a reason not to put this at the top of the file?

signac/contrib/collection.py
```diff
@@ -503,7 +522,7 @@ def __setitem__(self, _id, doc, _trust=False):
         if _trust:
             self._docs[_id] = doc
         else:
-            self._docs[_id] = json.loads(json.dumps(doc))
+            self._docs[_id] = self._validate_doc(json.loads(json.dumps(doc)))
```
vyasr: Is the `loads(dumps(doc))` just to ensure that the document is valid JSON? If not, then this is a no-op, right? And if so, should we wrap it in a `try...except` to provide a more informative error?

signac/contrib/collection.py
csadorf: Yes, because there are otherwise module interdependencies.

signac/contrib/collection.py
csadorf: Yes, exactly. I'll see if I can easily provide an informative error message.

csadorf: I forgot to mention, it also ensures type normalization.

PR #145, opened 2019-02-25
Align logo

csadorf:


csadorf:
This doesn't really work. Applied the logo resize directly to the base branch.

PR #146, opened 2019-02-25
Allow signac find to display just sp or doc if desired

vyasr:
## Description
Enables printing of only printing statepoints or documents (without printing both) when using signac find.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

This pull request is based on

## Checklist:

If necessary:


vyasr:
This should be safe to make the release branch the target since I rebased it on current develop, so up to you.

csadorf:
This will have to wait until 1.1.

signac/__main__.py
```diff
@@ -1223,7 +1231,22 @@ def main():
...
-        help="Show the state point and document of each job.")
+        help="Show the state point and document of each job. Equivalent to "
+        "--sp NUM --doc NUM.")
+    parser_find.add_argument(
+        '--sp',
+        type=int,
+        nargs='?',
+        const=3,
+        help="Show the state point of each job. The parameter is the level of "
+        "nesting to display.")
```
csadorf: I think it would be better to use the argument to limit which state point key(s) are shown.
For example: `signac find --sp foo` would only show the `foo` state point, whereas `signac find --sp foo bar` shows both `foo` and `bar`. Both `--sp` and `--doc` would always imply `--show` and the `--show` argument would still be used to specify the nesting depth.

signac/__main__.py
```diff
@@ -342,26 +343,33 @@ def main_find(args):
...
-        def format_lines(cat, _id, s):
-            if args.one_line:
-                if isinstance(s, dict):
-                    s = json.dumps(s, sort_keys=True)
-                return _id[:len_id] + ' ' + cat + '\t' + s
-            else:
-                return pformat(s, depth=args.show)
+        if args.sp:
+            warnings.warn(
+                "Specifying --show and --sp results in --sp being ignored")
```
csadorf: The `warnings` module is not good to show warnings like those, because it comes with a traceback. It would be better to either use the `logging` module or simply `print()`.

signac/__main__.py
vyasr: If you specify `--doc` with some set of keys and `--show`, would you expect to get the full statepoint printed, or would you have to specify `--sp` as well in order to get it to print that? I'm not sure which API is more natural.

csadorf: Since it in its current form it would be impossible to say "do not show the state point at all", I would opt for the latter.

vyasr: OK that's consistent with how I've written the code right now.

csadorf:
@vyasr I am a little bit confused. You marked two conversations of this PR as resolved, but didn't address the comments. Were you planning on addressing them or do you simply not agree with them?

vyasr:
@csadorf I'm also confused. I thought I pushed changes... :D 
Pushed now.

csadorf:
@vyasr Please make sure to target this PR into `master`, thank you.

signac/__main__.py
```diff
@@ -1223,7 +1232,20 @@ def main():
         type=int,
         nargs='?',
         const=3,
-        help="Show the state point and document of each job.")
+        help="Show the state point and document of each job. Equivalent to "
+        "--sp NUM --doc NUM.")
```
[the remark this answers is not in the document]
vyasr: Yes this was wrong. I've updated the docs.

vyasr:
The error is that h5 pypy error that has popped up on a couple other branches as well, I think. It isn't related to these changes.

csadorf:
> The error is that h5 pypy error that has popped up on a couple other branches as well, I think. It isn't related to these changes.

All tests passed after restart. Ready for review?

vyasr:
Yes, ready for review!

vyasr:
Good call, I made it check whether keys exist before filtering what to print.

Issue #148, opened 2019-02-27
Incorrect error message when trying to move job across file systems


csadorf:
### Description

As reported by @rosecers, trying to move a job between file systems using the `job.move()` function fails internally with `OSError: [Errno 18] Invalid cross-device link: `, however signac reports a  `signac.contrib.errors.DestinationExistsError`, because of too coarse exception handling.

### To reproduce

Move a job between different file systems.

### Error output

```
Traceback (most recent call last):
  File "/home/rosecers/.local/lib/python3.5/site-packages/signac/contrib/job.py", line 391, in move
    os.rename(self.workspace(), dst.workspace())
OSError: [Errno 18] Invalid cross-device link: 'REDACTED' -> 'REDACTED'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "operations.py", line 524, in <module>
    flow.run()
  File "/home/rosecers/.local/lib/python3.5/site-packages/flow/operations.py", line 194, in run
    operation(job)
  File "operations.py", line 513, in move
    job.move(other_project)
  File "/home/rosecers/.local/lib/python3.5/site-packages/signac/contrib/job.py", line 393, in move
    raise DestinationExistsError(dst)
signac.contrib.errors.DestinationExistsError: 17adbd4ffbcaf50566f885b366a773d0
```

### System configuration

Please complete the following information:

 - Operating System [e.g. macOS]: n/a
 - Version of Python [e.g. 3.7]: n/a
 - Version of signac [e.g. 1.0]: n/a

csadorf:
Fixed with merge of PR #149 .

PR #149, opened 2019-02-27
Properly handle Job.move() error conditions.

csadorf:
Fixes issue #148 .

### DIFF
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -441,9 +441,17 @@ class Job(object):
         This function will attempt to move this instance of job from
         its original project to a different project.
 
-        :param project: The project to move this job to.
-        :type project: :py:class:`~.project.Project`
-        :raises DestinationExistsError: If the job is already initialized in project.
+        :param project:
+            The project to move this job to.
+        :type project:
+            :py:class:`~.project.Project`
+        :raises DestinationExistsError:
+            If the job is already initialized in project.
+        :raises RuntimeError:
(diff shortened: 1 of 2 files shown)


csadorf:
@rosecers This patch improves the error handling of `Job.move()` and should at least give you a more insightful error message for the issue you encountered.

I have not changed the behavior of allowing cross-device moves, simply because they may fail and then you end up with corrupted data spaces. It's much safer to copy the jobs across devices and then manually delete the source. We can possibly automate that in a future release (v1.1), but I don't want to risk adding unsafe behavior one day before our final release.

csadorf:
@vyasr I think the coverage decreased, because I don't test the cross-device error state, but I'm not sure how to mock that as part of the unit tests.

vyasr:
I agree, I think the drop is OK. The raw `else..raise` at the end is also not meaningfully testable, if you knew what other error codes to check then you would add special handling behavior.

signac/contrib/job.py
```diff
@@ -441,16 +441,33 @@ def move(self, project):
...
+        :param project:
+            The project to move this job to.
+        :type project:
+            :py:class:`~.project.Project`
+        :raises DestinationExistsError:
+            If the job is already initialized in project.
+        :raises RuntimeError:
+            If the job is not initialized or the destination is on a different
+            device.
+        :raies OSError:
```
vyasr: raies->raises

csadorf: fixed

PR #150, opened 2019-02-27
Documentation for JSONDict constructor.

csadorf:
I added a docstring to the root-namespace exposed JSONDict class and
moved the `parent` constructor argument to the end of the argument
order.

Having the filename be the first argument is much more natural to users
who would want to use this class stand-alone. All internal calls to this
constructor used `filename=...` syntax, so this change does not require
any other changes to the code base.

### DIFF
--- a/signac/core/jsondict.py
+++ b/signac/core/jsondict.py
@@ -194,11 +194,11 @@ def buffer_reads_writes(buffer_size=DEFAULT_BUFFER_SIZE, force_write=False):
 
 
 class JSONDict(SyncedAttrDict):
-    """A dict-like mapping interface to a persistent JSON-file.
+    """A dict-like mapping interface to a persistent JSON file.
 
     The JSONDict is a :class:`~collections.abc.MutableMapping` and therefore
     behaves similar to a :class:`dict`, but all data is stored persistently in
-    the associated JSON-file on disk.
+    the associated JSON file on disk.
 
     .. code-block:: python
 
@@ -220,7 +220,7 @@ class JSONDict(SyncedAttrDict):
         {'foo': {'bar': False}}
 
     :param filename:
-        The filename of the associated JSON-file on disk.
+        The filename of the associated JSON file on disk.
     :param write_concern:
         Ensure file consistency by writing changes back to a temporary file
         first, before replacing the original file. Default is False.
--- a/signac/core/jsondict.py
+++ b/signac/core/jsondict.py
@@ -202,11 +202,11 @@ class JSONDict(SyncedAttrDict):
 
     .. code-block:: python
 
-            doc = JSONDict('data.json', write_concern=True)
-            doc['foo'] = "bar"
-            assert doc.foo == doc['foo'] == "bar"
-            assert 'foo' in doc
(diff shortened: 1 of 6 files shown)


signac/core/jsondict.py
```diff
@@ -194,8 +194,41 @@ def buffer_reads_writes(buffer_size=DEFAULT_BUFFER_SIZE, force_write=False):
...
-    def __init__(self, parent=None, filename=None, write_concern=False):
+    """A dict-like mapping interface to a persistent JSON-file.
+
+    The JSONDict is a :class:`~collections.abc.MutableMapping` and therefore
+    behaves similar to a :class:`dict`, but all data is stored persistently in
+    the associated JSON-file on disk.
+
+    .. code-block:: python
+
+            doc = JSONDict('data.json', write_concern=True)
```
bdice: Is this indented too far? Maybe 4 instead of 8? (Not sure what the standard is across the code.)

signac/core/jsondict.py
```diff
@@ -194,8 +194,41 @@ def buffer_reads_writes(buffer_size=DEFAULT_BUFFER_SIZE, force_write=False):
...
+            doc['foo'] = "bar"
+            assert doc.foo == doc['foo'] == "bar"
+            assert 'foo' in doc
+            del doc['foo']
+
+    This class allows access to keys both with the slicing and attributes.
```
bdice: This class allows access to values through key indexing or attributes named by keys.

A couple reasons:
1. You're not accessing keys, you're accessing values.
2. Slicing is specific to list-like objects where `[start:stop:step]` syntax applies.

signac/core/jsondict.py
```diff
@@ -194,8 +194,41 @@ def buffer_reads_writes(buffer_size=DEFAULT_BUFFER_SIZE, force_write=False):
...
+    .. code-block:: python
+
+            >>> doc['foo'] = dict(bar=True)
+            >>> doc
+            {'foo': {'bar': True}}
+            >>> doc.foo.bar = False
+            {'foo': {'bar': False}}
+
+    :param filename:
+        The filename of the associated JSON-file on disk.
```
bdice: JSON file (no hyphen)

signac/core/jsondict.py
```diff
@@ -194,8 +194,41 @@ def buffer_reads_writes(buffer_size=DEFAULT_BUFFER_SIZE, force_write=False):
...
+    :param write_concern:
+        Ensure file consistency by writing changes back to a temporary file
+        first, before replacing the original file.
```
bdice: Maybe specify the default is `False`?

signac/core/jsondict.py
csadorf: Thank you for helping me with that formulation. I've been struggling with that for a while now.

signac/core/jsondict.py
csadorf: We use the formulation `HDF5-file` everywhere else, change that, too?

bdice: Yes please!

Issue #153, opened 2019-02-27
deep copy not copying data


berceanu:
```
statepoint = copy.deepcopy(job.sp)
del statepoint['a']
```
results in changes to the workspace folder, i.e. the removing of parameter 'a' from `signac_statepoint.json`.

csadorf:
Thank you for reporting! Which Python version are you running?

csadorf:
@berceanu I'm having trouble reproducing the bug. Can you verify that [this unit test](https://github.com/glotzerlab/signac/pull/154/files#diff-5cbe52c4c7165713f09d3ae38168277eR1394) should trigger the described behavior?

bdice:
This triggers the confusing behavior:
```python
import copy
import os
import signac
pr = signac.init_project('issue_153')
job = pr.open_job({'a': 1})
job.init()
print(job._id)  # 42b7b4f2921788ea14dac5566e6f06d0
copy_sp_bug = copy.deepcopy(job.sp)
copy_sp_bug['a'] = '🐛'
print(job.sp)  # {'a': 1}
print(copy_sp_bug)  # {'a': '🐛'}
print(list(iter(project)))  # [signac.contrib.job.Job(project=signac.contrib.project.Project({'project': 'issue_153', 'project_dir': '/home/bdice/tmp_signac_153', 'workspace_dir': '/home/bdice/tmp_signac_153/workspace'}), statepoint={'a': '🐛'})]
assert job in project  # fails! We have a job handle that doesn't exist in the project anymore...?
```

csadorf:
After looking more closely at this issue, we have come to the conclusion that it actually makes most sense that a copy or a deepcopy will still modify the underlying state point and thus job albeit operating on a different in-memory reference. That is because a state point and a job for that matter are directly tied to a specific location on the file system and that doesn't change just because we have a different reference.

What users actually intend when they create a deepcopy of a state point is to generate a `dict` that is no longer tied to the job and thus the file system. For this, users are advised to use `Job.sp()` or `Job.sp._as_dict()`.

We have fixed the deepcopy behavior of `Job` and added unit tests to ensure that we actually implement the behavior described above in pull request #207 .

To help avoid user confusion, we will need to improve the documentation on how to convert state points to dicts (https://github.com/glotzerlab/signac-docs/issues/39).

PR #154, opened 2019-02-27
Fix/deep copy issue 153

csadorf:
Attempt to fix issue #153 .

### DIFF
--- a/signac/core/synceddict.py
+++ b/signac/core/synceddict.py
@@ -273,6 +273,15 @@ class _SyncedDict(MutableMapping):
     def __str__(self):
         return str(self())
 
+    def __copy__(self, memo):
+        # Create a new version of the object not linked to the original parent
+        return type(self)(initialdata=self._convert_to_dict(self._data.copy()),
+                          parent=self._parent)
+
+    def __deepcopy__(self, memo):
+        # Create a new version of the object not linked to the original parent
+        return type(self)(initialdata=self._convert_to_dict(self._data.copy()))
+
     def _as_dict(self):
         with self._suspend_sync():
             return self._convert_to_dict(self._data.copy())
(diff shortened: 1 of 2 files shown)


vyasr:
We still need to decide what the expected behavior is in this case, correct? Specifically, whether or not the copied statepoint should still be associated with the original file?

csadorf:
> We still need to decide what the expected behavior is in this case, correct? Specifically, whether or not the copied statepoint should still be associated with the original file?

I think a copied state point should be completely independent of its origin. That's how I want to and to tend use this.

vyasr:
OK, so a deep-copied statepoint should essentially behave just like a dictionary that supports attribute-based access, but a shallow copy will still modify the underlying file? I think that's fine, but perhaps we should also update all the docstring warnings added in #154 to clarify that `copy.deepcopy` also works?

vyasr:
@csadorf do you want to finish up this PR, or should I try and find someone else to wrap it up?

csadorf:
> @csadorf do you want to finish up this PR, or should I try and find someone else to wrap it up?

I've been thinking a bit more about this and I believe that there is no optimal solution this. A state point is by definition not just a dict, but a dict associated with a job. Even a deep-copy should not be disassociated with the job. Users should use `_as_dict()` or `__call__()` to generate a dict from the state point.

So, in conclusion, I think we should absolutely implement copy and deepcopy explicitly, but we should not disassociate from the job class.

I'm ok with finishing this myself, but I'd appreciate a brief discussion about this.

vyasr:
Yes, we can hold off until next week and have a brief in-person chat. I tend to agree with you, which is why I wanted to be very clear in the docs if we changed that behavior, so let's finalize next week.

vyasr:
This PR is superseded by #207 and can be closed once that is merged.

csadorf:
I'm going to close this now to avoid confusion.

PR #156, opened 2019-02-28
Fix/schema subset

vyasr:
The schema subsetting needs to be able to loop over the provided `jobs_or_statepoints` once per key.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -50,6 +50,7 @@ Fixed
  - Fix ``Collection.update()`` behavior such that existing documents with identical primary key are updated. Previously, a ``KeyError`` would be raised.
  - Fix issue where the ``Job.move()`` would trigger a confusing ``DestinationExists`` exception when trying to move jobs across devices / file systems.
  - Fix issue that caused failures when the ``python-rapidjson`` package is installed. The ``python-rapidjson`` package is used as the primary JSON-backend when installed.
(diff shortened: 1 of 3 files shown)


csadorf:
I'm afraid this will have to go into the next patch release.

changelog.txt
```diff
@@ -50,6 +50,7 @@ Fixed
  - Fix ``Collection.update()`` behavior such that existing documents with identical primary key are updated. Previously, a ``KeyError`` would be raised.
  - Fix issue where the ``Job.move()`` would trigger a confusing ``DestinationExists`` exception when trying to move jobs across devices / file systems.
  - Fix issue that caused failures when the ``python-rapidjson`` package is installed. The ``python-rapidjson`` package is used as the primary JSON-backend when installed.
+ - Fix issue where schema with multiple keys would subset incorrectly if the list of jobs or statepoints was provided as an iterator rather than an iterable.
```
csadorf: I think this should say "as an iterator rather than a sequence".

csadorf:
Additional minimal example for reproducing the issue: https://gist.github.com/berceanu/80f002aad0714a47bbcacf205949ce42

changelog.txt
csadorf: I just fixed it real quick.

PR #161, opened 2019-02-28
added logic to check if h5py is installed

mikemhenry:
## Description
I tried to add the check if h5py is installed only to tests that absolutely needed h5py.

## Motivation and Context
We should skip tests that need h5py instead of just crashing with an import error. This logic existed in other tests but now exists for all tests (that failed for me).

## Types of Changes
- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

This pull request is based on
- [x] *master*, because it is a bug fix or an update to the documentation.
- [ ] *develop*, because it introduces a new feature.

## Checklist:

If necessary:


tests/test_job.py
```diff
@@ -826,6 +831,7 @@ def test_doc_formatting(self):
         self.assertEqual('{job.document.a.b}'.format(job=job), str(job.doc.a.b))
         self.assertEqual('{job.document[a][b]}'.format(job=job), str(job.doc.a.b))
 
+    @unittest.skipIf(not H5PY, 'test requires the h5py package')
```
csadorf: The fact that this test requires `h5py` is concerning.

csadorf: I just read through the whole function, it actually makes sense.

changelog.txt
```diff
@@ -16,6 +16,10 @@ Highlights
  - Projects and job search results are displayed nicely in Jupyter Notebooks.
  - Support for compressed Collection files.
 
+next
+----
+
+ - Fixed tests that depend on ``h5py`` to only run if ``h5py`` is installed.
```
csadorf: We don't usually announce purely development-related things in the changelog.

csadorf:
I've enabled build of forks on CircleCI, but I'm not able to trigger the build manually. Could you make some tiny commit so that we can see if it works at all?

bdice:
> 
> 
> I've enabled build of forks on CircleCI, but I'm not able to trigger the build manually. Could you make some tiny commit so that we can see if it works at all?

Try `git commit --allow-empty -m "Trigger CI."`

csadorf:
It's possible that this only works for newly created PRs? This is clearly enabled.
screen shot 2019-03-01 at 1 21 12 pm

mikemhenry:
~I had the same issue with CircleCI and pull requests on bitbucket, I was hoping it was a limitation of the bitbucket API and not intrinsic to CircleCI~

Looks like it is working!

mikemhenry:
These failures are interesting and not something I expected, I guess we will see how pypy3 goes, but its weird that the python3.6 test was the only slow one using the CPython implementation

csadorf:
There are non-determistic failures with PyPy3. I don't know what the cause is. 

The benchmark failure was probably just a fluke.

@mikemhenry Could you add an additional test run where `h5py` is not available? To test against dependency leaks? Easiest might be to add these two lines:
``
echo "raise ImportError" > h5py.py
python -m unittest discover ...
```
But there might be better approaches.

mikemhenry:
I'm not exactly sure what you mean, like add "no `h5py`" to the test matrix?

csadorf:
Yes exactly.

mikemhenry:
You know I thought about that, like we should have a test that is the minimal dependencies as defined in `setup.py` Do you think that would be enough? I think there is too many combinations if we wanted to do something like "has `h5py` doesn't have `mpi4py`, but we should test that everything passes when we have just the core deps.

PS it looks like we need to add a `h5py` to the extras https://github.com/glotzerlab/signac/blob/master/setup.py#L45-L48

csadorf:
> You know I thought about that, like we should have a test that is the minimal dependencies as defined in `setup.py` Do you think that would be enough? I think there is too many combinations if we wanted to do something like "has `h5py` doesn't have `mpi4py`, but we should test that everything passes when we have just the core deps.
> 

Yeah, I don't think we need to test all combinations, but we should at least test that signac does not fail with import errors in case that *extra* dependencies are not installed.

> PS it looks like we need to add a `h5py` to the extras https://github.com/glotzerlab/signac/blob/master/setup.py#L45-L48

Could you add an "extra" dependency called `h5`?

mikemhenry:
Can do! That sounds like a good middle ground :)

vyasr:
It looks like we still want to add an additional CircleCI job to test without h5py, are there any other outstanding to-dos on this PR?

csadorf:
@mikemhenry Are you still planning on adding the missing CircleCI jobs or should I merge as is?

PR #167, opened 2019-03-15
Amended signac/contrib/job.py to add option to initialize with replic…

rosecers:
…a indices

## Description
L408: Added parameter "n" to init function, defaulted to None, to represent the number of replicas to initialize
L421-427: Provided n definition and type
L432-440: Indented pre-existing function code to be executed in the n=None default case
L441-446: Provided initialization of replicas in the n!=None case. Note: return statement is not defined, must be defined before merging.

## Motivation and Context
https://github.com/glotzerlab/signac-docs/issues/1

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [X] New feature

This pull request is based on
- [X] *develop*, because it introduces a new feature.

## Checklist:

- [X] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
- [X] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
- [X] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.txt).
- [X] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
- [X] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [N/A] I have updated the API documentation as part of the package doc-strings.
- [N/A] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
- [X] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt).

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -4,6 +4,11 @@ Changelog
 
 The **signac** package follows `semantic versioning <https://semver.org/>`_.
 
+next
+----
+
+Added functionality to initialize multiple replicas of a given statepoint
+
 Version 1.0
 ===========
 
--- a/contributors.txt
+++ b/contributors.txt
@@ -7,3 +7,5 @@ Bradley Dice
 Tim Moore
 Pengji Zhou
 Eric Harper
+Rose Cersonsky
+Yuan Zhou
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -405,7 +405,7 @@ class Job(object):
         except (AssertionError, ValueError):
             raise JobsCorruptedError([self._id])
 
-    def init(self, force=False):
+    def init(self, force=False, n=None):
         """Initialize the job's workspace directory.
 
         This function will do nothing if the directory and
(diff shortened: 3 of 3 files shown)


csadorf:
> This looks fine to me. It would be nice to add a test to check this behavior, but I'll leave that decision to @csadorf.

I think that there are a few questions that need to be addressed before we should consider merging this feature:

1. What is the name of the replica index/seed and how do we control it?
2. What is the expected behavior of the return value of the `init()` function? So far it returns the job itself, that behavior appears undefined now.
3. Considering that this feature extends **core** API, I would like to see a clear comparison of how one would implement the exact same behavior with the current API and the proposed API extension. Is the increased API complexity justified compared to implementing the exact same behavior with the current API?

Especially the last point is important to me. By adding this parameter, we implicitly imply that there is some kind of special relationship between these jobs to novice users. However, that is not the case and may lead to confusion.

bdice:
I share all of the considerations / concerns that @csadorf raises above (hadn't gotten around to writing a comment yet). In particular, the implied "special relationship" among replicas is probably not intuitive to new users.

Alternatives to this approach include:

1. Create a set of utility functions for statepoint grids, since this is a one-dimensional case of grid initialization. I inherited some code from @PaulDodd that does this, which might be a good model.

2. Add examples code snippets to the documentation "recipes" for this specific case, since it takes just two lines. Probably something like:

```python
for i in range(10):
   project.open_job(dict(**main_statepoint, replica_index=i)).init()
```

It could be accompanied with an example of grouping by all keys besides the replica index to compute a mean and standard deviation over some value in the job document.

Thorough tests must be added for this new functionality, especially if the API is changed.

csadorf:
Thank you very much for the contribution! I am going to close this pull request for now since it seems that the concept is not mature enough yet. I am happy to discuss this idea as part of a feature proposal.

PR #174, opened 2019-04-01
Fix/h5py circle ci fail

csadorf:
Fix CIRCLE-CI issues.


bdice:
@csadorf Thanks for breaking this off into a separate PR!

PR #176, opened 2019-04-13
Update pull request template to reflect change of branching model.

csadorf:
Update pull request template for switch to *OneFlow* branching model.


csadorf:
@bdice I've updated the guidelines, alternatively we could remove that section completely. What do you think?

bdice:
@csadorf Thanks! I vote to remove this section, for simplicity. In my opinion, maintainers should decide if something is worth releasing as a hotfix, not the contributor of the fix.

PR #181, opened 2019-04-29
Feature/h5store attrs

csadorf:
Expose the `attrs` property of `h5py.File` for `H5Store` and `H5Group`.

For example:
```
h5s.foo.attrs['parameter'] = 0
```

### DIFF
--- a/signac/core/h5store.py
+++ b/signac/core/h5store.py
@@ -230,6 +230,10 @@ class H5Group(MutableMapping):
             else:
                 return super(H5Group, self).__eq__(other)
 
+    @property
+    def attrs(self):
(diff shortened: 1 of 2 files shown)


bdice (changes requested):
I would like one line of documentation for the function, saying that it is an [AttributeManager](http://docs.h5py.org/en/stable/high/attr.html#AttributeManager) with an appropriate reference to the `h5py` docs.

Out of curiosity, what is your use case for this? Any recommendations on when this might be especially helpful?

csadorf:
It seems like this would unnecessarily complicate the API. Could be revisited in the future.

PR #182, opened 2019-05-01
Improve the representation of `H5Group`.

csadorf:
Implements `H5Group.__repr__`.

For example:

```python\
>>> H5Store('data.h5')['foo']
H5Group(store=H5Store(filename='data.h5'), path='/foo')
```

Improves the REPL experience.

### DIFF
--- a/tests/test_h5store.py
+++ b/tests/test_h5store.py
@@ -200,10 +200,9 @@ class H5StoreTest(BaseH5StoreTest):
     def test_repr(self):
         with self.open_h5store() as h5s:
             key = 'test_repr'
+            self.assertEqual(repr(h5s), repr(eval(repr(h5s))))
             h5s[key] = self.get_testdata()
-            repr(h5s[key])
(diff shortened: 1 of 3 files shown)


tests/test_h5store.py
```diff
@@ -201,6 +201,7 @@ def test_repr(self):
         with self.open_h5store() as h5s:
             key = 'test_repr'
             h5s[key] = self.get_testdata()
+            repr(h5s[key])
```
bdice: In `freud`, I test these `repr`s that correspond to constructor arguments with `assert repr(obj) = repr(eval(repr(obj)))` or something like that. Could be useful as a check.

csadorf: This will only work in combination with `from signac.core.h5store import H5Group` unless I add the whole module path, thoughts?

csadorf: I'll follow the `numpy` example, that does not include the module name for `np.ndarray`.

PR #188, opened 2019-05-19
Implement feature to search jobs with a sp- and doc-integrated filter.

csadorf:
## Description

This patch enables the search for jobs with a statepoint- and document-
combined filter. To specify whether a key is a statepoint- or a
document-key use prefixes 'sp.' and 'doc.' respectively. No prefix is
equivalent to the 'sp.' prefix.

This patch is backwards compatible with the exception that the index
scheme was slightly modified.

Related to issue glotzerlab/signac-dashboard#25 .

## Minimal example

This patch enables queries such as this one:

```python
>>> project.find_jobs({'foo': 0, 'doc.bar': True})
signac.contrib.project.JobsCursor({'project': 'project', 'filter': '{'sp.foo': 0, 'doc.bar': True}'})
>>> # equivalent to:
>>> project.find_jobs({'foo': 0}, {'bar': True})
signac.contrib.project.JobsCursor({'project': 'project', 'filter': '{'sp.foo': 0, 'doc.bar': True}'})
```
Equivalently on the command line:
```bash
$ signac find foo 0 doc.bar true
```
All keys without a prefix are automatically interpreted as having the `sp.` prefix.

## Motivation and Context

It is difficult to impossible to search jobs by filter and document-filter encoded by a simple string. That makes it impossible for example to search properly using signac-dashboard. This patch enables searching with one combined filter, while searching with two filters is still possible

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

## Tasks

- [ ] Add warning about the use of `sp` and `doc` as root-level keys in documents.

### DIFF
--- a/signac/contrib/filterparse.py
+++ b/signac/contrib/filterparse.py
@@ -120,8 +120,23 @@ def _add_prefix(filter, prefix):
                     "The argument to a logical operator must be a sequence (e.g. a list)!")
         elif '.' in key and key.split('.', 1)[0] in ('sp', 'doc'):
             yield key, value
+        elif key in ('sp', 'doc'):
+            yield key, value
         else:
             yield prefix + '.' + key, value
+
+
+def _root_keys(filter):
+    for key, value in filter.items():
+        if key in ('$and', '$or'):
+            assert isinstance(value, (list, tuple))
+            for item in value:
+                for key in _root_keys(item):
+                    yield key
+        elif '.' in key:
+            yield key.split('.', 1)[0]
+        else:
+            yield key
 
 
 def _parse_filter(filter):
--- a/signac/contrib/filterparse.py
+++ b/signac/contrib/filterparse.py
@@ -4,6 +4,7 @@
 from __future__ import print_function
 import sys
 from ..core import json
+from ..common import six
 
 
 def _print_err(msg=None):
@@ -71,17 +72,17 @@ def _cast(x):
 
 def _parse_simple(key, value=None):
     if value is None or value == '!':
-        return {key: {'$exists': True}}
+        return key, {'$exists': True}
     elif _is_json(value):
-        return {key: _parse_json(value)}
+        return key, _parse_json(value)
     elif _is_regex(value):
-        return {key: {'$regex': value[1:-1]}}
+        return key, {'$regex': value[1:-1]}
     elif _is_json(key):
         raise ValueError(
             "Please check your filter arguments. "
             "Using as JSON expression as key is not allowed: '{}'.".format(key))
     else:
-        return {key: _cast(value)}
+        return key, _cast(value)
 
 
 def parse_filter_arg(args, file=sys.stderr):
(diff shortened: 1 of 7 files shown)


vyasr:
I updated your note to point to what I think is the relevant signac-dashboard issue, but please correct it if that's not the one you wanted.

csadorf:
> I updated your note to point to what I think is the relevant signac-dashboard issue, but please correct it if that's not the one you wanted.

Thx, I've edited it slightly to say "related to" instead of "resolves", because dashboard might still need a minor update to work with this properly (at least we would need to test that).

signac/contrib/filterparse.py
```diff
@@ -91,14 +92,65 @@ def parse_filter_arg(args, file=sys.stderr):
...
-                value = args[i+1]
-            else:
-                value = None
-            q.update(_parse_simple(key, value))
+        q = dict(parse_simple(args))
+
         return _with_message(q, file)
+
+
+def parse_simple(tokens):
```
vyasr: I would prefer to keep this "private" by renaming to `_parse_simple` and renaming the current `_parse_simple` to `_parse_simple_single` or so.

signac/contrib/filterparse.py
```diff
@@ -91,14 +92,65 @@ def parse_filter_arg(args, file=sys.stderr):
...
+        if key in ('$and', '$or'):
+            if isinstance(value, list) or isinstance(value, tuple):
+                yield key, [dict(_add_prefix(item.items(), prefix)) for item in value]
+            else:
+                raise ValueError(
+                    "The argument to a logical operator must be a sequence (e.g. a list)!")
+        elif '.' in key and key.split('.', 1)[0] in ('sp', 'doc'):
+            yield key, value
+        elif key in ('sp', 'doc'):
+            yield key, value
```
vyasr: What is the purpose of this branch?

On a related note, should we explicitly disallow `sp` and `doc` as statepoint/document keys in addition to dots in keys?

signac/contrib/project.py
```diff
@@ -1306,7 +1295,7 @@ def _read_cache(self):
             return cache
 
     def index(self, formats=None, depth=0,
-              skip_errors=False, include_job_document=True):
+              skip_errors=False, include_job_document=True, **kwargs):
```
vyasr: Do we actually need to support arbitrary kwargs here?

vyasr (changes requested):

signac/contrib/filterparse.py
csadorf: I'm ok with the name change, however, I was actually hoping to expose this function publicly, to make it easier to implement custom user scripts like this:
```python
import click

@click.command()
@click.argument('filter', nargs=-1)
def cli(filter):
    jobs = project.find_jobs(parse_simple(filter))

cli()
```

signac/contrib/filterparse.py
csadorf: For these cases, we need to ensure to __not__ add 'sp' or 'doc' as prefixes, for instance with a filter that is expressed like this: `{"sp": {"foo": 0}}`.

Concerning your second point, `sp` and `doc` are actually valid keys, you would need to "escape" them like this: ``{"sp.sp.foo": 0}``. However, I agree that we should probably discourage that, possibly even with warnings, and make sure to explicitly test the correctness of the behavior in that case.

signac/contrib/project.py
csadorf: No, that is an error. I was trying to introduce backwards compatibility here, and I think we should probably still do that in the long run, but as of right now, the `kwargs` are unnecessary.

csadorf (commented):
Thx for your review @vyasr!

signac/contrib/filterparse.py
bdice: I support making this a public method / official API - if it's not public then `signac-dashboard` has to rely on a private method.

vyasr: Got it, in that case I support making it part of the API as well. I still like renaming the `_parse_simple` method to indicate that it operates on one filter component at a time, but we don't have to change the `parse_simple` method.

signac/contrib/filterparse.py
vyasr: I didn't consider that using `sp` and `doc` as root-level keys in a standard query format was an option, but in hindsight that makes sense. I support discouraging the usage of those as keys to future-proof ourselves against future changes where that might cause problems, though.

csadorf:
I'm going to close this until we work on 1.3.

vyasr:
@vishav1771 please comment on here and let me know if you're OK with switching to work on this for now (regarding my comments from #309).

vishav1771:
@vyasr That's okay. 👍

vishav1771:
This is moved to #332 .

PR #193, opened 2019-05-19
Fix and improve repr of Project, Job, and JobsCursor.

csadorf:
Extend unit tests to check for identify of `eval(repr(x))` of affected
classes.

### DIFF
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -160,10 +160,9 @@ class Project(object):
         return str(self.get_id())
 
     def __repr__(self):
-        return "{type}({{'project': '{id}', 'project_dir': '{rd}',"\
-               " 'workspace_dir': '{wd}'}})".format(
-                   type=self.__class__.__module__ + '.' + self.__class__.__name__,
-                   id=self.get_id(), rd=self.root_directory(), wd=self.workspace())
+        return "{type}.get_project('{root}')".format(
+                   type=self.__class__.__name__,
+                   root=self.root_directory())
 
     def _repr_html_(self):
         return repr(self) + self.find_jobs()._repr_html_jobs()
@@ -1607,6 +1606,10 @@ class JobsCursor(object):
         # next() method for this class.
         self._next_iter = None
 
+    def __eq__(self, other):
+        return self._project == other._project and self._filter == other._filter\
+            and self._doc_filter == other._doc_filter
+
     def __len__(self):
         # Highly performance critical code path!!
         if self._filter or self._doc_filter:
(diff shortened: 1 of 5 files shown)


csadorf:
Any other `__repr__()` functions we might want to revise?

vyasr:
Collections would be nice. Also, I've personally found it awkward that `__str__` and `__repr__` are different for schemas, especially when I'm working in REPL environments and just want to see the schema by typing `project.detect_schema()`, but I don't know how everyone else feels about that. Maybe the thought was that it would be too verbose a default in those scenarios?

csadorf:
> Collections would be nice. Also, I've personally found it awkward that `__str__` and `__repr__` are different for schemas, especially when I'm working in REPL environments and just want to see the schema by typing `project.detect_schema()`, but I don't know how everyone else feels about that. Maybe the thought was that it would be too verbose a default in those scenarios?

Would you want the `__repr__` of Collections to be identical to `__str__`: `"<Collection file=<_io.TextIOWrapper name='collection.txt' encoding='utf-8'>>"` ?

I'd prefer the `ProjectSchema` representation to remain as compact as it is right now, but I'm open to suggestions.

vyasr:
Hmm, is there a meaningful representation of the document index itself? I suppose the index for a collection is not as information rich as a project index, I was hoping to get something more like the output of `__str__` for a schema but I don't think that's possible. Perhaps an overview of file paths included or something, or even just mimicking the schema's `__repr__` method?

I just typically find myself writing `print(project.detect_schema())` any time I perform schema detection (or saving and then printing), and if others experience that too then I think we're maybe complicating the usage a little bit. If that's just me, though, then I'm fine keeping it as is if you prefer that.

csadorf:
> Hmm, is there a meaningful representation of the document index itself? I suppose the index for a collection is not as information rich as a project index, I was hoping to get something more like the output of `__str__` for a schema but I don't think that's possible. Perhaps an overview of file paths included or something, or even just mimicking the schema's `__repr__` method?
> 
> I just typically find myself writing `print(project.detect_schema())` any time I perform schema detection (or saving and then printing), and if others experience that too then I think we're maybe complicating the usage a little bit. If that's just me, though, then I'm fine keeping it as is if you prefer that.

Can you create minimal demo implementations of what you would prefer? That would make the discussion a bit more concrete.

When you do, please follow guidelines outlined [here](https://stackoverflow.com/a/2626364), unless you disagree with them.

signac/contrib/project.py
```diff
@@ -160,10 +160,9 @@ def __str__(self):
         return str(self.get_id())
 
     def __repr__(self):
-        return "{type}({{'project': '{id}', 'project_dir': '{rd}',"\
-               " 'workspace_dir': '{wd}'}})".format(
-                   type=self.__class__.__module__ + '.' + self.__class__.__name__,
-                   id=self.get_id(), rd=self.root_directory(), wd=self.workspace())
+        return "{type}.get_project('{root}')".format(
+                   type=self.__class__.__name__,
+                   root=self.root_directory())
```
bdice: I feel like this is less useful than the previous behavior...

csadorf: Ok, I found the previous representation a bit too verbose and slightly redundant. But we can revert that change.

bdice: Solution: we'll keep this `repr` as-is and add the other info to the `_repr_html_` method.

Issue #134, opened 2019-02-22
The signac shell interpreter should keep a command history


csadorf:
### Feature description

The signac shell interpreter should pertain a command history on a per-project basis.
Invoking `signac shell` should therefore auto-complete commands based on previous sessions in the same project.

### Proposed solution

We are already using the `readline` module to setup auto-completion, I think it's just a matter of saving the command history as described [here](https://docs.python.org/3.6/library/readline.html#example).

csadorf:
Resolved with merge of PR #194 .

PR #194, opened 2019-05-19
Keep signac shell command history on a per-project basis.

csadorf:
Resolves issue #134.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -17,6 +17,11 @@ Highlights
  - Support for compressed Collection files.
 
 
+next
+----
+
+ - Keep signac shell command history on a per-project basis.
+
 [1.1.0] -- 2019-05-19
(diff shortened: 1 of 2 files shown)


csadorf:
Not sure why the coverage decreased. Maybe, because I'm skipping the history code segment on PyPy?

vyasr:
It looks like the coverage tracking as a whole doesn't recognize that `test_shell.py` is actually calling the `__main__.py` functions, the package must not know how to deal with entry points. The entire `__main__.py` has [0% coverage](https://codecov.io/gh/glotzerlab/signac/compare/cbe589d155c505e3c892fca63db173e627e9c1b2...b9d2226737794b15c2bc82335b1b45426bcee218/tree/signac). I wouldn't worry about it.

PR #198, opened 2019-05-26
Use synchronization and conversion for _SyncedList as well as _SyncedDict

### LINKED ISSUE #196: _SyncedList dictionary children should be _SyncedDicts
### Description

I was editing a job document. When getting a list that is a child of a `_SyncedDict`, you get a `_SyncedList`. However, when getting a dict that is a child of a `_SyncedList`, you don't get a `_SyncedDict` (just a `dict`).

### To reproduce

```python
import signac

project.doc['key'] = [{'a': 1}, {'b': 2}]
print(type(project.doc.key))
print(type(project.doc.key[0]))

project.doc.key[0]['a'] = 3
assert project.doc.key[0]['a'] == 3  # still 1
```

bdice:
## Description
In order to fix #196, I needed to apply the same kinds of conversion and parent synchronization logic to `_SyncedList` that we already apply to `_SyncedDict`. To do so, I created a parent class `_SyncedCollection`, following the hierarchy defined in [collections.abc](https://docs.python.org/3/library/collections.abc.html).

Currently, there is some problem with buffering that I haven't figured out yet (see the test status). @csadorf Could you lend me a hand with this problem? I looked at it for a while but wasn't able to figure out why the data wasn't saving when exiting the `with jsd2.buffered() as b2:` context manager in the tests.

I plan to add more extensive tests for this and ensure ~100% coverage of the new code, once the buffering issue is resolved.

In theory, these changes would also allow for a `JSONList` class that works in a similar way to `JSONDict` -- but that's outside the scope of this PR.

## Motivation and Context
Resolves #196.

## Types of Changes

- [x] Bug fix

## Checklist:

If necessary:

### DIFF
--- a/signac/core/synceddict.py
+++ b/signac/core/synceddict.py
@@ -4,17 +4,22 @@
 "Synchronized dictionary."
 import logging
 from contextlib import contextmanager
-from functools import wraps
 from copy import deepcopy
 
 from ..common import six
 
 if six.PY2:
+    from collections import Collection
     from collections import Mapping
+    from collections import Sequence
     from collections import MutableMapping
+    from collections import MutableSequence
 else:
+    from collections.abc import Collection
     from collections.abc import Mapping
+    from collections.abc import Sequence
     from collections.abc import MutableMapping
+    from collections.abc import MutableSequence
 try:
     import numpy
     NUMPY = True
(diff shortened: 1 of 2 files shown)


signac/core/synceddict.py
```diff
@@ -101,16 +55,19 @@ def _validate_key(key):
...
-        if type(root) is type(self):
+        if isinstance(root, _SyncedDict):
             for k in root:
                 root[k] = self._dfs_convert(root[k])
+        elif isinstance(root, _SyncedList):
+            for i, k in enumerate(root):
+                root[i] = self._dfs_convert(k)
         elif isinstance(root, Mapping):
-            ret = type(self)(parent=self)
+            ret = _SyncedDict(parent=self)
```
csadorf: I believe the reason that unit tests are failing is because the `_SyncedDict` is a mixed-in class. Other classes such as `JSONDict` inherit from this class. It is therefore important to return `type(self)(...)` and not `_SyncedDict(...)` here.

I am not sure whether the same holds for the `_SyncedList` block, we may need a unit test that explicitly tests inheritance.

csadorf (commented):
I think I've identified the problem. I also added a few other suggestions.

signac/core/synceddict.py
bdice: I tried swapping that out for `type(self)` but it didn't seem to fix the problem.

signac/core/synceddict.py
```diff
@@ -148,20 +105,32 @@ def _save(self):
...
+            for index, item in enumerate(new):
+                if index < len(old):
+                    if old[index] == new[index]:
+                        continue
+                    elif isinstance(new[index], Collection) and isinstance(old[index], Collection):
+                        self._dfs_update(old[index], new[index])
+                        continue
+                old[index] = self._dfs_convert(new[index])
+            for index in range(len(new), len(old)):
+                del old[index]
```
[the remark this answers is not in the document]
bdice: I tried that earlier and it failed because it was calling `_dfs_update` recursively on strings (which are treated as an instance of `Collection`). I attempted to stop the recursion if `old[key]` or `new[key]` are instances of `six.string_types` but that seems to cause data not to be updated.

signac/core/synceddict.py
```diff
@@ -101,16 +55,19 @@ def _validate_key(key):
         return key
 
     def _dfs_convert(self, root):
-        if type(root) is type(self):
+        if isinstance(root, _SyncedDict):
```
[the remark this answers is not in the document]
bdice: I will handle this later (after we establish the code's correctness, as you mentioned).

signac/core/synceddict.py
csadorf: Are you sure that this is triggered at all and if yes, can you check which types are triggering it?

bdice: `type(root)` is `dict`, `type(self)` is `signac.core.attrdict.SyncedAttrDict`

bdice:
To get further on this PR, I'll probably need to invest significant time or get assistance. I am still eager to work on it, but I've lost my "train of thought" about how these internal pieces work, and I will need to re-acquaint myself.

csadorf:
> To get further on this PR, I'll probably need to invest significant time or get assistance. I am still eager to work on it, but I've lost my "train of thought" about how these internal pieces work, and I will need to re-acquaint myself.

Ok, let's pick this up after the release of v1.1.1 then.

csadorf:
Closing this for now for lack of progress.

vyasr:
This PR is replaced by #484.

PR #200, opened 2019-06-04
Feature/issue 104 implement json

xadams:
## Description
This change introduces the read_json and to_json methods to the Collection class for easy conversion between Collections and JSON strings

## Motivation and Context
This change addresses the issue raised in: https://github.com/glotzerlab/signac/issues/104#issuecomment-472898475

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

I was unsure if and how to update the API documentation, but believe I have covered the remaining points from the Contributing Guidelines. Thanks in advance for reviewing this!


tests/test_collection.py
```diff
@@ -614,6 +614,28 @@ def test_read(self):
...
+class CollectionTestToFromJson(unittest.TestCase):
+
+    def setUp(self):
+        self._tmp_dir = TemporaryDirectory(prefix='signac_collection_')
+        self._fn_collection = os.path.join(self._tmp_dir.name, 'test.txt')
+        self._fn_json = os.path.join(self._tmp_dir.name, 'test.json')
+        self.addCleanup(self._tmp_dir.cleanup)
+        with open(self._fn_collection, 'w') as file:
+            for i in range(3):
+                file.write('{"a": 0}\n')
```
bdice: I'd prefer to see the collection built using the normal Collection API for doing this, instead of manually writing the lines. We can rely on that behavior to work, and then this test is specifically focused on the JSON reading/writing.

bdice:
@xadams Overall this looks good! I see there are a lot of changes to whitespace - did you format the code somehow? It looks like the changes are flake8-compliant but it made it a little confusing to read the diff.

tests/test_collection.py
xadams: In there an existing test I can consult for making this change? I was basing this on the FileCollectionBadJson test as this style is different from what I'm accustomed to.

xadams:
Thank you both for the comments! The whitespace was done automatically by my IDE and should now be removed. I wouldn't necessarily trust IntelliJ's built in style guide as its imposed line length at least is outdated according to several forums.

tests/test_collection.py
bdice: This might work (untested):

```python
    def setUp(self):
        self.c = Collection.open(filename=':memory:')
        docs = [dict(_id=str(i)) for i in range(10)]
        self.c.update(docs)
        self.c.flush()

   def test_write_and_read(self):
        self.c.to_json(self._fn_json)
        self.assertGreater(os.path.getsize(self._fn_json), 0)
        c = Collection.read_json(self._fn_json)
        self.assertEqual(len(list(c)), 10)
        self.assertEqual(len(c.find()), 10)
        c.close()
```

signac/contrib/collection.py
```diff
@@ -938,6 +938,26 @@ def dump(self, file=sys.stdout):
         else:
             self._dump(file)
 
+    def to_json(self, file=None):
+        json_string = json.dumps([i for i in self.find()])
```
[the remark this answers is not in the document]
xadams: It appears to work! I've pushed this change.

bdice:
Forgot to mention this earlier (wrote a comment and didn't submit it) - the `to_json` and `from_json` methods are missing docstrings. Those would be good to add.

signac/contrib/collection.py
```diff
@@ -938,6 +938,35 @@ def dump(self, file=sys.stdout):
         else:
             self._dump(file)
 
+    def to_json(self, file=None):
+        """Dump the collection as a JSON file.
+
+        :param file: The file to write the JSON string to
```
vyasr: @csadorf @bdice do we have a standard for docstrings for functions that sometimes (but don't always) return values?

csadorf: No, we don't.

csadorf:
@xadams Thank you very much for this contribution? Do you have an ORCID that I could add to the `contributors.yaml` file?

PR #203, opened 2019-07-02
Fix issue in DictManager that would lead to incorrectly listed dicts.

csadorf:
## Description

The regex was matching files that included the suffix, but did not
necessarily *end* with the given suffix.

## Motivation and Context

The `DictManager` would list dicts with filenames that do not actually end with the given suffix. For example, `mystore.h5` would be listed, but also `mystore.5h.lock`.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/core/dict_manager.py
+++ b/signac/core/dict_manager.py
@@ -108,7 +108,7 @@ class DictManager(object):
 
     def __iter__(self):
         for fn in os.listdir(self.prefix):
-            m = re.fullmatch('(.*){}'.format(self.suffix), fn)
+            m = re.match('^(.*){}$'.format(self.suffix), fn)
             if m:
                 yield m.groups()[0]
 
--- a/signac/core/dict_manager.py
+++ b/signac/core/dict_manager.py
@@ -108,7 +108,7 @@ class DictManager(object):
 
     def __iter__(self):
         for fn in os.listdir(self.prefix):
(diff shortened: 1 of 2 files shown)


bdice:
@csadorf Good fix on the `^...$` syntax -- I was thinking about that a little more after I approved the PR. 😄

signac/core/dict_manager.py
```diff
@@ -108,7 +108,7 @@ def __delattr__(self, name):
 
     def __iter__(self):
         for fn in os.listdir(self.prefix):
-            m = re.fullmatch('(.*){}'.format(self.suffix), fn)
+            m = re.match('^(.*){}$'.format(self.suffix), fn)
```
b-butler: I think using the `^(.*){}$` syntax is much nicer.

b-butler: As @bdice said. Good choice.

csadorf: It's also necessary as long as we support Python 2.7, since `re.fullmatch()` is Py3 only.

b-butler: Ah, I am not familiar with that, though I assumed something like that with the commit text. Regardless, I also think it is more clear about the purpose of the code whether for python 2 or 3.

Issue #168, opened 2019-03-20
Can't assign dictionaries containing integer keys into job.document


bdice:
### Description

I wanted to store a key-value mapping of the form `{1: 'a', 2: 'b'}` into a sub-key of a job document. I incorrectly expected that this would be encodable as JSON. I researched it and found out that JSON only allows string keys. However, the error triggers when checking for dots, and this doesn't seem like the right place. A second problem with silent failure is also confusing.

### To reproduce

First problem: error triggers for unexpected reason.
```python
>>> job.doc['mapping'] = {1: 'a', 2: 'b'}
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File ".../lib/python3.6/site-packages/signac/core/synceddict.py", line 185, in __setitem__
    self._data[self._validate_key(key)] = self._dfs_convert(value)
  File ".../lib/python3.6/site-packages/signac/core/synceddict.py", line 106, in _dfs_convert
    ret[k] = root[k]
  File ".../lib/python3.6/site-packages/signac/core/synceddict.py", line 185, in __setitem__
    self._data[self._validate_key(key)] = self._dfs_convert(value)
  File ".../lib/python3.6/site-packages/signac/core/synceddict.py", line 90, in _validate_key
    if '.' in key:
TypeError: argument of type 'int' is not iterable
```
Second problem: silent failure should be loud.
```python
>>> job.doc = {1: 'a', 2: 'b'}
>>> job.doc
{}
```

### System configuration

Please complete the following information:

 - Operating System: Ubuntu 18.04 (Bash on Windows)
 - Version of Python: 3.6.8
 - Version of signac: 1.0.0

bdice:
For comparison, [tinydb](https://github.com/msiemens/tinydb/) converts non-string keys into strings by [just using `json.dumps`](https://github.com/msiemens/tinydb/blob/50eee49e392e3be32597c945a62f6210e166d9b2/tinydb/storages.py#L112). This behavior may be preferable to getting errors / data loss, perhaps accompanied by a warning.
```python
>>> json.dumps({1: "a", 2: "b"})
'{"1": "a", "2": "b"}'
```

csadorf:
The behavior that you are seen is unintended and especially your second minimal example is surprising to me. But I don't think that explicit conversion is a good idea.

vyasr:
[...]

Issue #170, opened 2019-03-28
Improve "dots in keys" error message.


bdice:
### Feature description

If there's a dot in a key, the error message that shows up makes it really hard to find _where_ the problematic key is located. @rosecers reported challenges with her project because the invalid key was difficult to trace. (Ultimately @csadorf located the invalid key in the project document, not the job statepoints/documents where one might have expected to find it.)

### Proposed solution

We could edit the `_validate_key` method (https://github.com/glotzerlab/signac/blob/88ede71b9d16a61c92f43f05c9662ef6725518b2/signac/core/synceddict.py#L92), so that the error also prints the key.

Something like: `The use of '.' (dots) in keys is invalid: {key}`

Ideally we could show a path to the problematic JSON file, but the `SyncedDict` implementing the validation method is a parent class of the `JSONDict` so it may require a little more thought / planning to show that.

csadorf:
Fixed with merge of #205.

PR #205, opened 2019-07-08
Fix/issue 168

csadorf:
## Description

Fixes issue #168 and #170 .

## Motivation and Context

This patch refactors the state point management with in the `Job` class and also adds a general key type check and conversion to the SyncedDict class. That means that the `Job.sp`, `Job.doc`, and `Project.doc` attributes and their aliases should now behave consistently with respect to their key conversion and validation behavior.

Valid key types are now explicitly listed in the source code, namely: `str`, `int`, `bool`, and `NoneType`.
Trying to use a key of different type should trigger a `KeyTypeError`.

There are two minor incompatibilities introduced with this patch:

1) I discovered a bug where a value of type `tuple` would not be immediately converted to a list when assigned to a state point key and remain a tuple. The value would be a list after reloading the job. The unit test actually expected this odd behavior, I therefore changed a unit test with respect to this bug.
2) Instead of raising an `InvalidKeyError`, which is a `ValueError`, trying to use a floating point number as a key will now raise an `KeyTypeError`, which is a `TypeError`. Using floats as keys was already illegal, since they must contain dots in their JSON representation. However, this would have triggered a `InvalidKeyError`, instead of a `KeyTypeError`.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -152,7 +152,7 @@ class Job(object):
             else:
                 raise
         # Update this instance
-        self._statepoint = SyncedAttrDict(dst._statepoint(), parent=_sp_save_hook(self))
+        self._statepoint = SyncedAttrDict(dst._statepoint._as_dict(), parent=_sp_save_hook(self))
         self._id = dst._id
         self._wd = dst._wd
         self._fn_doc = dst._fn_doc
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -24,6 +24,10 @@ else:
 logger = logging.getLogger(__name__)
 
 
+class SPDict(SyncedAttrDict):
+    pass
+
+
 class _sp_save_hook(object):
 
     def __init__(self, job):
@@ -60,14 +64,11 @@ class Job(object):
 
         # Ensure that the job id is configured
         if _id is None:
-            self._statepoint = json.loads(json.dumps(statepoint))
-            self._id = calc_id(self._statepoint)
+            self._statepoint = SPDict(statepoint, parent=_sp_save_hook(self))
+            self._id = calc_id(self._statepoint())
         else:
-            self._statepoint = dict(statepoint)
+            self._statepoint = SPDict(statepoint, parent=_sp_save_hook(self))
             self._id = _id
-
-        # Prepare job statepoint
-        self._sp = SyncedAttrDict(self._statepoint, parent=_sp_save_hook(self))
 
         # Prepare job working directory
         self._wd = os.path.join(project.workspace(), self._id)
@@ -155,9 +156,8 @@ class Job(object):
             else:
                 raise
         # Update this instance
-        self._statepoint = dst._statepoint
+        self._statepoint = SPDict(dst._statepoint(), parent=_sp_save_hook(self))
         self._id = dst._id
-        self._sp = SyncedAttrDict(self._statepoint, parent=_sp_save_hook(self))
         self._wd = dst._wd
         self._fn_doc = dst._fn_doc
         self._document = None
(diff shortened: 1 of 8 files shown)


signac/contrib/job.py
```diff
@@ -214,42 +210,30 @@ def statepoint(self):
...
-            As with :attr:`Job.statepoint`, use `job.sp()` instead of
-            `job.sp` if you need a deep copy that will not modify the
-            underlying persistent JSON file.
-        """
-        return self.statepoint
-
-    @sp.setter
-    def sp(self, new_sp):
-        self.statepoint = new_sp
+    sp = statepoint  # alias
```
bdice: Will this still appear in the documentation? Why not do the same thing to `doc` / `document`?

signac/contrib/project.py
```diff
@@ -372,7 +372,7 @@ def open_job(self, statepoint=None, id=None):
             # second best case
             job = self.Job(project=self, statepoint=statepoint)
             if job._id not in self._sp_cache:
-                self._sp_cache[job._id] = dict(job._statepoint)
+                self._sp_cache[job._id] = job._statepoint._as_dict()
```
bdice: Is there a reason to use the `__call__` syntax in  `self._statepoint = SyncedAttrDict(dst._statepoint(), parent=_sp_save_hook(self))` and `_as_dict()` here? If I recall correctly, both do the same thing. I don't think I have a preference, just trying to understand for consistency.

csadorf: I needed to look this up. The `__call__()` function will also trigger a load-operation. So using `_as_dict()` would be correct here.

signac/contrib/job.py
csadorf: Yes, but I've reverted this, so that we can give it a distinct doc-string.

csadorf:
@bdice I think I've addressed all of your comments.

signac/contrib/project.py
vyasr: I might suggest using `job.statepoint` here for future proofing against any change to internals.

csadorf: I think I used `._statepoint` to avoid the extra function call caused by the `@property`, but hopefully that does not really make any difference.

signac/contrib/job.py
```diff
@@ -241,15 +228,19 @@ def sp(self, new_sp):
...
+
+            dirname, filename = os.path.split(self._fn_doc)
+            fn_tmp = os.path.join(dirname, '._{uid}_{fn}'.format(
+                uid=uuid.uuid4(), fn=filename))
+            with open(fn_tmp, 'wb') as tmpfile:
+                tmpfile.write(json.dumps(new_doc).encode())
+            if six.PY2:
+                os.rename(fn_tmp, self._fn_doc)
+            else:
+                os.replace(fn_tmp, self._fn_doc)
```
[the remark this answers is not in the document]
csadorf: Yes, we probably should. I'm going to update that.

csadorf:
* [ ] Move all `Job.doc` and `Project.doc` IO logic into the `JSONDict `class.

csadorf:
@bdice @vyasr Since this patch is touching core logic, I'd appreciate if you could give this PR another pass.

@tcmoore3 I've followed your suggestion and refactored the reset-document logic into the `JSONDict` class. This also means that the buffered mode will be respected for document resets.

csadorf:
@vyasr I've merged this branch so that we can move forward, but maybe you can still have a look at this. We can still push a fix if you find any issue.

PR #207, opened 2019-07-14
Fix/deep copy issue 153 based on issue 168

csadorf:
Implement well-defined copy and deepcopy behavior for instances of `Job`.

This patch is based on the code base of #205 .

Fixes issue #153 .

### DIFF
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -152,7 +152,7 @@ class Job(object):
             else:
                 raise
         # Update this instance
-        self._statepoint = SyncedAttrDict(dst._statepoint(), parent=_sp_save_hook(self))
+        self._statepoint = SyncedAttrDict(dst._statepoint._as_dict(), parent=_sp_save_hook(self))
         self._id = dst._id
         self._wd = dst._wd
         self._fn_doc = dst._fn_doc
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -24,6 +24,10 @@ else:
 logger = logging.getLogger(__name__)
 
 
+class SPDict(SyncedAttrDict):
+    pass
+
+
 class _sp_save_hook(object):
 
     def __init__(self, job):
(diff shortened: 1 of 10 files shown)


csadorf:
Note: Merge after #205 .

csadorf:
@vyasr This patch has the unit tests from #154 , but in addition has unit tests for the copy/deepcopy behavior of instances of `Job` and a patch of the `Job` class that fixes the `copy` behavior. I suggest we merge this PR instead of #154 .

signac/contrib/job.py
```diff
@@ -144,7 +146,7 @@ def reset_statepoint(self, new_statepoint):
             else:
                 raise
         # Update this instance
-        self._statepoint = SyncedAttrDict(dst._statepoint._as_dict(), parent=_sp_save_hook(self))
+        self._statepoint._data = dst._statepoint._data
```
vyasr: To make sure I'm understanding, is doing this enforcing that the new statepoint is tied to the exact same file as before by reusing the underlying dict stored in _data?

signac/contrib/job.py
```diff
@@ -22,14 +23,15 @@
 
 class _sp_save_hook(object):
 
-    def __init__(self, job):
-        self.job = job
+    def __init__(self, *jobs):
+        self.jobs = list(jobs)
```
vyasr: Does this change belong in this branch? It seems like a better fit for the grouping PR.

vyasr: Ah I see, this is necessary for enabling pickling? Because of the way that the state is being tracked below?

signac/contrib/job.py
csadorf: The job's state point is tied to the manifest file through the hook class within the `Job` class constructor. This line is merely copying the in-memory structure. No need to rebuild that.

signac/contrib/job.py
csadorf: No, this change is necessary, because we need to be able to add multiple jobs to the same state point hook. A shallow copy of a job will share the same hook.

csadorf:
@vyasr I've answered your questions.

signac/contrib/job.py
vyasr: Ah right, the important thing is maintaining the same parent, which is preserved in `_data`. Thanks.

csadorf:
> I like the current unittests and implementation.
> 
> From my understanding to make sure I have it correct is that copy effectively links changes to all copies while deepcopy does not propagate changes in statepoint to other objects that have references to it. However, all copies of either type point to the same underlying file unless a deepcopied object or the original job object is changed at which point only the changed one still points to an underlying file.
> 
> This is the behavior that is implied by the tests.

That's exactly correct.

PR #220, opened 2019-08-16
added notes and deprecations throughout project

tommy-waltmann:
I added @deprecated annotations, removed modules, and made notes about parts of the code that are not necessary.

In addition, in lower level namespaces like contrib/filesystems.py I could not import the __version__ variable from the highest level namespace. I had to create another __version__ variable in each file where I needed it. This is bad practice and will have to be removed at some point. 

## Motivation and Context

As we move to a more polished version of the code, we must remove methods/modules/classes/etc. that are not necessary or are confusing to users.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

- [x ] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
- [x ] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
- [ x] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
- [ x] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:


signac/contrib/project.py
```diff
@@ -1177,6 +1191,7 @@ def repair(self, fn_statepoints=None, index=None, job_ids=None):
             raise JobsCorruptedError(corrupted)
 
     def _sp_index(self):
+        "Update and return the statepoint index cache"
```
vyasr: For convention's sake (and future proofing against extending the docstring), I would triple quote this.

signac/contrib/project.py
```diff
@@ -1187,7 +1202,7 @@ def _sp_index(self):
         return self._index_cache.values()
 
     def _build_index(self, include_job_document=False):
-        "Return a basic state point index."
+        "Generate a basic state point index."
```
vyasr: Same as above (triple quote).

signac/contrib/project.py
```diff
@@ -730,6 +736,8 @@ def _get_statepoint_from_workspace(self, jobid):
                 raise JobsCorruptedError([jobid])
             raise KeyError(jobid)
 
+    @deprecated(deprecated_in="1.3", removed_in="2.0", current_version=__version__,
+                details="Use open_job(jobid).statepoint() function instead.")
```
vyasr: The details should read `Use open_job(id=jobid).statepoint()...`.

signac/contrib/project.py
```diff
@@ -506,6 +510,8 @@ def detect_schema(self, exclude_const=False, subset=None, index=None):
         statepoint_index = self.build_job_statepoint_index(exclude_const=exclude_const, index=index)
         return ProjectSchema.detect(statepoint_index)
 
+    @deprecated(deprecated_in="1.3", removed_in="2.0", current_version=__version__,
+                details="Use find_jobs().ids instead.")
     def find_job_ids(self, filter=None, doc_filter=None, index=None):
         """Find the job_ids of all jobs matching the filters.
```
vyasr: If we're deprecating both `build_job_search_index` and `find_job_ids`, can we get rid of the `JobSearchIndex` class as well? Users who really want to use indices should probably just use Collections anyway.

signac/contrib/project.py
```diff
@@ -434,6 +436,8 @@ def build_job_search_index(self, index, _trust=False):
         """
         return JobSearchIndex(index=index, _trust=_trust)
 
+    @deprecated(deprecated_in="1.3", removed_in="2.0", current_version=__version__,
+                details="Use the detect_schema function instead.")
```
vyasr: For consistency with the other deprecation details strings, add `()` to `detect_schema`.

vyasr:
@tommy-waltmann thanks for the contribution! Before making any more changes, you should probably sync up with master and then fix style issues. You can see that the style-check CI job failed; you can test that locally by running `flake8`. For more information, see the [docs on contributing to signac](https://docs.signac.io/projects/core/en/latest/support.html?highlight=flake8#set-up-a-development-environment). Note that since you've already made some commits with style issues, you may have to run flake8 manually to find the issues, but going forward setting up the commit hook can help save you time by catching style issues before you commit.

signac/contrib/project.py
tommy-waltmann: Yes, I think that would be a good idea, although let's check with others before we do something this drastic.

bdice:
Looking at the documentation for `deprecation`, I think we might not have to specify `current_version` at all: https://deprecation.readthedocs.io/en/latest/#deprecation.deprecated

When adding a `@deprecated` flag, we usually want this to trigger a DeprecationWarning until it is actually removed, so the `current_version` is probably not necessary. This is what @zhou-pj did in https://github.com/glotzerlab/signac-flow/pull/148. By ignoring `current_version`, we could ignore some of the challenges with getting the package version to be read by other files in the package.

csadorf:
> Looking at the documentation for `deprecation`, I think we might not have to specify `current_version` at all: https://deprecation.readthedocs.io/en/latest/#deprecation.deprecated
> 
> When adding a `@deprecated` flag, we usually want this to trigger a DeprecationWarning starting immediately, until it is actually removed, so the `current_version` is probably not necessary. This is what @zhou-pj did in [glotzerlab/signac-flow#148](https://github.com/glotzerlab/signac-flow/pull/148). By ignoring `current_version`, we could ignore some of the challenges with getting the package version to be read by other files in the package. Also affects #223.

I agree, it appears that we might not need that argument unless we want to deprecate in some *future* version.

vyasr:
@csadorf Should we also deprecate the `get_job` functions in `project.py`, both the `classmethod` and the module-level function? I see why it was implemented, but it may not be necessary given how we've largely consolidated around the project interface as the primary user-facing API.

bdice:
> 
> 
> @csadorf Should we also deprecate the `get_job` functions in `project.py`, both the `classmethod` and the module-level function? I see why it was implemented, but it may not be necessary given how we've largely consolidated around the project interface as the primary user-facing API.

Wait, why? I use `signac.get_job()` all the time.

csadorf:
> > @csadorf Should we also deprecate the `get_job` functions in `project.py`, both the `classmethod` and the module-level function? I see why it was implemented, but it may not be necessary given how we've largely consolidated around the project interface as the primary user-facing API.
> 
> Wait, why? I use `signac.get_job()` all the time. Specifically in the `signac shell`, this makes it easy to `cd` into a job directory and get data from the job document.

Same, no indication for deprecation.

vyasr:
> > > @csadorf Should we also deprecate the `get_job` functions in `project.py`, both the `classmethod` and the module-level function? I see why it was implemented, but it may not be necessary given how we've largely consolidated around the project interface as the primary user-facing API.
> > 
> > 
> > Wait, why? I use `signac.get_job()` all the time. Specifically in the `signac shell`, this makes it easy to `cd` into a job directory and get data from the job document.
> 
> Same, no indication for deprecation.

If it's being used then that's fine, I asked because I don't personally make use of it.

signac/contrib/filesystems.py
```diff
@@ -27,12 +28,19 @@
 GRIDFS_LARGE_FILE_WARNING_THRSHLD = int(1e9)  # 1GB
 FILESYSTEM_REGISTRY = dict()
 
+# this is here because of issues importing the same variable in
+# signac/__init__.py from the top level namespace
+__version__ = '1.2.0'
```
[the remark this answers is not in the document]
tommy-waltmann: Python 2.7 environments crash when I remove the current_version from the @deprecated decorators, so when we stop supporting python 2, this can be removed. Until then I think we should just leave it how it is.

signac/contrib/filesystems.py
```diff
@@ -27,12 +28,19 @@
...
 
+
+@deprecated(deprecated_in="1.3", removed_in="2.0", current_version=__version__,
+            details="No references to this function outside the module.")
 def _register_fs_class(fs):
```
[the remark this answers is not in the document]
tommy-waltmann: Functions and classes like these are used by public functions and classes, so we cannot just remove them, especially since marking things @deprecated just generates warnings, not errors.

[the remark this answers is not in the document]
csadorf: In this case only the related public classes and functions should be deprecated, not the internal functions.

.gitignore
```diff
@@ -45,3 +45,6 @@ _*
 *.log
 
 .DS_Store
+
+# sublime project files
+signac.sublime-*
```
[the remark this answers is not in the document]
tommy-waltmann: Agreed, but at some point if I want to use sublime text this will have to be a part of the git ignore file at some point.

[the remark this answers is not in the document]
csadorf: Yes, that's why I'm fine with just leaving this as part of this PR, just as a reminder that we try to keep PRs on-topic.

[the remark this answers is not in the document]
bdice: @tommy-waltmann @csadorf I would recommend putting this into your personal `~/.gitexcludes` file. Example: https://github.com/bdice/dotfiles-public/blob/master/home/.gitexcludes

signac/contrib/filesystems.py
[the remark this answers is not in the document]
csadorf: @tommy-waltmann So I did some research/testing. I think the optimal solution is to place the `__version__` variable in a separate `version.py` module within the package root. This makes it possible to import the variable both into the root and within the sub-packages.

I was not able to reproduce any issues with respect to Python 2 in combination with omitting this parameter, but either way, I would appreciate if you could apply above proposed change.

signac/common/host.py
```diff
@@ -18,7 +19,13 @@
 SESSION_PASSWORD_HASH_CACHE = SimpleKeyring()
 SESSION_USERNAME_CACHE = dict()
 
+# this is here because of issues importing the same variable in
+# signac/__init__.py from the top level namespace
+__version__ = '1.2.0'
```
csadorf: * [ ] Import from `signac/version.py` (see previous comment).

signac/contrib/__init__.py
```diff
@@ -25,7 +25,6 @@
 
 logger = logging.getLogger(__name__)
 
-
```
csadorf: Is this something the linter complained about? Otherwise, please revert this change.

signac/syncutil.py
```diff
@@ -2,11 +2,11 @@
 import shutil
 import filecmp
 import logging
+
 from copy import deepcopy
 from contextlib import contextmanager
 from filecmp import dircmp
 
-
```
csadorf: Please revert PR-unrelated changes to this module.

signac/common/crypt.py
```diff
@@ -52,19 +58,29 @@ def _decode(cls, msg):
         else:
             return base64.b64decode(msg).decode()
 
+    @deprecated(deprecated_in="1.3", removed_in="2.0", current_version=__version__,
+                details="The crypt module is deprecated.")
```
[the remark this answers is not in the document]
tommy-waltmann: All the documentation says that the @deprecated decorators is used on functions, and when I try to add the decorator above a class, failures occur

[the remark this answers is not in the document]
csadorf:
```python
# test.py
import deprecation

@deprecation.deprecated()
class Foo:
    pass

if __name__ == '__main__':
    print('deprecation version', deprecation.__version__)
    foo = Foo()
```

```
$ python test.py
test.py:11: DeprecatedWarning: Foo is deprecated
  foo = Foo()
deprecation version 2.0.6

```
Am I missing something? Are you perhaps confusing the `deprecated` package with the `deprecation` package?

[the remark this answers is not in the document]
tommy-waltmann: I'll look more into what may be going on on my end.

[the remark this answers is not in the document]
csadorf: Just realized I might have added to the confusion with my first comment. Let me know if you have trouble sorting this out. I am happy to help.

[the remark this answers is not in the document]
tommy-waltmann: I think I have figured out the issue. For some reason classes which have subclasses create errors when the @deprecated decorator is applied to them, even if all their subclasses also have the @deprecated decorator applied to them as well.

Try this out for yourself by extending your example to make sure I am correct on this.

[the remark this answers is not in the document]
tommy-waltmann: deprecating the __init__ method of the class as well as all its classmethods would be functionally equivalent to deprecating the class, so I'll do that for now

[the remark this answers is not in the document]
csadorf: > I think I have figured out the issue. For some reason classes which have subclasses create errors when the @deprecated decorator is applied to them, even if all their subclasses also have the @deprecated decorator applied to them as well.
> 
> Try this out for yourself by extending your example to make sure I am correct on this.

Yes, it seems that you can't really use the decorator for classes with sub-classes. Is that a major issue? How often do we encounter this problem?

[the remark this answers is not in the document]
tommy-waltmann: Most of the classes we want to deprecate are a part of some sort of inheritance tree. Usually we want to deprecate the entire tree.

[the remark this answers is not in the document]
csadorf: I recommend that you [create an issue](https://github.com/briancurtin/deprecation/issues?q=is%3Aopen+is%3Aissue) on the `deprecation` GitHub repo. We might be missing something or maybe they have a work-around. In either case, we should document this blocker.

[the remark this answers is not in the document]
tommy-waltmann: I did that, and created a quick fix by deprecating the __init__ method in classes we want to deprecate.

I believe I have addressed all the issues that you've brought up, so I am going to request another round of reviews of what we have right now on this PR.

benchmark.py
```diff
@@ -271,7 +272,7 @@ def store_result(key, doc):
     default_doc = create_doc(args)
     default_doc['meta']['versions'] = {
         'python': '.'.join(map(str, sys.version_info)),
-        'signac': signac.__version__,
+        'signac': __version__,
```
csadorf: The changes to this module should not be necessary.

signac/common/crypt.py
```diff
@@ -28,15 +29,13 @@ def get_keyring():
...
-# signac/__init__.py from the top level namespace
-__version__ = '1.2.0'
-
 
 # this class is deprecated
 class SimpleKeyring(object):
     """Simple in-memory keyring for caching."""
 
+    @deprecated(deprecated_in="1.3", removed_in="2.0", current_version=__version__,
+                details="The crypt module is deprecated.")
```
csadorf: Is this class sub-classed? Otherwise, I'd prefer to apply the decorator the class whenever possible.

signac/common/host.py
```diff
@@ -38,6 +35,7 @@ def get_default_host(config=None):
             raise ConfigError("No hosts specified.")
 
 
+# deprecated
```
csadorf: I think I'd prefer to add one comment to the top of this module that the whole module is deprecated instead of commenting on each private function here. The per-function comments are better applied where we don't deprecate the whole module.

signac/contrib/__init__.py
```diff
@@ -25,6 +25,7 @@
 
 logger = logging.getLogger(__name__)
 
+
```
csadorf: Linter? Otherwise, I'd prefer to not have unrelated style fixes in the PR.

signac/syncutil.py
```diff
@@ -3,6 +3,7 @@
 import filecmp
 import logging
 
+
```
csadorf: seems unrelated

benchmark.py
tommy-waltmann: The __version__ variable is now located at signac.version.__version__ not signac.__version__, so I think this change is necessary.

signac/common/crypt.py
tommy-waltmann: I am just keeping the deprecation method consistent between classes which are subclassed and classes which aren't. Although I don't really have a preference for one or the other.

requirements.txt
```diff
@@ -0,0 +1 @@
+deprecation>=2
```
bdice: @csadorf In the words of @rosecers, this comment is a "signal-boost" that signac is no longer dependency-free. I'm not sure if that affects anything else beyond this repo (potentially CI?) and the conda-forge recipe.

csadorf: The added implicit dependencies are `packaging` and `pyparsing`, which are both *noarch*. So I'm ok with that. We have been using the language of "minimal dependencies" instead of "dependency free" for a while now.

signac/common/crypt.py
bdice: Please deprecate the class directly if you're not prohibited from doing so by inheritance.

csadorf:
For future reference: https://github.com/briancurtin/deprecation/issues/43

csadorf:
@vyasr @bdice Since you both requested changes at some point, I'd like you to approve this PR before we merge it.

bdice:
@tommy-waltmann We'll need to do something to fix the failing test. https://github.com/glotzerlab/signac/blob/a9c0ff6de9e262c72acc1d41ac866573dbd24b6b/tests/test_project.py#L1442-L1443

Here, there is a `DeprecationWarning` being raised when the `index` is provided, which is masking the `ValueError` expected by this test. This appears to only affect Python 2 (I'm guessing it has different behaviors about warnings vs. errors). We have a few options. We could:
1. skip this test on Python 2,
2. remove this test (since it's only testing an invalid argument that is also deprecated),
3. attempt to hide the `DeprecationWarning` using something like
```python
with self.assertRaises(ValueError):
    with warnings.catch_warnings():
        warnings.simplefilter('ignore')
        # run test
```

signac/contrib/project.py
bdice: This still needs to be done. @tommy-waltmann The change you want looks like this, roughly (comments are only for your guidance, don't include them):
```python
# Move the implementation to this private method
def _find_jobs_ids(...):
    # do stuff - copy the implementation from the current public method
    pass

# Keep a public method that is deprecated
@deprecated(...)
def find_job_ids(...):
    return _find_job_ids(...)
```

signac/contrib/project.py
```diff
@@ -1073,6 +1090,7 @@ def import_from(self, origin=None, schema=None, sync=None, copytree=None):
             origin=origin, project=self, schema=schema, copytree=copytree))
         return paths
 
+    # TODO: jobs_ids isn't used right now
```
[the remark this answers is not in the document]
bdice: I created #224. @tommy-waltmann Please remove the TODO.

signac/contrib/project.py
```diff
@@ -1177,6 +1196,9 @@ def repair(self, fn_statepoints=None, index=None, job_ids=None):
             raise JobsCorruptedError(corrupted)
 
     def _sp_index(self):
+        """
+        Update and return the statepoint index cache
```
```suggestion
        Update and return the statepoint index cache.
```

bdice:
@tommy-waltmann Please edit this file to add yourself, as well: https://github.com/glotzerlab/signac/blob/master/contributors.yaml

requirements.txt
vyasr: Is this discussion just about whether the file needs to exist?

signac/contrib/indexing.py
```diff
@@ -44,13 +52,16 @@ def _is_blank_module(module):
         return not bool(file.read().strip())
 
 
+# this class is deprecated
```
vyasr: I don't think we need this since the entire module is marked as deprecated.

signac/contrib/indexing.py
```diff
@@ -124,6 +135,7 @@ def process(self, doc, dirpath, fn):
         return doc
 
 
+# this class is deprecated
```
vyasr: Same as above.

signac/contrib/indexing.py
```diff
@@ -345,6 +371,7 @@ def _index_signac_project_workspace(root,
         logger.debug("Indexed workspace '{}', {} entries.".format(root, i+1))
 
 
+# this class is deprecated
```
vyasr: And here.

signac/contrib/indexing.py
```diff
@@ -410,6 +439,7 @@ def crawl(self, depth=0):
             yield doc
 
 
+# this class is deprecated
```
vyasr: Remove.

signac/contrib/indexing.py
```diff
@@ -44,13 +52,16 @@ def _is_blank_module(module):
         return not bool(file.read().strip())
 
 
+# this class is deprecated
 class BaseCrawler(object):
```
vyasr: Is there a reason we don't just deprecate all of the classes directly, rather than going through and deprecating methods?

signac/contrib/utility.py
```diff
@@ -111,8 +117,11 @@ def set_verbosity_level(verbosity, default=None, increment=10):
         level=default - increment * verbosity)
 
 
+# this class is deprecated
```
vyasr: Remove, add deprecation warning to class.

signac/contrib/indexing.py
csadorf: It's a work-around for this bug: https://github.com/briancurtin/deprecation/issues/43 

So, unfortunately necessary for classes that we inherit from ...

vyasr: Wonderful. @tommy-waltmann that helps you then, you can ignore about half my comments :)

tommy-waltmann:
So, I have noticed that there is an error with 1 specific python 2.7 unit test. The problem is as follows:

The test is testing a bad index argument given to the project.create_linked_view() method within a "with assertRaises(ValueError)" block. We decided to put a DeprecationWarning for when the index argument is used and python 2.7 is saying this unit test fails because the code is raising the DeprecationWarning before the ValueError comes later.

I looked at the implementation of create_linked_view(), and it seems that the index argument still does do something.

Should we modify the unit test, remove the index argument from the api, or something else?

signac/contrib/project.py
```diff
@@ -87,6 +89,16 @@ def _resolve_statepoint_filter(self, q):
             else:
                 yield 'statepoint.{}'.format(k), v
 
+    def _find_job_ids(self, filter, doc_filter=None):
```
```suggestion
    def _find_job_ids(self, filter=None, doc_filter=None):
```

signac/contrib/project.py
```diff
@@ -424,6 +430,7 @@ def __contains__(self, job):
         """
         return job.get_id() in self.find_job_ids()
```
bdice: Let's replace the internal uses of `find_jobs_ids()` with `_find_job_ids()`.
```suggestion
         return job.get_id() in self._find_job_ids()
```

signac/contrib/project.py
```diff
@@ -506,6 +515,8 @@ def detect_schema(self, exclude_const=False, subset=None, index=None):
         statepoint_index = self.build_job_statepoint_index(exclude_const=exclude_const, index=index)
         return ProjectSchema.detect(statepoint_index)
 
+    @deprecated(deprecated_in="1.3", removed_in="2.0", current_version=__version__,
+                details="Use find_jobs().ids instead.")
     def find_job_ids(self, filter=None, doc_filter=None, index=None):
```
bdice: I didn't realize this until now -- we have a `find_job_ids` function in both `JobSearchIndex` and `Project`. We're deprecating both, I think. Please apply the same changes as above (make a private method with the implementation, and a deprecated public method with the same arguments that calls the private method).

signac/contrib/project.py
vyasr: Other places where it's used that I see: `open_job`, `min_len_unique_id`, `check`, `repair`, `_build_index`, and in `JobsCursor.__len__` and `JobsCursor.__iter__`. Should just require adding the underscore once you add the new method.

signac/contrib/project.py
vyasr: Good catch, I knew the method was in both places but misread where the private method was added. Definitely need to have a private method in the `Project` namespace for the internal calls. That's even more important than the deprecation in `JobSearchIndex` I think.

signac/contrib/project.py
```diff
@@ -87,6 +89,16 @@ def _resolve_statepoint_filter(self, q):
...
+    def _find_job_ids(self, filter=None, doc_filter=None):
+        if filter:
+            filter = dict(self._resolve_statepoint_filter(filter))
+            if doc_filter:
+                filter.update(doc_filter)
+        elif doc_filter:
+            filter = doc_filter
+        return self._collection._find(filter)
+
+    @deprecated(deprecated_in="1.3", removed_in="2.0", current_version=__version__)
```
csadorf: The `JobSearchIndex` class is not explicitly exposed as part of the public API, but only implicitly as a return value of `Project.build_job_search_index()`. Since that function is now deprecated, I don't think we need to deprecate the function as part of this class.

signac/contrib/project.py
```diff
@@ -730,6 +745,8 @@ def _get_statepoint_from_workspace(self, jobid):
                 raise JobsCorruptedError([jobid])
             raise KeyError(jobid)
 
+    @deprecated(deprecated_in="1.3", removed_in="2.0", current_version=__version__,
+                details="Use open_job(id=jobid).statepoint() function instead.")
```
csadorf: This function is used internally, so unless this is refactored, we should probably move its logic into a private method. But we could (should) do that as part of a separate refactoring PR.

signac/core/utility.py
```diff
@@ -3,8 +3,12 @@
 # This software is licensed under the BSD 3-Clause License.
 import re
 import subprocess
+from deprecation import deprecated
+from ..version import __version__
 
 
+@deprecated(deprecated_in="1.3", removed_in="2.0", current_version=__version__,
+            details="This function is obsolete, use equivalent function in connection module.")
```
csadorf: Since the `connection` module is also deprecated, I advise that we just say it's obsolete.

signac/version.py
```diff
@@ -0,0 +1,6 @@
+# Copyright (c) 2019 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+
+
+__version__ = "1.2.0"
```
csadorf: We should explicitly expose `__version__` here.
```suggestion
__version__ = "1.2.0"

__all__ = ['__version__']
```

signac/contrib/project.py
csadorf: > I didn't realize this until now -- we have a `find_job_ids` function in both `JobSearchIndex` and `Project`. We're deprecating both, I think. Please apply the same changes in `Project` that you did to `JobSearchIndex` (make a private method with the implementation, and a deprecated public method with the same arguments that calls the private method).

I missed that earlier. No need to deprecate methods, just the one in the `Project` class. The `JobSearchIndex` is only implicitly exposed to users as a return value of a now deprecated method. See also my latest review.

signac/contrib/project.py
bdice: Created #227.

tommy-waltmann:
I'm a little confused, are we planning on merging this PR as it currently exists, or are there more things left to do?

bdice:
> 
> 
> I'm a little confused, are we planning on merging this PR as it currently exists, or are there more things left to do?

I'm handling the remaining edits and merging it.

PR #222, opened 2019-08-23
Move errors to signac.errors

b-butler:
* Deprecation warning to error modules
* core, contrib, common error modules have deprecation warning

## Description

Move all the signac specific exceptions to `signac.errors`.
## Motivation and Context

Removes code complexity and put similar functionality together.
## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/common/config.py
+++ b/signac/common/config.py
@@ -7,7 +7,7 @@ import logging
 
 from .configobj import ConfigObj, ConfigObjError
 from .validate import get_validator, cfg
-from .errors import ConfigError
+from ..errors import ConfigError, PermissionsError
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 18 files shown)


csadorf (commented):
I do not agree that removing **all** exception classes from their respective contexts is simplifying the package code. I agree that there should probably be one central error or exception module from which most exception classes should derive from, but error classes like `WorkspaceError` do not need to be separated from the `project` module for example. I do not believe that this is standard practice either. For example, the `numpy` package defines exception classes close to the respective error context.

I would need to see more arguments in support of this change in order to be comfortable to approve it.

csadorf:
> @csadorf I don't have too strong an opinion either way other than trying to simplify the error import logic. I see your point and would say that any error that was defined in the module where it was raised could/should stay in that module. Although I would argue that the ones defined in the subpackage error module could just be in a central error module under the direct package.

Considering that we still need to simplify the overall package structure, I might recommend that we hold off consolidating the error classes for now.

vyasr:
I think this PR can wait until the overall package structure is simplified. The decision will be easier once we reorganize and know where all modules actually live.

csadorf:
Closing this PR until we tackle the package reorganization.

PR #231, opened 2019-10-05
Changed Project __contains__ method to run in constant time.

bdice:
## Description
I was running a cheap operation on a large data space (50k statepoints) with `python project.py run`. I noticed there was a _very_ large initialization time before the operations began. I hit Ctrl+C a few times and identified that the `__contains__` check was taking time proportional to `O(N_jobs)`.

The specific code path being triggered was:

1. `FlowProject.run` calls the `select` filter over pending operations:
https://github.com/glotzerlab/signac-flow/blob/c4d84fecaa18cd7556be85090e4332482bd79d38/flow/project.py#L1692

2. The `select` filter checks if `job not in self`:
https://github.com/glotzerlab/signac-flow/blob/c4d84fecaa18cd7556be85090e4332482bd79d38/flow/project.py#L1650

3. The `__contains__` method calls `self._find_job_ids()` with no arguments:
https://github.com/glotzerlab/signac/blob/48e3158022a8ba84a4448abb6abce134d9a7eb8f/signac/contrib/project.py#L411

4. The `_find_job_ids()` function returns a list of `self._job_dirs()` immediately:
https://github.com/glotzerlab/signac/blob/48e3158022a8ba84a4448abb6abce134d9a7eb8f/signac/contrib/project.py#L524-L526

5. `_job_dirs()` iterates over all the subdirectories in `self._wd`, checking each of them against the regular expression `JOB_ID_REGEX` before yielding.
https://github.com/glotzerlab/signac/blob/48e3158022a8ba84a4448abb6abce134d9a7eb8f/signac/contrib/project.py#L374-L376

6. Finally, `job.get_id()` is checked against the list of all job ids returned by `_job_dirs()`.

In my understanding, this should be equivalent to a constant-time check like the one I proposed in this PR. The regular expression validation is not necessary because we're getting the job id directly from the `Job` object itself, and we really just need to know if that id string is a subfolder of `self.wd`.

## Motivation and Context
Before implementing this change, the `select` filter was running at ~300 iterations per second. After, it ran the `select` filter almost immediately (couldn't time it effectively, but there were 45,000 jobs so it's about 100x faster for this size of data space).

Existing tests pass, and I welcome suggestions for additional tests if this might introduce any unexpected behavior.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature
- [x] Performance enhancement 🐎 🚗 

## Checklist:

If necessary:


bdice:
@vyasr I added you as a reviewer since the Pull Assigner wasn't working. I would like your input on whether this is a "safe change" with respect to signac's data model, before I go further with this PR.

bdice:
Offline discussion with @vyasr: This is probably an ok change but it makes our implementation more concrete and less abstract. I'm perfectly fine with that (in light of our potential rewrites for a more flexible storage engine in signac v2.0) because it's reasonable to expect that other potential backends should be able to implement their own `__contains__` check in O(1), without relying on generic O(N) methods.

@csadorf If you'd like to weigh in, feel free. Otherwise we'll plan on merging this ~3 days after @vyasr reviews.

vyasr:
This must be merged after #232.

Issue #213, opened 2019-07-25
`signac view` fails when statepoints contain forwardslashes


klywang:
### Description

`signac view` cannot create folders when statepoints contain forward slashes (ie, if a statepoint value is '/oasis/scratch/some_file.txt')

### To reproduce

`signac view`

Example of statepoint: 
 `{'pot_file': '/oasis/scratch/path/to/file/some_file.out'}`

### Error output

Error: [Errno 13] Permission denied: '/oasis'### System configuration

Please complete the following information:

 - Operating System [e.g. macOS]: macOS Moajve (10.14.4)
 - Version of Python [e.g. 3.7]: 3.7.3
 - Version of signac [e.g. 1.0]: 1.2.0

vyasr:
I think the proper behavior here is to just throw a more descriptive error message. Right now it just gives a permission denied error in this case, we should at least change it to raise an Exception that informs the user why it doesn't work. I don't think that this is necessarily something that we should support, because it would require modifying the values to avoid containing file separators.

csadorf:
I agree that adding a more descriptive error message and potentially suggesting a work-around is the way to go. We have been pretty liberal about what a state point key and value may contain as long as it is JSON-encodable and issues like that are kind of bound to happen.

My suggested solution is to explicitly check for `os.sep` in keys and values and simply error out. A potential work-around is to only include keys and values that do not contain these values.

@klywang Would that be a satisfactory solution?

klywang:
@csadorf Yeah that sounds good!

mikemhenry:
Just to clarify the behavior we want:
paths are okay in state points but will prevent view from working (with a helpful error message)

vyasr:
Yes, I think that's the best option. We could potentially disallow them in 2.0, but I don't think it's necessary since it only causes problems for views.

vyasr:
Resolved by #233.

PR #233, opened 2019-10-10
check if the statepoint key or values contain a path separator (a sla…

jennyfothergill:
…sh?), and raise an error if they do, also prints the bad items

## Motivation and Context

prevents unexpected behavior in `signac view`

issue #213

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:


vyasr:
@atravitz would appreciate a second review if you don't mind! This PR is meant to resolve #213.

atravitz:
are there any other symbols (other than / or \) that could cause trouble while we're fixing this? Like * or spaces in strings?

vyasr:
That's a good call. @jennyfothergill could you add a test for this branch of the code? You'll basically want to add a test that creates a signac project with some odd statepoint keys, try calling signac view, and make sure that either views generate correctly or that signac throws an error.

tests/test_project.py
```diff
@@ -1731,6 +1731,21 @@ def test_create_linked_view_heterogeneous_schema_problematic(self):
         with self.assertRaises(RuntimeError):
             self.project.create_linked_view(view_prefix)
 
+    def test_create_linked_view_with_slash_raises_error(self):
+        sp_badkey = [{'a/b': 0, 'b': 10} for i in range(5)]
+        sp_badkeyval = [{
+            'a/b': 'dir/file{}'.format(i), 'b': i % 3
```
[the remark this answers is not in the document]
jennyfothergill: Yeah, that's a better idea. I've updated it so it will iterate through a list of bad characters and then the test also loops through that list, so adding new ones should be easy. I didn't account for nested keys or values, but I don't know if I should/I'm not sure the best way to do it.

jennyfothergill:
@mikemhenry had a cool idea to whitelist the character used for POSIX compliant filenames instead of worrying about which characters we should check for, so I've switched it to that.

jennyfothergill:
This is probably outside the scope of this PR, but how should statepoint key/values that are lists, arrays, etc be handled?

vyasr:
Those types of values are fine, they just turn into folders with parenthetical lists e.g. `a/(1, 2, 3)/b/(4, 5, 6)`. Which are valid but ugly filenames, and perhaps not perfectly portable.

I'm fine with the whitelisting solution. I'm curious whether @csadorf thinks it's too restrictive, particularly w.r.t. backwards compatibility.

csadorf:
> I'm fine with the whitelisting solution. I'm curious whether @csadorf thinks it's too restrictive, particularly w.r.t. backwards compatibility.

Yes, I think that is too restrictive. Many [modern file systems](https://en.wikipedia.org/wiki/Ext4) are able to handle arbitrary Unicode characters just fine. I don't think we should unnecessarily restrict users, especially those who are not from the anglophone areas of the world, to effectively use the view feature by limiting valid characters to the small set of ASCII chars.

With respect to the original issue, let's focus on checking for the file path separation characters (`os.sep)` and otherwise be not too restrictive. If we are really concerned about checking valid file paths, then we could resort to a library like [pathvalidate](https://pypi.org/project/pathvalidate/), but I don't think this is necessary within the scope of this PR.

mikemhenry:
It is only restrictive in the sense that `view` is picky about what values it allows for keys & values. I'm fine with going back to `os.sep` as the only string we prohibit, but I feel like we will have to periodically add strings to our prohibited list as users run into them. Maybe this is a case of YAGNI and we will just worry about it when we get a new issue.

tests/test_project.py
```diff
@@ -1731,6 +1731,17 @@ def test_create_linked_view_heterogeneous_schema_problematic(self):
...
 
+    def test_create_linked_view_with_slash_raises_error(self):
+        bad_chars = [os.sep, " ", "*"]
+        statepoints = [{
+            'a{}b'.format(i): 0, 'b': 'bad{}val'.format(i)
+            } for i in bad_chars]
+        view_prefix = os.path.join(self._tmp_pr, 'view')
+        for sp in statepoints:
+            self.project.open_job(sp).init()
+            with self.assertRaises(NotImplementedError):
```
csadorf: see comment above

signac/contrib/linked_view.py
```diff
@@ -39,6 +39,28 @@ def create_linked_view(project, prefix=None, job_ids=None, index=None, path=None
...
+            ]
+
+    if any(bad_items):
+        err_msg = " ".join([
+            "In order to use view, statepoints should not contain {}:".format(
+                bad_chars
+            ),
+            *bad_items
+        ])
+        raise NotImplementedError(err_msg)
```
csadorf: I would prefer this to be a `RuntimeError`.

csadorf:
@jennyfothergill Thank you very much for your work. I would like to encourage you to add your name to the list of contributors. It helps us track who contributed to the project and gives you some additional exposure in the open-source sphere. ;)

signac/contrib/linked_view.py
jennyfothergill: OK, I'll make this change!

jennyfothergill:
Thanks very much!

Issue #238, opened 2019-10-31
Proposal: Lazy statepoint loading


bdice:
### Feature description

I am experimenting with refactoring statepoint loading to be lazy. Currently, `project.open_job()` is called on every job during a `flow run` command, which causes the statepoints to be loaded. However, the statepoint information isn't even used in many cases. This may be a case where signac is executing unnecessary I/O (which is costly for large projects like the ones I'm working with).

### Proposed solution

The property `job.statepoint` currently just returns `job._statepoint`, meaning that the internal state is effectively identical to the publicly-accessible state. I am investigating making the property `job.statepoint` load lazily, with `job._statepoint = None` until the statepoint is requested.

I see a HUGE performance boost (from 30 seconds to <0.5 seconds for a simple operation on 3000 jobs) since the I/O is dropped dramatically, but some tests fail because job statepoint corruption (which raises a `JobsCorruptedError`) is not detected at the time of `project.open_job()`. I think that lazy loading is _generally_ safe except in cases where the statepoint hash and job id don't match (corrupted jobs).

I just wanted to open this for discussion - I am not sure where we stand on approaches to handling corrupted data. Specifically, I want to know whether we think that validating `hash(statepoint) == id` (and thus the I/O cost of a statepoint load) is always necessary **when opening a job by id**.

### Additional context

Maybe investigate `async` as an alternative...? Still expensive but maybe less so.

bdice:
We discussed this offline (I can't remember who was a part of the discussion). Our conclusion was that loading a job by id should _not_ require validating that `hash(statepoint) == id`. A user (or I/O error) may corrupt the data space by breaking that invariant, and the user will still be allowed to open the job by id. On accessing `job.statepoint`, job corruption will be checked. @glotzerlab/signac-committers Was this a conclusion you recall and/or agree with?

csadorf:
I am ok with lazy loading and only validating the state point metadata when it is accessed.

b-butler:
@bdice I don't remember being a part of the conversation, but I agree with the approach.

vyasr:
@bdice and I did discuss and agree on this approach. I need to post my pseudo-prototype for #249 along with a class diagram so that we can progress on that front. IMO that serves as the cleanest path forward for implementing lazy loading by isolating the logic for synchronization and ensuring that a clear set of invariants are well-defined and well-tested for the different possible use-cases of nesting and buffering.

bdice:
[...]

Issue #81, opened 2018-01-26
Better API for the modification of configuration files


csadorf:
**[Original report](https://bitbucket.org/glotzer/signac/issue/81) by Carl Simon Adorf (Bitbucket: [csadorf](https://bitbucket.org/csadorf), GitHub: [csadorf](https://github.com/csadorf)).**

----------------------------------------

Modifying global, local, and project config files currently requires the use of low-level API as part of the `common.config` namespace. Manipulating the configuration should be simplified with a high-level API.

As an example, it should be possible to change the project's workspace configuration like this:
```
#!python
project.config['workspace_dir'] = '/path/to/new/workspace'
```

csadorf:
We could consider to make `Project.workspace` a property.

csadorf:
I suggest we move this to 1.1, any objections?

bdice:
@vyasr I think this is superseded by #246? If so, please close.

vyasr:
Yes, I don't want to close until that PR is merged but will do so until it is.

PR #239, opened 2019-10-31
Lazy statepoint loading

bdice:
## Description
Changes behavior of `Job` to load its statepoint lazily, when opened by id.

## Motivation and Context
Implementation of #238.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


signac/contrib/project.py
```diff
@@ -363,7 +363,12 @@ def open_job(self, statepoint=None, id=None):
                     id = matches[0]
                 elif len(matches) > 1:
                     raise LookupError(id)
-            return self.Job(project=self, statepoint=self.get_statepoint(id), _id=id)
```
bdice: Note that the method `get_statepoint()` is deprecated anyway, so removing it here is potentially helpful?

bdice:
There is a lot of refactoring that I think would be possible if this feature were accepted. Specifically, we have deprecated `Project.get_statepoint(self, jobid, fn=None)` for v1.3 (to be removed in v2.0) but we are still actively using it in `Project.open_job()`.

If statepoints were loaded lazily, then it would be easy to remove most (or all) statepoint-reading logic from the `Project` class and leave it all to the `Job`, which could be opened by `id` alone. Project-level caching of statepoints is valuable, and this would only apply to cases that miss the cache. For those cases, we currently have methods like `Project._get_statepoint_from_workspace(self, jobid)`. I think this would allow for some nice simplification of the core of the Project / Job codebase while potentially improving performance in other places where lazy statepoint loads would benefit the internals.

bdice:
@glotzerlab/signac-maintainers I would like some conceptual feedback on this PR's stated benefits/costs. I think that something like this, or a more automatically/aggressively updated statepoint cache, should be considered to improve performance.

csadorf:
> @glotzerlab/signac-maintainers I would like some conceptual feedback on this PR's stated benefits/costs, though the code is not necessarily "ready for review" since I don't intend to merge it until the concepts are improved and other possible solutions are discussed. I think that something like this lazy-load-by-id, or an automatically/aggressively updated statepoint cache, should be considered to improve performance.

I think before we delve more deeply into any conceptual discussion or possibly even approaches for implementation, I think it would be crucial to identify the exact use cases and bottlenecks that we want to address. Looking at the [benchmarks](https://app.circleci.com/jobs/github/glotzerlab/signac/3861/parallel-runs/0/steps/0-107), it appears that this PR is already improving performance significantly, albeit not by an order magnitude. Is the use case that this PR is supposed to speed up already covered by our benchmarks?

bdice:
@csadorf The benchmarks `iterate` and `iterate_single_pass` are using the new code path, since `list(project)` opens jobs by id. This may be sufficiently covered by the benchmarks as-is, unless you wanted something else.

csadorf:
According to the benchmarks, it appears that loading each state point is somewhat slower (1.5-2x). @bdice Maybe you can do a little bit of profiling? I'm willing to move forward regardless, but I think we should take some care not slow down signac in a very relevant use case.

bdice:
I have done some profiling. The primary difference in cost is an error check that is needed in the lazy code but not the original code path. 

Suppose we try to open a job id like `'ffffffffffffffffffffffffffffffff'` that does not exist in the workspace. The **current master branch** fails when initializing the job by statepoint, in the call to `self.get_statepoint(id)`.
https://github.com/glotzerlab/signac/blob/cfc41de9b971af14a31ca0bcfce4aedca22bba5e/signac/contrib/project.py#L366
The path to the statepoint does not exist, raising a `KeyError`.
https://github.com/glotzerlab/signac/blob/cfc41de9b971af14a31ca0bcfce4aedca22bba5e/signac/contrib/project.py#L721

The **lazy branch** does not load the statepoint in `Job.__init__`, so we _must_ check whether the job id exists in the workspace directory to know if we're fetching an already-existing job (the job must already exist when opening by id):
https://github.com/glotzerlab/signac/blob/51b6c261496a764ac864fb5746ce371574555fff/signac/contrib/project.py#L369-L371

Another small improvement that could be made someday in the future (not in this PR):
`project._wd` and `project._rd` should be cached and only updated if `project.config` is modified. Getting the project workspace directory accounts for ~13% of the runtime in both the old and new code. In contrast to the suggestions in #81, it would be ideal for the `project.config` to be immutable to make this easier.

csadorf:
I'd like to merge #243 prior to this PR.

csadorf:
@bdice Please rebase on/merge master so that we can run the latest benchmark script with this branch.

bdice:
Another thing I just noticed -- the **master branch** doesn't ensure that statepoints and ids correspond (that is, there is no corruption check) when opening jobs. I thought this was the expected behavior, so I made sure that this check was performed when loading lazily. This has performance implications (around 5% of runtime for a simple statepoint fetching loop, `[job.sp() for job in project]`), which also partly explains the slower speed of the **lazy branch**.

On **master**, this never triggers a `JobsCorruptedError`:
```python
import signac
import json
pr = signac.init_project('test')

# Clear all jobs
for job in pr:
    job.remove()

job = pr.open_job({'a': 0}).init()

# Corrupt the job statepoint/id mapping
with open(job.fn('signac_statepoint.json'), 'w') as sp:
    json.dump({'a': 1}, sp)

print('Existing job handle:', job._id, job.sp())
id_, sp_ = job._id, job.sp()
del job

# Re-initialize the project so that we skip the project._sp_cache
del pr
pr = signac.init_project('test')

job2 = pr.open_job(id=id_)
print('New job handle by id:', job2._id, job2.sp())

job3 = pr.open_job(sp_)
print('New job handle by sp:', job3._id, job3.sp())
```

**master** output:
```python
Existing job handle after corruption: 9bfd29df07674bc4aa960cf661b5acd2 {'a': 0}
New job handle by id: 9bfd29df07674bc4aa960cf661b5acd2 {'a': 1}
New job handle by sp: 9bfd29df07674bc4aa960cf661b5acd2 {'a': 0}
```

**lazy** output:
```python
Existing job handle: 9bfd29df07674bc4aa960cf661b5acd2 {'a': 0}
  File "/home/bdice/code/signac/signac/contrib/job.py", line 380, in _check_manifest
    assert calc_id(manifest) == self._id
AssertionError

During handling of the above exception, another exception occurred:

  File "bug.py", line 24, in <module>
    print('New job handle by id:', job2._id, job2.sp())
  File "/home/bdice/code/signac/signac/contrib/job.py", line 243, in sp
    return self.statepoint
  File "/home/bdice/code/signac/signac/contrib/job.py", line 231, in statepoint
    statepoint = self._check_manifest()
  File "/home/bdice/code/signac/signac/contrib/job.py", line 382, in _check_manifest
    raise JobsCorruptedError([self._id])
signac.contrib.errors.JobsCorruptedError: ['9bfd29df07674bc4aa960cf661b5acd2']
```

csadorf:
I see, that would explain the slowdown. Do you think that a ~3x slowdown is justified for that increased security?

bdice:
@csadorf I am implementing a different mechanism that I hope makes this branch always faster. Also, are you reading the benchmarks backwards? The slowdown is 1.4x when reading all statepoints, not 3x. The 3x is how much faster this branch is when it uses the lazy code paths.

>>> 1.5x faster than 'origin/master' on average with a speedup of 3.3x in the best category and a slowdown of 1.4x in the worst category.
https://circleci.com/gh/glotzerlab/signac/3897

csadorf:
> @csadorf I am implementing a different mechanism that I hope makes this branch always faster. 

It might be justified either way, but that would of course make that decision easier. That said, one thing to consider is, that I don't think we had any problems with the current behavior until now.

> Also, are you reading the benchmarks backwards? The slowdown is 1.4x when reading all statepoints, not 3x. The 3x is how much faster this branch is when it uses the lazy code paths.
> 
> > 1.5x faster than 'origin/master' on average with a speedup of 3.3x in the best category and a slowdown of 1.4x in the worst category.
> > https://circleci.com/gh/glotzerlab/signac/3897

Sorry, I indeed looked at the wrong number. It's a slowdown of 1.4-2.1x.

Slightly unrelated, I'm wondering whether the median would be a better value to report instead of the mean.

bdice:
@csadorf I pushed a change that alters the behavior of `project._sp_cache`. This makes the lazy code path always faster, in commit 9824ec7. I'm still working on identifying/testing some possible edge cases.

I wanted to avoid the `os.path.exists` check (which requires disk access). Now, I initialize the statepoint cache with `{id: None for id in project._job_dirs()}`. Effectively, this is caching a list of the job ids as the keys, until the job statepoints (values) are actually read from disk. If the cached value is `None`, we can assume with high certainty that the job exists in calls to `project.open_job(id=id)`, without having to read its statepoint.

_This paragraph might not be 100% accurate, I'll revise later._
Once a job statepoint has been accessed from disk (or if the statepoint is known because the job was opened by statepoint), the job registers itself and the statepoint cache is updated so that `project._sp_cache[id] = statepoint`.

_This paragraph might not be 100% accurate, I'll revise later._
If a job is already in the statepoint cache and is removed from the project with `job.remove()`, it remains in the statepoint cache, and it is still possible to open the job by statepoint or id. **This bug also occurs in the master branch, so I don't think this branch makes anything worse.**

Minimal reproduction script:
```python
import signac
pr = signac.init_project('test')

# Clear all jobs
for job in pr:
    job.remove()

job = pr.open_job({'a': 0}).init()

print('Existing job handle:', job._id, job.sp())
id_, sp_ = job._id, job.sp()
print('Cache before job is removed:', pr._sp_cache)
job.remove()
print('Cache after job is removed:', pr._sp_cache)

job2 = pr.open_job(id=id_)
print('Job opened by id:', job2._id, job2.sp())

job3 = pr.open_job(sp_)
print('Job opened by sp:', job3._id, job3.sp())
```

csadorf:
Being able to open a job by ID even if it's not in the project is not a bug.

bdice:
> Being able to open a job by ID even if it's not in the project is not a bug.

But there's no statepoint associated with it. You're saying `project.open_job(id='f'*32)` should work in `master`? It raises a KeyError.

csadorf:
> > Being able to open a job by ID even if it's not in the project is not a bug.
> 
> But there's no statepoint associated with it. You're saying `project.open_job(id='f'*32)` should work in `master`? It raises a KeyError. Does your expectation of this behavior depend on the statepoint cache on disk / in memory?

No, it's fine to be able to open by id, even if there is no job in the project if the state point happens to be in some lookup table like the cache. It's not ok to open some arbitrary job by id where we wouldn't know the state point.

This hasn't been an issue until now, because we *always* knew the state point after constructing the instance of `Job`, that's no longer the case with your changes. So, maybe we need to be more explicit about this now.

bdice:
This should be handled after #249. This PR is somewhat close but needs careful testing. I would prefer to refactor the internal classes before performing this optimization.

signac/contrib/project.py
```diff
@@ -1248,15 +1278,21 @@ def _get_statepoint(self, job_id, fn=None):
...
+                self._sp_cache[job_id] = statepoint
         except KeyError as error:
+            # Fall back to a file containing all state points because the state
+            # point could not be read from the job workspace.
             try:
-                statepoint = self.read_statepoints(fn=fn)[job_id]
+                statepoints = self.read_statepoints(fn=fn)
+                # Update the project's state point cache
+                self._sp_cache.update(statepoints)
+                statepoint = statepoints[job_id]
```
bdice, line 1290: A more aggressive solution would be to update the internal state point cache inside the methods `_get_statepoint_from_workspace` and `read_statepoints`. I tested this and all of our existing tests pass -- I think this would be valid. In every case where we currently call those methods, we manually update the cache with the results. Moving that "read to cache" logic into the implementation of those methods makes sense to me, but I would like a second opinion. That can also be done in a follow-up PR.

bdice:
@csadorf @klywang @cbkerr This PR is ready for review. I will update the PR description with an overview of the design and some representative benchmarks. I'm tagging Kelly and Corwin as reviewers because they expressed specific interest in performance-related improvements.

klywang:
> If a job is already in the statepoint cache and is removed from the project with `job.remove()`, it remains in the statepoint cache, and it is still possible to open the job by statepoint or id. **This bug also occurs in the master branch, so I don't think this branch makes anything worse.**
> 

I imagine this would be something to implement in a different PR, but why doesn't `job.remove()` clear the cache after removing the job?

signac/contrib/job.py
```diff
@@ -263,6 +276,34 @@ def update_statepoint(self, update, overwrite=False):
...
+            If an error occurs while reading/parsing the state point manifest.
+
+        """
+        fn_manifest = os.path.join(self._wd, self.FN_MANIFEST)
+        try:
+            with open(fn_manifest, "rb") as file:
+                manifest = json.loads(file.read().decode())
+        except OSError as error:
+            if error.errno != errno.ENOENT:
+                raise error
```
klywang, line 301: Does this need to be tested?

signac/contrib/project.py
```diff
@@ -1232,31 +1258,41 @@ def _get_statepoint(self, job_id, fn=None):
...
+            # State point cache hit
+            return self._sp_cache[job_id]
+        except KeyError:
+            # State point cache missed
+            self._sp_cache_misses += 1
+            if (
+                not self._sp_cache_warned
+                and self._sp_cache_misses > self._sp_cache_miss_warning_threshold
+            ):
+                logger.debug(
```
klywang, line 1274: And does this error need to be tested as well?

bdice:
> > If a job is already in the statepoint cache and is removed from the project with job.remove(), it remains in the statepoint cache, and it is still possible to open the job by statepoint or id. This bug also occurs in the master branch, so I don't think this branch makes anything worse.

> I imagine this would be something to implement in a different PR, but why doesn't job.remove() clear the cache after removing the job?

@klywang The first comment was a misunderstanding on my part about the expectations of the data model. The best explanation is here from Simon:

> No, it's fine to be able to open by id, even if there is no job in the project if the state point happens to be in some lookup table like the cache. It's not ok to open some arbitrary job by id where we wouldn't know the state point.

This explains the `job.remove()` behavior, which should not remove items from the cache. Once a job is registered, the state point cache will track the `id: statepoint` in-memory until the end of the session. This means that a job can be initialized, removed, and re-opened _by id_. The project's state point cache is simply a lookup table of `id: statepoint` and does not mean that all job ids in the cache actually exist on disk. The way we determine `job in project` is to check for the existence of a directory with the job's id.

signac/contrib/project.py
bdice: I see some lines aren't in the test coverage but I don't think there is very much that is actionable / worth doing. The changes here are just a re-work of the previous code and the crucial parts (fetching from cache, fetching from workspace, fetching from a file of state points) are all covered. https://codecov.io/gh/glotzerlab/signac/pull/239/src/signac/contrib/project.py

The lines I see that are missing coverage here are:
- Some error checking if the fallback to `read_statepoints` fails with an `OSError` (this is hard to test)
- The cache miss warning (this might be testable but it's just a logging message)

signac/contrib/job.py
bdice: This `OSError` is not something I know how to test. This is probably raised by things like hardware issues, insufficient filesystem privileges, or dropped network connections.

signac/contrib/job.py
```diff
@@ -84,12 +84,21 @@ class Job:
...
 
-    def __init__(self, project, statepoint, _id=None):
+    def __init__(self, project, statepoint=None, _id=None):
         self._project = project
 
-        # Set statepoint and id
-        self._statepoint = SyncedAttrDict(statepoint, parent=_sp_save_hook(self))
-        self._id = calc_id(self._statepoint()) if _id is None else _id
+        if statepoint is None and _id is None:
+            raise ValueError("Either statepoint or _id must be provided.")
```
bdice, line 91: This _should_ be tested.

signac/contrib/project.py
klywang: Okay that makes sense!

klywang:
> @klywang The first comment was a misunderstanding on my part about the expectations of the data model. The best explanation is here from Simon:
> 
> > No, it's fine to be able to open by id, even if there is no job in the project if the state point happens to be in some lookup table like the cache. It's not ok to open some arbitrary job by id where we wouldn't know the state point.
> 
> This explains the job.remove() behavior, which should not remove items from the cache. Once a job is registered, the state point cache will track the id: statepoint in-memory until the end of the session. This means that a job can be initialized, removed, and re-opened by id. The project's state point cache is simply a lookup table of id: statepoint and does not mean that all job ids in the cache actually exist on disk. The way we determine job in project is to check for the existence of a directory with the job's id.

Thank you for the clarification!

cbkerr (changes requested):
Thanks for taggging me @bdice; I&rsquo;ve learned a lot by starting to review this!

Here are some initial questions.

I&rsquo;ve numbered my comments so you can respond inline easily.

## In contrib/job.py

1.  Is there anything stopping the initialization of a corrupted Job instance by providing a `statepoint` and `_id` such that `calc_id(statepoint) != _id`? The only related check I see for this is in the  `project.open_job()` method:
    
        if (statepoint is None) == (id is None):
           raise ValueError("Either statepoint or id must be provided, but not both.")
    
    But still one of the lines calls `Job()` with both `statepoint` and `_id` arguments.
    
    -   Possible idea to try to fix this, to replace the line `self._id = calc_id(self.statepoint()) if _id is None else _id`.
        
            # Set id.
            if _id is not None:
                if statepoint is None:
                    # The id must have been provided
                    self._id = _id
                else:
                    # make sure provided _id and statepoint are consistent
                    if calc_id(statepoint) != _id:
                        raise JobsCorruptedError([_id]) # or maybe ValueError
                    else:
                        self._id = _id     
            else:
                # The id is computed from the state point if not provided.
                self._id = calc_id(self.statepoint())
2.  I see that `self.statepoint()` calls `_check_manifest()`, which looks at `self.id`&#x2026;but what is `self.id` before it is set when the Job is instanced? Is this a problem when initializing the Job?
3.  In the code `self._id = calc_id(self.statepoint())`, do we use the getter method rather than the provided `statepoint` argument to the `__init__` method in case the job has been initialized before? It seems like an extra system call because the getter method calls `job._check_manifest()`.
4.  What&rsquo;s the difference between the errors raised in `job._read_manifest()`? In the docstring, both say the same thing: &ldquo;If an error occurs while reading/parsing the state point manifest.&rdquo; The solution here might be to not raise a JobsCorruptedError when catching a ValueError?
5.  The docstring describing the JobsCorruptedError in `_check_manifest` sounds like a combination of OSError and JobsCorruptedError:
    
         If the manifest hash is not equal to the job id, or if an error occurs while reading/parsing the state point manifest.
    
    Something is inconsistent between this bullet and the previous bullet. Maybe the docstring for `_check_manifest` should just say &ldquo;If the manifest hash is not equal to the job id.&rdquo; because `_check_manfiest()` calls `_read_manifest()` which would raise the OSError.

## In contrib/project.py

1.  This is slightly related to the [confusion over opening a job by an id not in the project](https://github.com/glotzerlab/signac/pull/239/#issuecomment-556069852). Is the following true?: If you open a job by id that doesn&rsquo;t match any existing job using
    
        fail_job = project.open_job('f'*32)
    
    the only error will be when you try to access `fail_job.statepoint()` (because it calls `_check_manifest()`) which would be **after** opening the job? I think we need to document this better or have it fail earlier.

signac/contrib/project.py
cbkerr: Since this PR seems almost done, my hunch would be to leave the aggressive solution for a follow-up PR. But I'm still understanding this PR and can comment better when I finish my review.

bdice:
Thanks for the review, @cbkerr! Responses inline below.

>     1. Is there anything stopping the initialization of a corrupted Job instance by providing a `statepoint` and `_id` such that `calc_id(statepoint) != _id`? The only related check I see for this is in the  `project.open_job()` method:
>        ```
>        if (statepoint is None) == (id is None):
>           raise ValueError("Either statepoint or id must be provided, but not both.")
>        ```

We guard against issues like this in a few ways: firstly, users are not supposed to call the `Job` constructor directly, they are expected to get jobs through `open_job` or by iterating / searching over a project. The only case in `open_job` where the `Job` is directly constructed with both `id` and `statepoint` is when opening by id and finding a state point in the project's state point cache. Here, we _do_ assume that the state point cache is a valid source of truth, because the state point cache data always comes from other presumed-valid sources of truth such as a cache on disk or an instantiated `Job` that has registered its id and state point with the project (once it has been lazily loaded and validated).

>        
>        But still one of the lines calls `Job()` with both `statepoint` and `_id` arguments.
>        
>        * Possible idea to try to fix this, to replace the line `self._id = calc_id(self.statepoint()) if _id is None else _id`.
>          ```
>          # Set id.
>          if _id is not None:
>              if statepoint is None:
>                  # The id must have been provided
>                  self._id = _id
>              else:
>                  # make sure provided _id and statepoint are consistent
>                  if calc_id(statepoint) != _id:
>                      raise JobsCorruptedError([_id]) # or maybe ValueError
>                  else:
>                      self._id = _id     
>          else:
>              # The id is computed from the state point if not provided.
>              self._id = calc_id(self.statepoint())
>          ```

There's a bit of nuance here: if opening by state point, we must go through the round-trip of encoding/decoding the state point _before_ we can calculate the id. This prevents the ea The argument to `calc_id` _must_ be JSON-encodable using the default encoder, while the provided `statepoint` argument may need to be altered (e.g. converting NumPy arrays to lists) by the custom JSON encoder class used by `SyncedAttrDict` and related classes.

We choose to trust the caller of the Job constructor that `calc_id(statepoint) == id`, because we can internally guarantee that in `open_job` and that's the only place the constructor is called. There is already a note in the docs of the `Job` class to suggest this and I will expand that documentation to address this specifically.

>     2. I see that `self.statepoint()` calls `_check_manifest()`, which looks at `self.id`…but what is `self.id` before it is set when the Job is instanced? Is this a problem when initializing the Job?

This proposed case is not reachable in the code (but it may be possible to improve the code). The line `self._id = calc_id(self.statepoint()) if _id is None else _id` only triggers `self.statepoint()` if `_id is None`. However, because of the first check requiring id-or-statepoint, `_id is None` is only possible if a state point is provided. Thus `self.statepoint` does not execute `_check_manifest` and simply returns `self._statepoint`.

Would this be clearer?

```python
        if statepoint is None and _id is None:
            raise ValueError("Either statepoint or _id must be provided.")
        elif statepoint is not None:
            # Set state point if provided
            self._statepoint = SyncedAttrDict(statepoint, parent=_sp_save_hook(self))
            # Set id. The id is computed from the state point if not provided.
            self._id = calc_id(self.statepoint()) if _id is None else _id
        else:
            # State point will be loaded lazily
            self._statepoint = None
            self._id = _id
```

>     3. In the code `self._id = calc_id(self.statepoint())`, do we use the getter method rather than the provided `statepoint` argument to the `__init__` method in case the job has been initialized before? It seems like an extra system call because the getter method calls `job._check_manifest()`.

Ah, this is confusing. `self.statepoint()` accesses the property `self.statepoint` and then uses its `__call__` method, which returns a "plain dict" instead of a synced dict. It looks like a getter but it's not. As mentioned above, the `Job.__init__` method cannot trigger `Job._check_manifest` by its construction. Only opening by id and lazy-loading the state point can trigger that method. For efficiency, all state points that are pulled from the project's in-memory cache or a disk cache are trusted to be correct (and not corrupted) on-disk. The `Project.check()` method can be used to explicitly verify all state points on disk.

>     4. What’s the difference between the errors raised in `job._read_manifest()`? In the docstring, both say the same thing: “If an error occurs while reading/parsing the state point manifest.” The solution here might be to not raise a JobsCorruptedError when catching a ValueError?

These errors are triggered by different causes, but I couldn't figure out how to explain the difference concisely. The error-raising behavior should be kept the same as before (so it's not a breaking change), but the documentation could be improved. Here's an explanation of potential causes: `OSError` is difficult to test and likely results from something like hardware failure, permissions issues, etc. while `JobsCorruptedError` is raised by `json.loads` as a `JSONDecodeError` (which is a subclass of `ValueError` -- we choose to catch the `ValueError` for generality and because it saves us from needing to import that specific error class). A file that can be read but isn't valid JSON will raise `ValueError`, which will be caught, and raise a `JobsCorruptedError`. A file that can't be read will raise `OSError`. I welcome your thoughts on ways to improve the docs here.

>     5. The docstring describing the JobsCorruptedError in `_check_manifest` sounds like a combination of OSError and JobsCorruptedError:
>        ```
>         If the manifest hash is not equal to the job id, or if an error occurs while reading/parsing the state point manifest.
>        ```
>
>        Something is inconsistent between this bullet and the previous bullet. Maybe the docstring for `_check_manifest` should just say “If the manifest hash is not equal to the job id.” because `_check_manfiest()` calls `_read_manifest()` which would raise the OSError.

Edited. You're correct that this should only say "If the manifest hash is not equal to the job id."

> ## In contrib/project.py
> 
>     1. This is slightly related to the [confusion over opening a job by an id not in the project](https://github.com/glotzerlab/signac/pull/239/#issuecomment-556069852). Is the following true?: If you open a job by id that doesn’t match any existing job using
>        ```
>        fail_job = project.open_job('f'*32)
>        ```
>        
>        
>        the only error will be when you try to access `fail_job.statepoint()` (because it calls `_check_manifest()`) which would be **after** opening the job? I think we need to document this better or have it fail earlier.

No, that's not a problem in the current implementation. If you open by an (invalid) id, it will check `Project._contains_job_id` which ensures that a directory exists in the workspace and raises `KeyError` if not. However, it does _not_ verify that a state point file exists in that directory. This means that an empty directory `ffffffffffffffffffffffffffffffff` could be opened as a job, but would raise an error when its state point is accessed. I believe this is a desired and acceptable outcome. If that happens, a user has broken an invariant of the signac data model (in this case, "all job workspace directories must contain a manifest and be named according to the hash of that manifest"). We haven't formalized our data model in a document but it might be a good idea (currently, our implementation _is_ the data model specification).

signac/contrib/job.py
```diff
@@ -104,6 +114,10 @@ def __init__(self, project, statepoint, _id=None):
         # Prepare current working directory for context management
         self._cwd = []
 
+        if statepoint is not None:
+            # Update the project's state point cache immediately if opened by state point
+            self._project._register(self)
```
csadorf, line 119: Should this be run directly after line 96?

csadorf (approved):
Great! 👍

cbkerr (approved):
Thank you for your detailed answers @bdice!

After addressing these small comment/docs things, it looks good to me!

1.  New doc in Jobs class looks good.
2.  Ah yes, and that conditional structure you suggest is more clear! The new structure can also address @csadorf&rsquo;s [comment](https://github.com/glotzerlab/signac/pull/239#discussion_r550069934) by moving the line `self._project._register(self)` inside the proper condition.
3.  Let&rsquo;s add a comment at the line I was confused about (`self._id = calc_id(self.statepoint()) if _id is None else _id`) to remind of this. Maybe something like?:
    
        # Simply returns the (properly JSON encoded) statepoint if provided already (no _check_manifest)
4.  Thanks for the explanation. To fix the documentation of the kinds of errors, I think we just need to separate reading and parsing like:
    
        JobsCorruptedError
              If an error occurs while parsing the state point manifest.
        OSError
              If an error occurs while reading the state point manifest.
    
    It would also help to have a comment to remind that JSONDecodeError is a subclass of ValueError at the line where we catch that.

project.py 
1.  Got it, so basically, if the user manually (outside of signac) creates a directory `'f'*32`, then opens the job `'f'*32`, they&rsquo;ve created their own trouble.

signac/contrib/job.py
bdice: This is fixed in #451.

bdice:
Thanks again, @csadorf and @cbkerr. I have addressed your suggestions in this PR or #451.

PR #241, opened 2019-11-07
Fix/circle ci pypy3 issues

csadorf:
I have updated the circle-ci configuration to fix an issue with the pypy3 dependency installation, an issue with the implementation of the `H5StoreTest` test class, and am no longer executing benchmarks for pypy.

The latter is mostly because we are using the following command to execute the benchmark: `python benchmark.py ...`, however that should probably say `pypy3 benchmark.py ...`. I'm not sure if there is a good way to easily execute the benchmark with both platforms without too much duplication of the CI-configuration, I'm all for it.

Usually, I'd have just merged these commits without review, but because I am making a bunch of more intrusive changes, I'd appreciate some feedback.

**I'd prefer if this PR was not squashed prior to merge.**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,6 +24,9 @@ jobs:
     docker:
       - image: circleci/python:3.7
 
+    environment:
+      PYTHON: python
+
     working_directory: ~/repo
 
     steps:
@@ -42,7 +45,7 @@ jobs:
           name: install dependencies
           command: |
             export PATH=$PATH:$HOME/.local/bin
-            pip install -U virtualenv --user
+            ${PYTHON} -m pip install -U virtualenv --user
             mkdir -p ./venv
             virtualenv ./venv
             . venv/bin/activate
@@ -51,9 +54,9 @@ jobs:
             else
               sudo apt update -qq && sudo apt install -y -qq libhdf5-dev
             fi
-            pip install -U pip==18
-            pip install -U codecov
-            pip install -U -e . -r requirements-dev.txt
+            ${PYTHON} -m pip install -U pip==18
+            ${PYTHON} -m pip install -U codecov
+            ${PYTHON} -m pip install -U -e . -r requirements-dev.txt
 
       - save_cache:
           key: python-env-v4-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "requirements-dev.txt" }}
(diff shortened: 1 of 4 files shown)


bdice:
@csadorf Looks like you might have made some commits with a different name/email than usual, since it doesn't show your user icon on all of them. Feel free to force-push a fix for that if you want.

.circleci/config.yml
```diff
@@ -98,6 +98,67 @@ jobs:
...
+      - restore_cache:
+          keys:
+            - python-benchmark-env-v1-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "requirements-benchmark.txt" }}
+            - python-benchmark-env-v1-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}
+            - python-benchmark-env-v1-{{ arch }}-{{ .Environment.CIRCLE_JOB }}
+            - python-benchmark-env-v1-{{ arch }}
+            - python-benchmark-env-v1
+
+      - run:
+          name: install dependencies
```
bdice: If this is exactly the same as the other "install dependencies" above, it would be ideal to combine them into a single variable-thing that you can repeat with YAML.

csadorf:
> @csadorf Looks like you might have made some commits with a different name/email than usual, since it doesn't show your user icon on all of them. Feel free to force-push a fix for that if you want.

I think it would be better to introduce a mailmap, since I'll probably be committing more under that different email.

bdice:
> 
> 
> > @csadorf Looks like you might have made some commits with a different name/email than usual, since it doesn't show your user icon on all of them. Feel free to force-push a fix for that if you want.
> 
> I think it would be better to introduce a mailmap, since I'll probably be committing more under that different email.

Ok, feel free to do that in this PR or a follow-up. If you make a `.mailmap`, please try to de-duplicate everyone. You can copy from freud's `.mailmap` if it's helpful.

vyasr:
One option for dealing with the need for different python executables is to set it as an environment variable for each benchmark job. Then you would change the benchmark CI job to use `${PYTHON} benchmark.py ...`.

csadorf:
> One option for dealing with the need for different python executables is to set it as an environment variable for each benchmark job. Then you would change the benchmark CI job to use `${PYTHON} benchmark.py ...`.

Good call, I've updated (and thus simplified) the configuration patch using the environment variable.

csadorf:
I've ended up squashing it anyways, because updating the commit message to be less misleading seemed like more work. Thank's everyone for helping!

PR #244, opened 2019-11-20
Make workspace a settable property. Define ProjectWorkspace for backwards compatibility

### LINKED ISSUE #81: Better API for the modification of configuration files

vyasr:
## Description
Makes the workspace of a project modifiable by a property API.

## Motivation and Context
Resolves #81. The config is already accessible by property, it's just the workspace that isn't yet.

I do have one question about this feature, and also the `config` property of the project. Currently any change to the config, including changing the workspace, modifies the in-memory project object, but that change will not persist beyond the current session since the underlying config file isn't updated. Considering that the command line interface for config modification updates the underlying file, and also the way that other things like statepoint/job document modification always persist to the underlying file, is that the expected behavior?

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -95,6 +95,28 @@ class JobSearchIndex(object):
         return self._collection._find(filter)
 
 
+class ProjectWorkspace(object):
+    """Lightweight class for operating on project workspaces."""
+    # This class is primarily defined for backwards compatibility because
+    # `workspace` is a method rather than a property of the Project class. Use
+    # of this class ensures that calling it like a method gives the expected
+    # behavior. It should be removed in version 2.0 when we make the workspace
+    # a property instead of a method.
+    def __init__(self, project):
+        self._project = project
+
+    def __repr__(self):
+        return self()
+
+    def __call__(self):
+        wd = os.path.expandvars(
+            self._project._config.get('workspace_dir', 'workspace'))
+        if os.path.isabs(wd):
+            return wd
+        else:
+            return os.path.join(self._project._rd, wd)
+
+
 class Project(object):
     """The handle on a signac project.
 
@@ -168,18 +190,11 @@ class Project(object):
         "The project root directory."
         return self._config['project_dir']
 
-    @property
-    def _wd(self):
-        wd = os.path.expandvars(self._config.get('workspace_dir', 'workspace'))
-        if os.path.isabs(wd):
-            return wd
-        else:
-            return os.path.join(self._rd, wd)
-
     def root_directory(self):
         "Returns the project's root directory."
         return self._rd
 
+    @property
     def workspace(self):
         """Returns the project's workspace directory.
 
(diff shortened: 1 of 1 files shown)


vyasr:
I literally just saw that this was mentioned as possibly undesirable in #239. If that is the case, I can consider alternatives for enabling this kind of feature. I do find it useful to be able to modify the workspace, but maybe there's a compromise. This doesn't immediately conflict with the suggestion there since that is a proposed future optimization, but it would preclude that optimization without making an API-breaking change so we can hold off with this PR until that's decided.

bdice:
I think it would be totally fine to require configuration settings to be changed by the command line interface, and for configurations to be immutable once a project is instantiated in Python. Allowing project configs to change in memory is a great way to introduce potential bugs, in my opinion. Things like caches can't be used safely without a static config. If you need another workspace in Python, just create a config object and create a new Project instance from that config.

csadorf:
Alternatively you just make the cache workspace path dependent. That would be very easy to realize. It probably would actually be critical, since we already allow an in-memory change of the configuration.

vyasr:
@csadorf I'm not sure I follow your suggestion.

bdice:
@csadorf Yes, incorrect caching behavior is probably already possible. I think it makes more sense to make things immutable than to enable multi-workspace caching. I benchmarked this for #239 and saw that signac spends around [~13% of its time](https://github.com/glotzerlab/signac/pull/239#issuecomment-555250864) just calling `project.workspace()`, which is silly for a value that is constant in 99.9% of real world use cases.

csadorf:
> @csadorf Yes, incorrect caching behavior is probably already possible. I think it makes more sense to make things immutable than to enable multi-workspace caching. I benchmarked this for #239 and saw that signac spends around [~13% of its time](https://github.com/glotzerlab/signac/pull/239#issuecomment-555250864) just calling `project.workspace()`, which is silly for a value that is constant in 99.9% of real world use cases. Immutable configs would solve this.

Agreed.

vyasr:
So what's the expected path forward? Deprecate the ability to update the project config, and remove in 2.0? And otherwise close #81 as wontfix?

csadorf:
> So what's the expected path forward? Deprecate the ability to update the project config, and remove in 2.0? And otherwise close #81 as wontfix?

I'd think so.

vyasr:
I'd implement that by subclassing the `Config` class (and making e.g. an `ImmutableConfig`) that for now throws a `DeprecationWarning` in `__setitem__` that will become an error in 2.0. Sound sufficient?

vyasr:
This PR will be closed when #246 is merged.

vyasr:
Closing, replaced by #246.

PR #246, opened 2019-11-21
Fix/make project config immutable

### LINKED ISSUE #81: Better API for the modification of configuration files

vyasr:
## Description
Makes the project config immutable. Note that I had to make the `_mutable` attribute set after construction because the constructor of `ConfigObj` calls `__setitem__` internally, so constructing the `ProjectConfig` object would always throw warnings (or error in the future). I'm open to alternative suggestions on how to resolve that.

## Motivation and Context
Resolves #81, supersedes #244 

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:
- [x] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -35,6 +35,11 @@ Fixed
 
  - Attempting to create a linked view for a Project on Windows now raises an informative error message.
 
+Deprecated
+++++++++++
+
+ - Modifying the project config, to be removed in 2.0 (#246).
+
 Removed
 +++++++
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -38,7 +38,7 @@ Fixed
 Deprecated
 ++++++++++
 
- - Modifying the project config, to be removed in 2.0 (#246).
+ - In-memory modification of the project configuration, to be removed in 2.0 (#246).
 
 Removed
 +++++++
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -24,7 +24,7 @@ from ..core import json
 from ..core.jsondict import JSONDict
 from ..core.h5store import H5StoreManager
 from .collection import Collection
-from ..common.config import load_config
+from ..common.config import load_config, Config
 from ..sync import sync_projects
 from .job import Job
 from .hashing import calc_id
@@ -95,6 +95,15 @@ class JobSearchIndex(object):
         return self._collection._find(filter)
 
 
+class ProjectConfig(Config):
+    """Extends the project config to make it immutable."""
+    def __setitem__(self, key, value):
+        warnings.warn("Modifying the project configuration is deprecated "
+                      "behavior and will be removed in version 2.0.",
+                      DeprecationWarning)
+        return super(ProjectConfig, self).__setitem__(key, value)
+
+
 class Project(object):
     """The handle on a signac project.
 
@@ -120,7 +129,7 @@ class Project(object):
     def __init__(self, config=None):
         if config is None:
             config = load_config()
-        self._config = config
+        self._config = ProjectConfig(config)
 
         # Ensure that the project id is configured.
         self.get_id()
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -95,11 +95,11 @@ class JobSearchIndex(object):
         return self._collection._find(filter)
 
 
-class ProjectConfig(Config):
+class _ProjectConfig(Config):
     """Extends the project config to make it immutable."""
     def __init__(self, *args, **kwargs):
         self._mutable = True
-        super(ProjectConfig, self).__init__(*args, **kwargs)
+        super(_ProjectConfig, self).__init__(*args, **kwargs)
         self._mutable = False
 
     def __setitem__(self, key, value):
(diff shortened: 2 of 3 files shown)


signac/contrib/project.py
```diff
@@ -95,6 +95,20 @@ def find_job_ids(self, filter=None, doc_filter=None):
...
 
+class ProjectConfig(Config):
+    """Extends the project config to make it immutable."""
+    def __init__(self, *args, **kwargs):
+        self._mutable = True
+        super(ProjectConfig, self).__init__(*args, **kwargs)
+
+    def __setitem__(self, key, value):
+        if not self._mutable:
+            warnings.warn("Modifying the project configuration is deprecated "
```
csadorf: 1. Let's use the standard language for other deprecations, i.e., "deprecated as of version 1.3 ..."
2. I'd try to make clear that modifying the project configuration *in this way* is deprecation.

So maybe something like this:

> Modifying the project configuration after project initialization is deprecated as of version 1.3 and will be removed in version 2.0.

signac/contrib/project.py
```diff
@@ -95,6 +95,20 @@ def find_job_ids(self, filter=None, doc_filter=None):
         return self._collection._find(filter)
 
 
+class ProjectConfig(Config):
+    """Extends the project config to make it immutable."""
+    def __init__(self, *args, **kwargs):
+        self._mutable = True
+        super(ProjectConfig, self).__init__(*args, **kwargs)
```
csadorf: Just call the super constructor first and then set the flag.

signac/contrib/project.py
```diff
@@ -120,7 +134,8 @@ class Project(object):
     def __init__(self, config=None):
         if config is None:
             config = load_config()
-        self._config = config
+        self._config = ProjectConfig(config)
+        self._config._mutable = False
```
csadorf: You should be able to avoid manipulating the flag like this by rearranging the constructor calls as outlined above.

signac/contrib/project.py
vyasr: And this is why I shouldn't make PRs late at night. Thanks.

signac/contrib/project.py
```diff
@@ -95,6 +95,22 @@ def find_job_ids(self, filter=None, doc_filter=None):
...
+        self._mutable = True
+        super(ProjectConfig, self).__init__(*args, **kwargs)
+        self._mutable = False
+
+    def __setitem__(self, key, value):
+        if not self._mutable:
+            warnings.warn("Modifying the project configuration after project "
+                          "initialization is deprecated as of version 1.3 and "
+                          "will be removed in version 2.0.",
+                          DeprecationWarning)
```
csadorf: We could add a unit tests and the following assertion, so that we don't forget to actually remove this with 2.0:
```suggestion
                          DeprecationWarning)
    from packaging import version
    assert  version.parse(__version__) < version.parse("2.0")
```

vyasr: What would you want the unit test for? Adding both a unit test and the assertion seems redundant to me, unless you mean adding a unit test that for now checks for a warning and later does an `assertRaises`. Since removing config modification will require modifying a few tests anyway, I'd prefer to just hold off on adding that test until 2.0 when we update all the tests for this change.

csadorf: If we add the assertion we need to make sure that this function is actually executed in *any* of the tests. Otherwise we risk that we miss this and our code starts to fail with an `AssertionException`. The point is that the test fails with 2.0 and we remember to remove this.

A natural test would be to simply access and optionally attempt to modify the configuration.

changelog.txt
```diff
@@ -35,6 +35,11 @@ Fixed
 
  - Attempting to create a linked view for a Project on Windows now raises an informative error message.
 
+Deprecated
+++++++++++
+
+ - Modifying the project config, to be removed in 2.0 (#246).
```
csadorf:
```suggestion
 - In-memory modification of the project configuration, to be removed in 2.0 (#246).
```
Otherwise it's a bit too vague and might imply that you would not be able to change the project configuration at all.

signac/contrib/project.py
```diff
@@ -95,6 +95,25 @@ def find_job_ids(self, filter=None, doc_filter=None):
         return self._collection._find(filter)
 
 
+class ProjectConfig(Config):
```
csadorf: I'd suggest to make this internal for now: `_ProjectConfig`.

tests/test_project.py
```diff
@@ -74,6 +74,12 @@ def test_root_directory(self):
     def test_workspace_directory(self):
         self.assertEqual(self._tmp_wd, self.project.workspace())
 
+    def test_config_modification(self):
```
csadorf:
```suggestion
    def test_config_modification(self):
        # In-memory modification of the project configuration is 
        # deprecated as of 1.3, and will be removed in version 2.0.
        # This unit test should reflect that change beginning 2.0,
        # and check that the project configuration is immutable.
```

Let's make it easier for future devs. :)

PR #247, opened 2019-11-22
adding signac diff functionality

### LINKED ISSUE #248: signac diff
### Feature description

Add the functionality to call `signac diff` to compare two or more state points. This could also be extended to comparing files, job docs, etc. 

### Proposed solution

partially solved at #247 

### Additional context

Any input from @bdice @csadorf @vyasr @b-butler @mikemhenry is very welcome!

atravitz:
## Description
compares two or more jobs, returns jobs ids with only their keys that are not shared by all other jobs.

## Motivation and Context
several requests for this functionality, often when two state points only vary by one or two values.

Resolves #248.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


vyasr:
Discussion on the appropriate scope and API of this PR is happening at #248. Once everything there is finalized we can update this PR.

signac/__main__.py
```diff
@@ -1120,6 +1140,45 @@ def main():
...
+    parser_diff.add_argument(
+        '-f', '--filter',
+        type=str,
+        nargs='+',
+        help="Limit the diff to jobs matching this state point filter.")
+    parser_diff.add_argument(
+        '-d', '--doc-filter',
+        type=str,
+        nargs='+',
+        help="Show documents of job matching this document filter.")
```
bdice:
```suggestion
        help="Show documents of jobs matching this document filter.")
```

signac/diff.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2017 The Regents of the University of Michigan
```
bdice:
```suggestion
# Copyright (c) 2019 The Regents of the University of Michigan
```

signac/diff.py
```diff
@@ -0,0 +1,27 @@
...
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+import signac
```
bdice: Use relative imports. Avoid absolute imports like `import signac` inside the `signac` package directory. Something like this would work:
```python
from .contrib.collection import _traverse_filter
```

signac/diff.py
```diff
@@ -0,0 +1,27 @@
...
+
+        :param jobs:
+    """
+    if len(jobs) == 0:
+        return {}
+
+    else:
+        sps = {}
+        for job in jobs:
+            sps[job] = set(signac.contrib.collection._traverse_filter(job.sp))
```
bdice:
```suggestion
            sps[job] = set(_traverse_filter(job.sp))
```

signac/diff.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2017 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+import signac
+
+def diff_jobs(*jobs):
+    """Find the difference between jobs or an entire project.
+
+        :param jobs:
```
bdice:
```suggestion
        :param jobs: One or more jobs whose state points will be diffed.
```

signac/diff.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2017 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+import signac
+
+def diff_jobs(*jobs):
+    """Find the difference between jobs or an entire project.
```
bdice:
```suggestion
    """Find differences among a list of jobs' state points.
```

tests/test_diff.py
```diff
@@ -0,0 +1,64 @@
...
+        self.addCleanup(self._tmp_dir.cleanup)
+        self._tmp_pr = os.path.join(self._tmp_dir.name, 'pr')
+        self._tmp_wd = os.path.join(self._tmp_dir.name, 'wd')
+        os.mkdir(self._tmp_pr)
+        self.config = signac.common.config.load_config()
+        self.project = self.project_class.init_project(
+            name='testing_test_project',
+            root=self._tmp_pr,
+            workspace=self._tmp_wd)
+        self.project.config['default_host'] = 'testing'
```
bdice: Not needed anymore.
```suggestion
```

tests/test_diff.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def tearDown(self):
+        pass
+
+    def open_job(self, *args, **kwargs):
```
bdice: I don't think we need to define this. Remove this method and use `self.project.open_job` instead of `self.open_job` (you might already be doing this everywhere).

tests/test_shell.py
```diff
@@ -251,6 +251,5 @@ def test_shell_with_jobs_and_selection_only_one_job(self):
         job = list(project.find_jobs({'a': 0}))[0]
         self.assertEqual(out.strip(), '>>> {} {} 1'.format(project, job))
 
-
```
bdice: Revert this -- you need two newlines at the top level indent, for flake8.

tests/test_diff.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class DiffTest(BaseDiffTest):
+
+    def test_two_jobs(self):
+        job1 = self.project.open_job({'a': 0, 'b': 1})
+        job2 = self.project.open_job({'a': 0})
+        expected = {str(job1.id): {'b': 1}, str(job2.id): {}}
+        result = signac.diff_jobs(job1, job2)
+        self.assertEqual(expected, result, f'{result} is not {expected}')
```
bdice: Unfortunately we can't use format strings on Python 3.5, which we still support.
```suggestion
        self.assertEqual(expected, result, '{} is not {}'.format(result, expected))
```

tests/test_diff.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def test_one_job(self):
+        job1 = self.project.open_job({'a': 0})
+        expected = {str(job1.id): {}}
+        result = signac.diff_jobs(job1)
+        self.assertEqual(expected, result, f'{result} is not {expected}')
```
bdice:
```suggestion
        self.assertEqual(expected, result, '{} is not {}'.format(result, expected))
```

tests/test_diff.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def test_no_jobs(self):
+        self.assertTrue(signac.diff_jobs() == {})
+
+    def test_nested(self):
+        job1 = self.project.open_job({'a': 0, 'b': {'c': True, 'd': 11}})
+        job2 = self.project.open_job({'a': 0, 'b': {'c': True, 'd': 4}})
+        expected = {str(job1.id): {'b': {'d': 11}}, str(job2.id): {'b': {'d': 4}}}
+        result = signac.diff_jobs(job1, job2)
+        self.assertEqual(expected, result, f'{result} is not {expected}')
```
bdice:
```suggestion
        self.assertEqual(expected, result, '{} is not {}'.format(result, expected))
```

tests/test_diff.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def test_same_job(self):
+        job1 = self.project.open_job({'a': 0, 'b': 1})
+        expected = {str(job1.id): {}}
+        result = signac.diff_jobs(job1, job1)
+        self.assertEqual(expected, result, f'{result} is not {expected}')
```
bdice:
```suggestion
        self.assertEqual(expected, result, '{} is not {}'.format(result, expected))
```

signac/diff.py
```diff
@@ -0,0 +1,50 @@
...
+        keys = dotted_key.split('.')
+        make_nested_dict(result, keys)[keys[-1]] = value
+
+    return result
+
+
+def diff_jobs(*jobs):
+    """Find differences among a list of jobs' state points.
+
+        :param jobs: One or more jobs whose state points will be diffed.
```
bdice: We need to add a line `:returns:` to this docstring.

bdice: Also the formatting of this `:param jobs:` implies that `jobs` is a specific argument, while we need it to be clear that it's a variable number of arguments.

This is somewhat difficult to make work correctly in the docs -- to the best of my recollection, the correct way to do it is:

```python
r"""Find differences among a list of jobs' state points.

The resulting diff is a dictionary where the keys are job ids and the values are each job's state point minus the intersection of all provided jobs' state points.

:param \*jobs: One or more jobs whose state points will be diffed.
:returns: A dictionary whose keys are job ids and whose values are the unique parts of that job's state point.
"""
```
Notice that we have to use a raw string with the `r"""` at the beginning, so that the backslash is interpreted as a literal backslash.

bdice:
Reviewers @vyasr @cbkerr and author @atravitz: I made some final touch-ups to this PR, and I am satisfied with the feature's quality and completeness for merging to `master`. I will be making a release tomorrow afternoon, probably around 5pm. If you have any review comments, it would be great to hear those and incorporate any needed changes before then.

signac/__main__.py
```diff
@@ -1120,6 +1138,44 @@ def main():
...
+        help="Limit the diff to jobs matching this state point filter.")
+    parser_diff.add_argument(
+        '-d', '--doc-filter',
+        type=str,
+        nargs='+',
+        help="Show documents of jobs matching this document filter.")
+    parser_diff.add_argument(
+        '-s', '--sort',
+        action='store_true',
+        help="Sort the state point keys for output.")
```
vyasr: Is the sorting actually implemented? If so, I'd remove this option for now.

signac/__main__.py
```diff
@@ -1098,8 +1117,7 @@ def main():
         'job_id',
         nargs='*',
         type=str,
-        help="One or more job ids. The job corresponding to a job "
-             "id must be initialized.")
+        help="One or more job ids. The jobs must be initialized.")
```
vyasr: "... The _corresponding_ jobs must be initialized." (Please also add that word to the other docstrings that you changed).

signac/diff.py
```diff
@@ -0,0 +1,82 @@
...
+    >>> print(job3)
+    e4289419d2b0e57e4852d44a09f167c0
+    >>> signac.diff_jobs(job1, job2, job3)
+    {'c4af2b26f1fd256d70799ad3ce3bdad0': {'diff2': 1, 'diff1': 0},
+     'b96b21fada698f8934d58359c72755c0': {'diff2': 1, 'diff1': 1},
+     'e4289419d2b0e57e4852d44a09f167c0': {'diff2': 2, 'diff1': 2}}
+    >>> signac.diff_jobs(*project)
+    {'c4af2b26f1fd256d70799ad3ce3bdad0': {'diff2': 1, 'diff1': 0},
+     'b96b21fada698f8934d58359c72755c0': {'diff2': 1, 'diff1': 1},
+     'e4289419d2b0e57e4852d44a09f167c0': {'diff2': 2, 'diff1': 2}}
```
vyasr: This entire example needs to be indented and placed in a `.. code-block:: python` to be rendered correctly.

tests/test_diff.py
```diff
@@ -0,0 +1,63 @@
...
+
+    project_class = signac.Project
+
+    def setUp(self):
+        self._tmp_dir = TemporaryDirectory(prefix='signac_')
+        self.addCleanup(self._tmp_dir.cleanup)
+        self._tmp_pr = os.path.join(self._tmp_dir.name, 'pr')
+        self._tmp_wd = os.path.join(self._tmp_dir.name, 'wd')
+        os.mkdir(self._tmp_pr)
+        self.config = signac.common.config.load_config()
```
vyasr: Do you need to load this? It's not being used.

tests/test_diff.py
```diff
@@ -0,0 +1,63 @@
...
+class BaseDiffTest(unittest.TestCase):
+
+    project_class = signac.Project
+
+    def setUp(self):
+        self._tmp_dir = TemporaryDirectory(prefix='signac_')
+        self.addCleanup(self._tmp_dir.cleanup)
+        self._tmp_pr = os.path.join(self._tmp_dir.name, 'pr')
+        self._tmp_wd = os.path.join(self._tmp_dir.name, 'wd')
+        os.mkdir(self._tmp_pr)
```
vyasr: I'm confused as to why these extra directories are being made. You can use the `_tmp_dir` as the project root, and the workspace will be created for you by `signac`, so you should be able to remove the above three lines.

tests/test_diff.py
```diff
@@ -0,0 +1,63 @@
...
+        self.config = signac.common.config.load_config()
+        self.project = self.project_class.init_project(
+            name='diff_test_project',
+            root=self._tmp_pr,
+            workspace=self._tmp_wd)
```
vyasr: Remove the workspace and just set the root to `self._tmp_dir`.

tests/test_diff.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def tearDown(self):
+        pass
```
vyasr: No need to define this function if you're not doing anything with it.

vyasr: Note that the `addCleanup` in `setUp` should be wiping the project directories for you, so it's fine to ignore the `tearDown` method.

vyasr:
Regarding the different use-cases outlined in #248, I think the existing API is sufficiently general to allow this. We would simply need to add an extra argument to the Python and CLI such as `diff_by` that would default to statepoint (which I think is the sensible default for diffing jobs). The output would be of the same form, except that if we chose to diff files the leaf nodes of the dict would probably have to be a list of files instead of a mapping of key-value pairs. Document and data diffing should be transparently supported using the same API and output structure.

signac/__main__.py
atravitz: that was my original intention, but I didn't get around to it and I'll take it out for now.

signac/__main__.py
```diff
@@ -384,6 +385,24 @@ def format_lines(cat, _id, s):
...
 
+def main_diff(args):
+    project = get_project()
+
+    if args.filter:
+        jobs = [project.open_job(id=job_id) for job_id in find_with_filter_or_none(args)]
+    elif args.job_id:
+        jobs = (_open_job_by_id(project, jid) for jid in args.job_id)
+    else:
+        jobs = project
```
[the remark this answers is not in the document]
atravitz: ah okay, I was following `main_statepoint()` as a reference. done!

atravitz:
suggestions from @vyasr have been addressed

signac/diff.py
```diff
@@ -0,0 +1,82 @@
...
+            sps[job] = set(_traverse_filter(job.sp()))
+
+        intersection = set.intersection(*sps.values())
+
+        diffs = {}
+        for job in jobs:
+            unique_sps = sps[job]-intersection
+            diffs[job.id] = _dotted_keys_to_nested_dicts(dict(unique_sps))
+
+        return(diffs)
```
vyasr: We don't parenthesize return values.

signac/diff.py
```diff
@@ -0,0 +1,82 @@
...
+
+def diff_jobs(*jobs):
+    r"""Find differences among a list of jobs' state points.
+
+    The resulting diff is a dictionary where the keys are job ids and the
+    values are each job's state point minus the intersection of all provided
+    jobs' state points. The comparison is performed over the combined set
+    of keys and values.
+
+    Example:
```
vyasr: Please add an extra new line here. I'm not 100% sure if it renders fine without that, so better safe than sorry.

bdice: This was indeed a bug -- I fixed it. You need newlines before _and_ after `.. code-block:: python`.

PR #250, opened 2019-12-02
Add id property for jobs

b-butler:
Add `id` property to `Job`

## Description

Adds `id` as a property of the `Job` class. I also added a test to ensure that it produces the correct string.

## Motivation and Context

Just follows the python trend of using properties instead of getters and setters.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -79,6 +79,11 @@ class Job(object):
         :rtype: str"""
         return self._id
 
+    @property
+    def id(self):
+        """The unique identifier for the job's statepoint."""
+        return self.get_id()
+
     def __hash__(self):
         return hash(os.path.realpath(self._wd))
 
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -6,6 +6,7 @@ import errno
 import logging
 import shutil
 from copy import deepcopy
+from deprecation import deprecated
 
 from ..core import json
 from ..core.attrdict import SyncedAttrDict
@@ -15,6 +16,7 @@ from .hashing import calc_id
 from .utility import _mkdir_p
 from .errors import DestinationExistsError, JobsCorruptedError
 from ..sync import sync_jobs
+from ..version import __version__
 
 
 logger = logging.getLogger(__name__)
@@ -72,6 +74,8 @@ class Job(object):
         # Prepare current working directory for context management
         self._cwd = list()
 
+    @deprecated(deprecated_in="1.3", removed_in="2.0", current_version=__version__,
+                details="Use job.id instead.")
     def get_id(self):
         """The unique identifier for the job's statepoint.
 
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -93,7 +93,7 @@ class Job(object):
 
     def __str__(self):
         "Returns the job's id."
-        return str(self.get_id())
+        return str(self.id)
 
     def __repr__(self):
         return "{}(project={}, statepoint={})".format(
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -82,7 +82,7 @@ class Job(object):
     @property
     def id(self):
         """The unique identifier for the job's statepoint."""
-        return self.get_id()
+        return self._id
 
     def __hash__(self):
         return hash(os.path.realpath(self._wd))
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -77,7 +77,7 @@ class Job(object):
     @deprecated(deprecated_in="1.3", removed_in="2.0", current_version=__version__,
                 details="Use job.id instead.")
     def get_id(self):
-        """The unique identifier for the job's statepoint.
+        """The job's statepoint's unique identifier.
 
         :return: The job id.
         :rtype: str"""
@@ -85,7 +85,10 @@ class Job(object):
 
     @property
     def id(self):
-        """The unique identifier for the job's statepoint."""
+        """The unique identifier for the job's statepoint.
+
+        :return: The job id.
+        :rtype: str"""
         return self._id
 
     def __hash__(self):
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -142,11 +142,10 @@ class Project(object):
         self._config = _ProjectConfig(config)
 
         # Ensure that the project id is configured.
-        id_ = self.get_id()
-        if id_ is None:
+        if self.id is None:
             raise LookupError(
-                "Unable to determine project id ."
-                "Are you sure '{}' is a signac project path?".format(
+                "Unable to determine project id. "
+                "Please verify that '{}' is a signac project path.".format(
                     os.path.abspath(self.config.get('project_dir', os.getcwd()))))
 
         # Prepare project document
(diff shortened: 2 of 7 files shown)


vyasr:
@b-butler I would just return `_id` instead.

bdice:
FYI, `id` is a function in the global namespace in Python. I think it's ok to have this here because it's under the class namespace but just want to avoid shadowing that by accident. The recommendation is to use `id_` instead of `id` where needed.

signac/contrib/job.py
```diff
@@ -79,6 +79,11 @@ def get_id(self):
         :rtype: str"""
         return self._id
 
+    @property
+    def id(self):
+        """The unique identifier for the job's statepoint."""
+        return self.get_id()
```
bdice:
```suggestion
        return self._id
```

csadorf:
So should we deprecate `get_id()` then? The reason that I used `get_id` over `id` was exactly what @bdice mentioned: `id` is a keyword in Python.

b-butler:
I think that as long as the keyword is in a class namespace, there isn't any confusion. It isn't a free `id`. There is always an object context associated with it.

vyasr:
Fwiw HOOMD (which has a substantially longer history and may have more empirical evidence to base this decision on) runs into the same problem with `hoomd.group.all` (see [this issue](https://github.com/glotzerlab/hoomd-blue/issues/488)).

bdice:
Keywords definitely [don't work](https://stackoverflow.com/questions/9746838/why-cant-attribute-names-be-python-keywords). But `id` isn't a keyword. I know HOOMD has had issues with `all` but I don't fully understand why. I commented on the other issue to ask.

Keywords:
```python
>>> import keyword
>>> keyword.kwlist
['False', 'None', 'True', 'and', 'as', 'assert', 'async', 'await', 'break', 'class', 'continue', 'def', 'del', 'elif', 'else', 'except', 'finally', 'for', 'from', 'global', 'if', 'import', 'in', 'is', 'lambda', 'nonlocal', 'not', 'or', 'pass', 'raise', 'return', 'try', 'while', 'with', 'yield']
```

bdice:
Related: https://stackoverflow.com/questions/6350847/is-id-a-keyword-in-python

b-butler:
@bdice good catch. Regardless having it as an attribute of an object shouldn't raise any confusion.

b-butler:
Should we also still raise a `LookupError` when accessing the `project.id` property? It might be confusing to raise an error when trying to access an attribute. I would think returning `None` might be more appropriate.

vyasr:
I believe the primary issue in hoomd is simply that groups are needed for many methods, so frequently users assign groups to local variables rather than calling e.g. `hoomd.group.all` everywhere. The documentation and examples implicitly endorse as a best practice based on the prevalence of the idiom `all = hoomd.group.all()` (which is obviously a problem); that's something that we could easily train users out of by removing it from examples, but it's even easier to change the group name. That particular problem is not likely to arise for `job.id`, but maybe there are others.

Regarding the `LookupError`, currently the constructor calls `get_id` as a test of whether the `Project` has been initialized with a valid root directory (one containing the project), so the error is always caught on construction. Since the project config will be immutable going forward, I think it should be safe to raise the error directly in the constructor and then remove that check entirely from `get_id` (which then becomes an alias for the property `project.id`). The only forms of corruption that would be possible are ones that we already don't protect against (like someone manually modifying or removing the config file after the `Project` class instance is created.

Definitely agree, deprecate `Project.get_id` as well.

b-butler:
@csadorf I didn't realize I hadn't set the flake8 pre-commit hook for signac core.

csadorf:
@bdice Please merge if you are satisfied.

signac/contrib/project.py
```diff
@@ -214,20 +219,23 @@ def workspace(self):
...
-            raise LookupError(
-                "Unable to determine project id ."
-                "Are you sure '{}' is a signac project path?".format(
-                    os.path.abspath(self.config.get('project_dir', os.getcwd()))))
+            return None
+
+    @property
+    def id(self):
+        """Get the project identifier."""
+        return self.get_id()
```
bdice: @b-butler Flip the implementation so that the deprecated method calls the preferred property.

signac/contrib/project.py
```diff
@@ -142,7 +142,12 @@ def __init__(self, config=None):
         self._config = _ProjectConfig(config)
 
         # Ensure that the project id is configured.
-        self.get_id()
+        id_ = self.get_id()
```
bdice: Use the property instead of the method. And remove this line, just use “if self.id is None” below since the variable isn’t re-used.

signac/contrib/project.py
```diff
@@ -142,7 +142,12 @@ def __init__(self, config=None):
...
+        if id_ is None:
+            raise LookupError(
+                "Unable to determine project id ."
```
```suggestion
                "Unable to determine project id."
```

bdice: Actually you need a space after the period so this concatenates properly with the following line.

signac/contrib/project.py
```diff
@@ -142,7 +142,12 @@ def __init__(self, config=None):
...
+                "Are you sure '{}' is a signac project path?".format(
```
bdice: I’m generally not a fan of questions in error messages. I would prefer an instruction like “Please verify that... is a signac project path.” I know we use the question style elsewhere, this is just personal preference.

b-butler:
The failed test is a performance test of `H5Store`. I am not sure how that happened.

bdice:
> The failed test is a performance test of `H5Store`. I am not sure how that happened.

Ignore that. We need to fix some of our tests.

csadorf:
> > The failed test is a performance test of `H5Store`. I am not sure how that happened.
> 
> Ignore that. We need to fix some of our tests.

I've updated master to skip those tests there. 

However, the benchmark test failed for py37, for whatever reason I can't trigger a re-run of the tests. @bdice can you do that?

bdice:
@b-butler After you check on the docs, this will be ready to merge -- go ahead and click the big green button when you're ready. :)

b-butler:
I believe the PR is ready.

csadorf:
@b-butler Thank you for merging. For next time, please make sure to have a final look at the PR checklist prior to merging. I noticed that for example, this patch is missing a changelog update (I missed that as well in my review), and this patch will also require an update to the framework documentation. So you should either just create the PR to update the docs or at least create a related issue.

Since you are a committer, you can just update the changelog without review.

csadorf:
> I believe the PR is ready. The docs look good, and I couldn't find any residual references to `get_id` in the docs.

I didn't see that comment, please disregard my comment about the docs then. I would just appreciate if you could update the changelog, that's all.

b-butler:
@csadorf good catch! I will definitely be more careful next time. The changelog is updated on master.

csadorf:
> The changelog is updated on master.

Did you push? :confused:

b-butler:
> > The changelog is updated on master.
> 
> Did you push? 😕

I thought I did; however, looking at my terminal history, I see it failed since it is a protected branch.

Issue #212, opened 2019-07-24
Proposal: Adopt the `pytest` library for testing


csadorf:
# Proposal

I propose to adopt the `pytest` library as the main testing framework as a replacement for the Python built-in `unittest` package.

# Pros
* Writing tests is easier.
* The pytest command line tool is easier to use and more powerful.
* The pytest output is much more descriptive.

# Cons
* We add a dependency to the testing workflow.

# Migration

Adopting `pytest` as the main testing framework would be very simple. In fact, it is already possible to execute all tests with `pytest` without the need to modify the signac source code or test code at all.
The following command executed within the signac package root directory will execute all tests:
```
~/signac $ pytest
```
These are the steps we would need to take:
1. [ ] Updates the dev-requirements to include pytest.
2. [ ] Update the circle-ci test commands.
3. [ ] Update the documentation on how to run tests.
4. [ ] (*optional*) Use [unittest2pytest](https://pypi.org/project/unittest2pytest/) to rewrite existing tests.

csadorf:
@glotzerlab/signac-maintainers @mikemhenry I'd like your opinion on this.

bdice:
Pytest is nice, but I'd consider the development time / maintenance costs of rewriting tests / potential benefits before diving in.

csadorf:
That's why I'm putting this up for debate. I don't think there are any major transition costs, since we don't have to rewrite any tests at all. I'd argue that maintenance will go down, since writing new tests will be simpler.

If at all, we just use automated tools to rewrite tests, but we probably need to make a few manual adjustments.

vyasr:
Bradley and I have discussed this possibility before for freud. Pytest is nice, but I haven't felt that the benefits merited the extra dependency. It does make writing new tests a bit shorter, but at this point writing tests with unittest just involves copying the boilerplate from another test, so it's basically zero effort. I don't think that rewriting our existing tests to use the pytest framework is a worthwhile investment of time, but I assume you would propose to just leave the existing tests in place (since pytest can process the vast majority of unittest features) if the autoconvert packages don't work well?

In general, I've largely heard good things about pytest. The main issue I've heard is that their development cycle is quick, so deprecations and removals of features were not infrequent over the past few years. I don't know if that's still the case though; their changelog looks reasonably stable at this point. I would prefer not to deal with deprecations in our testing framework, but that may no longer be an issue. 

[...]

Issue #101, opened 2018-10-22
signac shell, detect job and set job variable


csadorf:
**[Original report](https://bitbucket.org/glotzer/signac/issue/101) by Bradley Dice (Bitbucket: [bdice](https://bitbucket.org/bdice), GitHub: [bdice](https://github.com/bdice)).**

----------------------------------------

When I'm in a signac job workspace and call `signac shell`, I would really like it to set the `job` variable to be the job whose workspace I am in. I would be happy to implement this if you think it's a good idea, @csadorf.

csadorf:
This was implemented as part of PR #118 .

PR #253, opened 2019-12-06
Add schema version

### LINKED ISSUE #165: Use config's signac_version to track schema version for easier migration.
### Feature description

Currently it appears that we have an implied Project config value, `'signac_version': (0, 1, 0)`.
https://github.com/glotzerlab/signac/blob/ba14a1b60c7ba431d8c8e54d6909203248cccca9/signac/common/validate.py#L55

### Proposed solution

We should update the configuration to store the actual version of signac used to create or initialize the project. This will make migrations easier and make something of this currently unused config value. We explicitly have backward compatibility to 0.1.0 already so this won't break anything.

bdice:
## Description
Adds an explicit schema version for the signac data model.

## Motivation and Context
Resolves #165.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


bdice:
@csadorf I used the `Version` class and `parse_version` function, but I'm not sure if I like it. It adds "final" to the end of the version numbers when printed, like `1.0.0final`.

csadorf:
> @csadorf I used the `Version` class and `parse_version` function, but I'm not sure if I like it. It adds "final" to the end of the version numbers when printed, like `1.0.0final`.

Under what circumstances does this happen? For the warnings and error messages?

csadorf:
I just did a quick search, I'm not sure where we used that class anyways, maybe use `distutils.version.StrictVersion` instead?

bdice:
> > @csadorf I used the `Version` class and `parse_version` function, but I'm not sure if I like it. It adds "final" to the end of the version numbers when printed, like `1.0.0final`.
> 
> Under what circumstances does this happen? For the warnings and error messages?

Right, in the errors/warnings. To see it, just initialize a new project and then edit the config file to a higher major/minor version.

bdice:
@csadorf Yeah, distutils is another choice. I am unavailable for most of the weekend, so if you’re interested in trying it, go ahead and commit to this branch. We may want to check if that Version class is used elsewhere and deprecate it if distutils does the job.

csadorf:
I'll have a look at it.

csadorf:
I realized that there are few things we didn't fully consider. I'm almost done with implementing the missing pieces. Will push soon.

bdice:
@csadorf Reminder to push your changes when convenient.

signac/__main__.py
```diff
@@ -19,6 +19,8 @@
 import errno
 from pprint import pprint, pformat
 
+from packaging import version
```
bdice: https://stackoverflow.com/a/11887885
According to this StackOverflow post, `packaging` isn't in the standard library but comes as a dependency of `setuptools`. I just wanted to make a note of that.

signac/__main__.py
```diff
@@ -672,6 +674,36 @@ def main_update_cache(args):
...
+    from filelock import FileLock
+    project = get_project(_ignore_schema_version=True)
+
+    schema_version = version.parse(SCHEMA_VERSION)
+    config_schema_version = version.parse(project.config['schema_version'])
+
+    if config_schema_version > schema_version:
+        _print_err(
+            "The schema version of the project ({}) is newer than the schema "
+            "version supported by signac version {}: {}. Try updatign signac.".format(
```
bdice:
```suggestion
            "version supported by signac version {}: {}. Try updating signac.".format(
```

signac/__main__.py
```diff
@@ -672,6 +674,36 @@ def main_update_cache(args):
...
+        _print_err(
+            "The schema version of the project ({}) is up to date. "
+            "Nothing to do.".format(config_schema_version))
+    elif args.yes or query_yes_no(
+        "Do you want to migrate this project's schema version from '{}' to '{}'? "
+        "WARNING: THIS PROCESS IS IRREVERSIBLE!".format(
+            config_schema_version, schema_version), 'no'):
+        lock = FileLock(project.fn('.SIGNAC_PROJECT_MIGRATION_LOCK'))
+        with lock:
+            apply_migrations(project)
```
bdice: Should this be wrapped in a `try:` and remove the lock file on `finally:`?

signac/contrib/errors.py
```diff
@@ -31,3 +31,8 @@ def __init__(self, job_ids):
 class StatepointParsingError(Error, RuntimeError):
     "Indicates an error that occurred while tyring to identify a state point."
     pass
+
+
+class IncompatibleSchemaVersion(Error):
+    "The project's schema version is incompatible with this version."
```
bdice: If we use the word "version," we should always specify what kind of version (schema/signac package) to avoid confusion.
```suggestion
    "The project's schema version is incompatible with this version of signac."
```

signac/__main__.py
```diff
@@ -672,6 +674,36 @@ def main_update_cache(args):
...
+                config_schema_version, __version__, schema_version))
+    elif config_schema_version == schema_version:
+        _print_err(
+            "The schema version of the project ({}) is up to date. "
+            "Nothing to do.".format(config_schema_version))
+    elif args.yes or query_yes_no(
+        "Do you want to migrate this project's schema version from '{}' to '{}'? "
+        "WARNING: THIS PROCESS IS IRREVERSIBLE!".format(
+            config_schema_version, schema_version), 'no'):
+        lock = FileLock(project.fn('.SIGNAC_PROJECT_MIGRATION_LOCK'))
```
bdice: ```from .contrib.migrations import FN_MIGRATION_LOCKFILE```
```suggestion
        lock = FileLock(project.fn(FN_MIGRATION_LOCKFILE))
```

signac/contrib/migrations.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def _reload_project_config(project):
+    project_reloaded = project.get_project(
+        root=project._rd, search=False, _ignore_schema_version=True)
+    project._config = project_reloaded._config
+
+
+def _update_project_config(project, **kwargs):
+    "Update the project configuration, for internal use only."
```
bdice: Isn't "internal use only" implied by the fact that it's a private method?

signac/contrib/project.py
```diff
@@ -229,6 +234,32 @@ def get_id(self):
                 "Are you sure '{}' is a signac project path?".format(
                     os.path.abspath(self.config.get('project_dir', os.getcwd()))))
 
+    def _check_schema_compatibility(self):
+        """Checks whether this project's data schema is compatible with this version.
+
+        :raises RuntimeError:
+            In case that the schema version is incomaptible.
```
```suggestion
            If the schema version is incompatible.
```

signac/__main__.py
csadorf: Yes, so we should probably add `setuptools` to the dependencies, however, it is not clear to me if we need to require a specific version.

signac/__main__.py
csadorf: Yes, I think so.

signac/__main__.py
csadorf: I've applied this change.

signac/contrib/migrations.py
csadorf: Yes.

csadorf:
@bdice I plan on finalizing this before Friday. I need a free minute to consider how to potentially reduce the footprint of this patch (your concern).

I'll ping you again when I think it's ready and will ask you to informally review it again. I realize that this PR is kind of turned around now, sorry about that.

bdice:
> I realize that this PR is kind of turned around now, sorry about that.

Definitely not a problem - my goal was only to provide a draft implementation for further consideration. Thanks for taking it on.

signac/__main__.py
bdice: @csadorf I looked at the history of the setuptools git repo and changelog. `packaging` has been included as a vendored package (not a dependency) since setuptools v8.0 (Dec 13, 2014). I think that StackOverflow answer I linked above is out of date. I recommend that we add `packaging` as a dependency instead. It's a very common dependency so I expect it will be already met by most users' environments.

csadorf: But then in [v39](https://setuptools.readthedocs.io/en/latest/history.html#v39-0-0) it was changed to dependencies again?

Just tested, `packaging` is not installed by default in a "fresh" conda environment, but it's certainly very easy to install.

csadorf: I tested this and we need `packaging>=15.0`.

csadorf:
@bdice This should be ready now. Please have a look.

signac/common/validate.py
```diff
@@ -52,7 +41,7 @@ def get_validator():
 cfg = """
 project = string()
 workspace_dir = string(default='workspace')
-signac_version = version(default='0,1,0')
+schema_version = string(default='1')
```
bdice: Should this be a proper semantic version string?
```suggestion
schema_version = string(default='1.0.0')
```

csadorf:
> @csadorf Two change requests (I can't request changes on my own PR). I want to use proper semantic versions, e.g. `'1.0.0'` instead of `'1'`.

I opted to use a simpler system, because I don't think that semantic versioning actually makes sense in this context. What would it mean to have a minor or patch release of the scheme? Is there ever a scenario where you could operate on a minor or patch release without migrating?

I compare this to other API schemes, that usually use a simple sequential scheme. Many RESTful APIs use a single number to indicate which API version they are using. I don't think that semantic versioning makes sense in this context.

signac/common/validate.py
csadorf: See comment below.

bdice:
> > @csadorf Two change requests (I can't request changes on my own PR). I want to use proper semantic versions, e.g. `'1.0.0'` instead of `'1'`.
> 
> I opted to use a simpler system, because I don't think that semantic versioning actually makes sense in this context. What would it mean to have a minor or patch release of the scheme? Is there ever a scenario where you could operate on a minor or patch release without migrating?
> 
> I compare this to other API schemes, that usually use a simple sequential scheme. Many RESTful APIs use a single number to indicate which API version they are using. I don't think that semantic versioning makes sense in this context.

I agree that using an integer version (not a string version) is preferable, but then we don't need `packaging` at all, and there are other places where semantic versions are used but aren't needed.

csadorf:
> > > @csadorf Two change requests (I can't request changes on my own PR). I want to use proper semantic versions, e.g. `'1.0.0'` instead of `'1'`.
> > 
> > 
> > I opted to use a simpler system, because I don't think that semantic versioning actually makes sense in this context. What would it mean to have a minor or patch release of the scheme? Is there ever a scenario where you could operate on a minor or patch release without migrating?
> > I compare this to other API schemes, that usually use a simple sequential scheme. Many RESTful APIs use a single number to indicate which API version they are using. I don't think that semantic versioning makes sense in this context.
> 
> I agree that using an integer version (not a string version) is preferable, but then we don't need `packaging` at all, and there are other places where semantic versions are used but aren't needed.

Agreed, I'll make that change.

csadorf:
> > > > @csadorf Two change requests (I can't request changes on my own PR). I want to use proper semantic versions, e.g. `'1.0.0'` instead of `'1'`.
> > > 
> > > 
> > > I opted to use a simpler system, because I don't think that semantic versioning actually makes sense in this context. What would it mean to have a minor or patch release of the scheme? Is there ever a scenario where you could operate on a minor or patch release without migrating?
> > > I compare this to other API schemes, that usually use a simple sequential scheme. Many RESTful APIs use a single number to indicate which API version they are using. I don't think that semantic versioning makes sense in this context.
> > 
> > 
> > I agree that using an integer version (not a string version) is preferable, but then we don't need `packaging` at all, and there are other places where semantic versions are used but aren't needed.
> 
> Agreed, I'll make that change.

We determined that `packaging` was already inadvertently introduced as dependency as part of PR #253 so we'll leave this _as is_.

signac/version.py
```diff
@@ -5,4 +5,10 @@
 
 __version__ = "1.2.0"
 
-__all__ = ['__version__']
+SCHEMA_VERSION = "1.0.0"
```
bdice:
```suggestion
SCHEMA_VERSION = "1"
```

tests/test_project.py
```diff
@@ -1977,6 +1981,49 @@ def test_get_job_symlink_other_project(self):
         self.assertEqual(signac.get_job(symlink_path), job_a)
 
 
+class ProjectSchemaTest(BaseProjectTest):
+
+    def test_project_schema_versions(self):
+        impossibly_high_schema_version = '9999.0.0'
```
bdice:
```suggestion
        impossibly_high_schema_version = '9999'
```

Issue #265, opened 2019-12-30
H5Store locking issues


bdice:
### Description

I have been facing two issues that may or may not be related. Both issues are related to HDF5 file locking, a feature that was [implemented in HDF5 1.10.0 and changed in HDF5 1.10.1](https://support.hdfgroup.org/ftp/HDF5/releases/ReleaseFiles/hdf5-1.10.1-RELEASE.txt). This file locking feature is meant to help make SWMR (single writer, multiple reader) operations safer.

### To reproduce

The tests for storing pandas DataFrames in an H5Store fail. The file is already open from the `H5Store` instance, so when the `_h5get` function creates a [pandas `HDFStore` object](https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html#hdf5-pytables) (which uses pytables internally), the file fails to open because it's already locked by the `H5Store`.

### Error output

Example of a test failure message:
```python
======================================================================
ERROR: test_clear (__main__.H5StorePandasDataTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/bdice/miniconda3/envs/py38/lib/python3.8/site-packages/pandas/io/pytables.py", line 627, in open
    self._handle = tables.open_file(self._path, self._mode, **kwargs)
  File "/home/bdice/miniconda3/envs/py38/lib/python3.8/site-packages/tables/file.py", line 315, in open_file
    return File(filename, mode, title, root_uep, filters, **kwargs)
  File "/home/bdice/miniconda3/envs/py38/lib/python3.8/site-packages/tables/file.py", line 778, in __init__
    self._g_new(filename, mode, **params)
  File "tables/hdf5extension.pyx", line 492, in tables.hdf5extension.File._g_new
tables.exceptions.HDF5ExtError: HDF5 error back trace

  File "H5F.c", line 509, in H5Fopen
    unable to open file
  File "H5Fint.c", line 1400, in H5F__open
    unable to open file
  File "H5Fint.c", line 1615, in H5F_open
    unable to lock the file
  File "H5FD.c", line 1640, in H5FD_lock
    driver lock request failed
  File "H5FDsec2.c", line 941, in H5FD_sec2_lock
    unable to lock file, errno = 11, error message = 'Resource temporarily unavailable'

End of HDF5 error back trace

Unable to open/create file '/tmp/signac_test_h5store_952j661f/signac_test_h5store.h5'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #266, opened 2019-12-31
Make tests pass on Windows and test Windows on CI

### LINKED ISSUE #264: Run CI on Windows
### Feature description

I would like to add continuous integration support for signac on Windows. We provide packages for Windows, so I think it is reasonable for users to expect that most features are available and functional. However, we have no way to know if this is the case from our current CI on Linux. I have checked on my own system and there are several tests that fail on Windows.

### Proposed solution

I have started a branch that will add CircleCI executors for Windows. Currently, I only plan to test Python 3.8 to prevent our build matrix from exploding.

I am opening this issue to document the process and take note of related issues (#214, #107) that might have been caught by CI on Windows.

bdice:
## Description
We have been releasing packages as "noarch" but there were several incompatibilities with signac on Windows. To the best of my ability, I have corrected these issues so that all tests pass (or are skipped, where necessary).

The issues solved were present in tests and the core codebase, so it wasn't _just_ a testing problem.

The primary issues solved by this PR:
1. Inconsistencies with line separators (`\n` on POSIX, `\r\n` on Windows) and path separators (`/` on POSIX, `\` on Windows -- also backslashes need to be escaped in a variety of places using regular expressions, which made this more complicated).
2. The reverse of (1): in some places, we used `os.path.join` (which uses the platform's path separator) when we actually need forward slashes. This wouldn't have been caught on POSIX systems since the path separator is a forward slash, but caused failures on Windows.
3. Some errors are raised differently on Windows. For example, attempting to overwrite an existing file raises a `PermissionError` in several places, and its error code is different than on POSIX systems. We want to catch it appropriately so we can raise signac's `DestinationExistsError` in those cases.
4. Some potential file I/O issues were not caught by the existing tests. For example, the code snippet `job.data = {'a': 100}` should not be run while the H5Store is open or in a context manager. The setter for this method creates a temporary HDF5 file and then overwrites the `signac_data.h5` file, which may be problematic if the H5Store is currently open. This didn't cause issues on POSIX but fails on Windows, and in general it seems like a dangerous behavior to overwrite an open file.

## Motivation and Context
Resolves #264 and possibly also #265.

## Types of Changes

- [x] Documentation update
- [x] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,8 +2,7 @@ coverage==5.0.1
 numpy==1.18
 pandas==0.25.3; implementation_name!='cpython' --no-binary pandas
 pandas==0.25.3; implementation_name=='cpython' --no-binary :none:
-h5py==2.10; implementation_name!='cpython' --no-binary h5py
-h5py==2.10; implementation_name=='cpython' --no-binary :none:
+h5py==2.10; implementation_name=='cpython'
 tables==3.6.1
 click>=7.0
 ruamel.yaml>=0.15.89
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,9 +1,9 @@
-coverage==4
+coverage==5.0.1
 numpy==1.18
-pandas==0.25; implementation_name!='cpython' --no-binary pandas
-pandas==0.25; implementation_name=='cpython'
-h5py==2.9; implementation_name!='cpython' --no-binary h5py
-h5py==2.9; implementation_name=='cpython'
-tables==3.4.4
+pandas==0.25.3; implementation_name!='cpython' --no-binary pandas
+pandas==0.25.3; implementation_name=='cpython'
+h5py==2.10; implementation_name!='cpython' --no-binary h5py
+h5py==2.10; implementation_name=='cpython'
+tables==3.6.1
 click>=7.0
 ruamel.yaml>=0.15.89
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,9 +1,9 @@
 coverage==5.0.1
 numpy==1.18
 pandas==0.25.3; implementation_name!='cpython' --no-binary pandas
-pandas==0.25.3; implementation_name=='cpython'
+pandas==0.25.3; implementation_name=='cpython' --no-binary :none:
 h5py==2.10; implementation_name!='cpython' --no-binary h5py
-h5py==2.10; implementation_name=='cpython'
+h5py==2.10; implementation_name=='cpython' --no-binary :none:
 tables==3.6.1
 click>=7.0
 ruamel.yaml>=0.15.89
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,6 @@
 coverage==4
 mock==2
-numpy==1.15
+numpy==1.18
 pandas==0.25; implementation_name!='cpython' --no-binary pandas
 pandas==0.25; implementation_name=='cpython'
 h5py==2.9; implementation_name!='cpython' --no-binary h5py
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,4 @@
 coverage==4
-mock==2
 numpy==1.18
 pandas==0.25; implementation_name!='cpython' --no-binary pandas
 pandas==0.25; implementation_name=='cpython'
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -3,6 +3,6 @@ numpy==1.18
 pandas==0.25.3; implementation_name!='cpython' --no-binary pandas
 pandas==0.25.3; implementation_name=='cpython' --no-binary :none:
 h5py==2.10; implementation_name=='cpython'
-tables==3.6.1
+tables==3.6.1; implementation_name=='cpython'
 click>=7.0
 ruamel.yaml>=0.15.89
--- a/signac/__init__.py
+++ b/signac/__init__.py
@@ -10,7 +10,6 @@ This makes it easier to operate on large data spaces,
 streamlines post-processing and analysis and makes data
 collectively accessible."""
 
-from __future__ import absolute_import
 from . import contrib
 from . import db
 from . import cite
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -1,7 +1,6 @@
 # Copyright (c) 2018 The Regents of the University of Michigan
 # All rights reserved.
 # This software is licensed under the BSD 3-Clause License.
-from __future__ import print_function
 import os
 import stat
 import re
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -1109,7 +1109,7 @@ class Project(object):
 
         Alternatively the schema argument may be a string, that is converted into a schema function,
         for example: Providing ``foo/{foo:int}`` as schema argument means that all directories under
-        ``foo/`` will be imported and their names will be interpeted as the value for ``foo`` within
+        ``foo/`` will be imported and their names will be interpreted as the value for ``foo`` within
         the state point.
 
         .. tip::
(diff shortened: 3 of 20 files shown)


tests/test_shell.py
```diff
@@ -231,9 +236,9 @@ def test_shell_with_jobs_and_selection(self):
         for i in range(3):
             project.open_job(dict(a=i)).init()
         assert len(project)
+        python_command = 'python -m signac shell -f a.{}gt 0'.format('$' if WINDOWS else r'\$')
```
bdice: This is hacky but I couldn't figure out a better way that uses Python's standard library. On Windows, `$` won't work with the backslash to escape it. However, POSIX requires the backslash.

bdice:
There are a couple things I still need to fix that affect Python 3.5 and Python 3.6. The behavior of `re.escape` changed in Python 3.7, so that it no longer escapes `/` characters (because escaping forward slashes is not necessary). I'll fix that another day.

bdice:
@glotzerlab/signac-maintainers How much do we care about supporting H5Stores on PyPy? It has been a constant source of headaches for maintenance and I don't know of any PyPy users that we're actively supporting. I would like to drop tests requiring h5py from the PyPy CI if anyone else agrees.

signac/contrib/project.py
```diff
@@ -170,9 +169,9 @@ def __str__(self):
         return str(self.id)
 
     def __repr__(self):
-        return "{type}.get_project('{root}')".format(
+        return "{type}.get_project({root})".format(
                    type=self.__class__.__name__,
-                   root=self.root_directory())
+                   root=repr(self.root_directory()))
```
b-butler: Change isn't really related to windows compatibility, but makes sense.

tests/test_h5store.py
```diff
@@ -661,13 +664,16 @@ def test_single_writer_multiple_reader_same_process(self):
                         self.assertEqual(reader1['test'], True)
                         self.assertEqual(reader2['test'], True)
 
+    @unittest.skipIf(WINDOWS, 'This test fails for an unknown reason on Windows.')
```
b-butler: I am not too worried about this, but it would be nice to know why the tests fails.

tests/test_h5store.py
```diff
@@ -661,13 +664,16 @@ def test_single_writer_multiple_reader_same_process(self):
...
     def test_single_writer_multiple_reader_same_instance(self):
         from multiprocessing import Process
 
         def read():
             p = Process(target=_read_from_h5store, args=(self._fn_store,), kwargs=(dict(mode=None)))
```
b-butler: I am sure it already defaults to this, but since we are giving `kwargs` anyways `kwargs=(dict(mode='r'))` would be more explicit.

tests/test_h5store.py
```diff
@@ -703,9 +709,9 @@ def test_single_writer_multiple_reader_different_process_no_swmr(self):
     @unittest.skipUnless(python_implementation() == 'CPython', 'SWMR mode not available.')
     def test_single_writer_multiple_reader_different_process_swmr(self):
 
-        read_cmd = "python -c 'from signac.core.h5store import H5Store; "
```
b-butler: Is there a reason for these changes other than aesthetics? I don't see any difference in windows compatibility.

tests/test_shell.py
b-butler: It is for a test, so I wouldn't be too worried about it being 'hacky'. I don't think it is too bad though.

b-butler:
@bdice also Circleci still is trying to test the status of the linux tests with their old name.

signac/__init__.py
```diff
@@ -10,7 +10,6 @@
 streamlines post-processing and analysis and makes data
 collectively accessible."""
 
-from __future__ import absolute_import
```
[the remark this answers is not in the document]
bdice: Yes, I found one and then did a thorough search to remove them.

requirements-dev.txt
```diff
@@ -1,10 +1,9 @@
...
-mock==2
-numpy==1.15
-pandas==0.25; implementation_name!='cpython' --no-binary pandas
-pandas==0.25; implementation_name=='cpython'
-h5py==2.9; implementation_name!='cpython' --no-binary h5py
-h5py==2.9; implementation_name=='cpython'
-tables==3.4.4
+coverage==5.0.1
+numpy==1.18
+pandas==0.25.3; implementation_name!='cpython' --no-binary pandas
```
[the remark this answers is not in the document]
bdice: Good question -- I think we are trying to be safe. I haven't ever used Jython or IronPython or other Python implementations, but I think that binary CPython packages won't work with those either. I'm not sure.

signac/contrib/project.py
bdice: This is not obvious, but the change is needed for Windows compatibility The `repr` is needed to make the backslashes work correctly in the path string. Note that the quotes were also removed (since the `repr` includes quotes).

tests/test_h5store.py
bdice: I had to cut my losses, I couldn't figure it out after ~1 hour. I'm fine with skipping it.

tests/test_h5store.py
bdice: It's a compatibility fix for how Windows' `cmd` handles single quotes. It's different than bash. I just picked a syntax that I knew would work for both.

bdice:
> @bdice also Circleci still is trying to test the status of the linux tests with their old name.

That's normal - it'll be fixed when this PR is merged to master.

signac/contrib/project.py
bdice: For a few crucial classes with minimal "state" like this one, we try to adhere to the invariant `eval(repr(obj)) == obj`. That is, we want a `repr` that can be copy-pasted into a Python command and used.

tests/test_h5store.py
bdice: Good call. Fixed.

tests/test_h5store.py
b-butler: Makes sense. Ultimately unix-like systems dominate, so this shouldn't be too big an issue.

signac/contrib/project.py
b-butler: Ah okay, thanks for informing me.

requirements-dev.txt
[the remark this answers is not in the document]
b-butler: Nor am I, I just know that we explicitly test on `pypy`.

tests/test_h5store.py
b-butler: Ah, I wouldn't have thought of that.

b-butler:
Also, though I am sure this is apparent, my approval is conditional on allow us to skip the h5 testing on PyPy.

bdice:
I dropped HDF5 from the PyPy image but I'm still having trouble with another benchmark. It removed pandas 0.25.3 and re-installed 0.25.0 for the benchmark, which may have caused the error?

```python
  File "benchmark.py", line 45, in <module>
    import pandas as pd
  File "/root/repo/venv/site-packages/pandas/__init__.py", line 55, in <module>
    from pandas.core.api import (
  File "/root/repo/venv/site-packages/pandas/core/api.py", line 24, in <module>
    from pandas.core.groupby import Grouper, NamedAgg
  File "/root/repo/venv/site-packages/pandas/core/groupby/__init__.py", line 1, in <module>
    from pandas.core.groupby.generic import (  # noqa: F401
  File "/root/repo/venv/site-packages/pandas/core/groupby/generic.py", line 44, in <module>
    from pandas.core.frame import DataFrame
  File "/root/repo/venv/site-packages/pandas/core/frame.py", line 88, in <module>
    from pandas.core.generic import NDFrame, _shared_docs
  File "/root/repo/venv/site-packages/pandas/core/generic.py", line 71, in <module>
    from pandas.io.formats.format import DataFrameFormatter, format_percentiles
  File "/root/repo/venv/site-packages/pandas/io/formats/format.py", line 47, in <module>
    from pandas.io.common import _expand_user, _stringify_path
  File "/root/repo/venv/site-packages/pandas/io/common.py", line 9, in <module>
    import lzma
  File "/usr/local/lib-python/3/lzma.py", line 27, in <module>
    from _lzma import *
  File "/usr/local/lib_pypy/_lzma.py", line 15, in <module>
    from _lzma_cffi import ffi, lib as m
ImportError: /usr/local/lib_pypy/_lzma_cffi.pypy36-pp73-x86_64-linux-gnu.so: undefined symbol: lzma_index_stream_padding
```

Issue #249, opened 2019-11-26
Proposal: Unify dict classes and improve buffering and synchronization


vyasr:
Tl;dr: We need to improve synchronization and caching logic, and I think the first step is to combine the `_SyncedDict`, `SyncedAttrDict`, and `JSONDict` classes.

I apologize in advance for the lengthy nature of this issue. This issue will serve as a pseudo-signac Enhancement Proposal, I'll try and document very thoroughly and it can be a test case for the utility of such proposals :)

In view of our recent push for deprecations and our discussion of reorganizing namespaces and subpackages to prepare for signac 2.0, I'd like to also revisit discussion of the different dict classes. We have various open bugs and features (#234, #196, #239, #238, #198) that are related to improving our synchronization and caching processes. Our synchronization clearly has some holes in it, and in the process of making #239  @bdice has raised concerns about inconsistencies with respect to cache correctness and cache coherence, e.g. the fact that a job that exists and is cached will still exist in the cache after it is deleted (Bradley, feel free to add more information).

Fixing all of these is a complex problem, in part due to fragmentation in our implementation of various parts of the logic. I'd like to use this issue to broadly discuss the various problems that we need to fix, and we can spawn off related issues as needed once we have more of a plan of attack to address our problems. Planning this development more thoroughly is critical since the bugs that may arise touch on pretty much all critical code paths in signac. I think that a good first step is looking into simplifying the logic associated with our various dictionary classes. That change should make it easier to improve #198 since synchronization will be in one place. After that, I think it will be easier to consider the various levels of caching and properly define the invariants we want to preserve.

With respect to the various dictionary classes, I think we need to reassess and simplify our hierarchy:

-  The core class is the `_SyncedDict` class in `core/synceddict.py`, and I think it should exist in more or less its current form.
-  I understand the logic of separating out the `SyncedAttrDict` in `core/attrdict.py` since attribute-based access to a dictionary is technically a feature unrelated to synchronization. However, the class is very minimal, and I think that the benefits of maintaining this level of purity in distinction is outweighed by the increased difficulty users and newer developers have in finding code in the code base. I would like to merge these classes.
[...]

PR #269, opened 2020-01-10
Fix usage of _convert_to_dict to avoid modifying nesting SyncedDict objects when calling items or values

### LINKED ISSUE #234: Calling `_SyncedDict.values()` gives `dict` objects
### Description

Calling standard `MutableMapping` methods (`keys`, `values`, etc) on `_SyncedAttrDict` objects modifies the underlying dictionary in place. This behavior is unexpected, such an operation should only affect the returned value.

### To reproduce

```
>> import signac
>> project = signac.init_project('test')
>> job = project.open_job({'a': {'b': 'c'}}).init()
>> print(type(job.sp.a))
<class 'signac.core.attrdict.SyncedAttrDict'>
>> job.sp.values()
>> print(type(job.sp.a))
<class 'dict'>
```

There is an open question of whether `job.sp.values()` should return nested dictionaries as `SyncedDict` objects or normal `dict` objects. I prefer that `SyncedDict` objects be what it returns so that it's still going to sync, as does @bdice, but that is a separate question. Either way, calling values in the manner in this PR should not modify the object in place.

### System configuration

Linux-4.19.61-1-lts-x86_64-with-arch
3.7.4 (default, Oct  4 2019, 06:57:26)
[GCC 9.2.0]
1.2.0

vyasr:
## Description
A number of methods were calling `_convert_to_dict` on `self._data` rather than calling them on `self`. As a result, the member of the underlying dictionary was modified. Actually, we should be able to remove the `dict` branch of `_convert_to_dict` entirely, but that requires resolving #196 first (because that's the only way that branch is accessed). Further changes are probably worth postponing for when we address #249.

## Motivation and Context

Resolves #234.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -7,6 +7,14 @@ The **signac** package follows `semantic versioning <https://semver.org/>`_.
 Version 1
 =========
 
+next
+----
+
+Fixed
++++++
+
+ - Calling ``items`` or ``values`` on _SyncedDict objects does not mutate nested dictionaries (#234, #269).
+
 [1.3.0] -- 2019-12-20
 ---------------------
 
--- a/signac/core/synceddict.py
+++ b/signac/core/synceddict.py
@@ -265,11 +265,11 @@ class _SyncedDict(MutableMapping):
 
     def values(self):
         self._synced_load()
-        return self._convert_to_dict(self._data).values()
+        return self._convert_to_dict(self).values()
 
     def items(self):
         self._synced_load()
-        return self._convert_to_dict(self._data).items()
+        return self._convert_to_dict(self).items()
 
     def __repr__(self):
         return repr(self())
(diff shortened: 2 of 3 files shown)


bdice:
@vyasr Since the names of our CI checks changed with #266, you may need to merge `master` into this branch to make the required checks pass. I'm fine with you overriding the required checks if you want, since this PR is fairly small and doesn't touch anything altered in #266.

csadorf:
@vyasr Can you update the name of this PR to be a bit more descriptive?

vyasr:
Yes, sorry it was a single commit so it just autopopulated the title and I didn't change it.

PR #271, opened 2020-01-15
Fixed race condition when workspace is created. Resolves #267

### LINKED ISSUE #267: Fix race condition when workspace is created
### Description

In MPI parallel jobs, where every rank (or every `n` ranks) may initialize a statepoint on the fly, a race condition leads to the job crashing when no `workspace` directory is present. The explanation is that this directory is created on the fly simultaneously by all `job.init()`s of the same ensemble.

A detailed look identifies the `mkdir -p` in https://github.com/glotzerlab/signac/blob/2c0d9e4710bf9a758c53e74a9ad117acc20d43ae/signac/contrib/job.py#L337
as a culprit. The `mkdir` should not include the `-p` option, instead operate only on the target directory. Job workspaces should be created by default upon `signac init`.

Optionally, when they are deleted (for whatever reason), they could be recreated by an explicit `init` call, which would be properly guarded by MPI barriers. For now, the workaround is to create the workspace dir manually.

### To reproduce

```python
sp = {'just': 'a statepoint'}

# When initializing jobs on the fly in MPI setting:
job = project.open_job(sp)
device.comm.barrier_all() # HOOMD next
if device.comm.partition == 0 and device.comm.rank == 0:
    job.init() # this is the problematic line
device.comm.barrier_all()
```

This would be the suggested way to do it (pending changes to signac):
```python
sp = {'just': 'a statepoint'}

device.comm.barrier_all()
if device.comm.partition == 0 and device.comm.rank == 0:
    project.init()
device.comm.barrier_all()

# When initializing jobs on the fly in MPI setting:
job = project.open_job(sp)
device.comm.barrier_all()
if device.comm.partition == 0 and device.comm.rank == 0:
    job.init() # this is the problematic line
device.comm.barrier_all()
```
### Error output

If possible, copy any terminal outputs or attach screenshots that provide additional information on the problem.

```
Error occured while trying to create workspace directory for job '42e1468f85f2b689659f23bae00de995'.
State point manifest file of job '42e1468f85f2b689659f23bae00de995' appears to be corrupted.
  File "1brf_umbrella.py", line 125, in <module>
    job.init()
  File "/ccs/home/glaser/.conda/envs/myenv/lib/python3.7/site-packages/signac/contrib/job.py", line 427, in init
    self._init(force=force)
  File "/ccs/home/glaser/.conda/envs/myenv/lib/python3.7/site-packages/signac/contrib/job.py", line 354, in _init
    _mkdir_p(self._wd)
  File "/ccs/home/glaser/.conda/envs/myenv/lib/python3.7/site-packages/signac/contrib/utility.py", line 171, in _mkdir_p
    os.makedirs(path)
  File "/ccs/home/glaser/.conda/envs/myenv/lib/python3.7/os.py", line 221, in makedirs
    mkdir(name, mode)
...
```
### System configuration

 - Operating System [e.g. macOS]: Linux login2 4.14.0-115.8.1.el7a.ppc64le #1 SMP Thu May 9 14:45:13 UTC 2019 ppc64le ppc64le ppc64le GNU/Linux (really, any OS)
 - Version of Python [e.g. 3.7]: Python 3.7.4 (really, any python3 version)
 - Version of signac [e.g. 1.0]: signac 1.0.0 (but looking at the code the problem may also be present in newer versions)

vishav1771:
## Description

Project initialization add a empty Workspace directory if deleted by user.
## Motivation and Context

Solves issue #267 .

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -20,6 +20,11 @@ Fixed
 
  - Fixed issues on Windows with ``H5Store``, project import/export, and operations that move files (#264, #266).
 
+Changed
++++++++
+
+ - Workspace directory is created with ``Project`` instance (#231267 #271).
+
 
 [1.3.0] -- 2019-12-20
 ---------------------
@@ -36,7 +41,7 @@ Added
 Changed
 +++++++
 
- - Implemented ``Project.__contains__`` check in constant time (#231).
+ - Implemented ``Project.__contains__`` check in constant time (#267, #271).
 
 Fixed
 +++++
--- a/changelog.txt
+++ b/changelog.txt
@@ -14,12 +14,18 @@ Added
 +++++
 
  - Added Windows to platforms tested with continuous integration (#264, #266).
+ 
+Changed
++++++++
+
+ - Workspace directory is created when ``Project`` is initialized (#267, #271).
+
 
 Fixed
 +++++
 
  - Fixed issues on Windows with ``H5Store``, project import/export, and operations that move files (#264, #266).
- - Workspace directory is created when ``Project`` is initialized (#267, #271).
+
 
 
 [1.3.0] -- 2019-12-20
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,7 +23,7 @@ Fixed
 Changed
 +++++++
 
- - Workspace directory is created with ``Project`` instance (#231267 #271).
+ - Workspace directory is created with ``Project`` instance (#267, #271).
 
 
 [1.3.0] -- 2019-12-20
@@ -41,7 +41,7 @@ Added
 Changed
 +++++++
 
- - Implemented ``Project.__contains__`` check in constant time (#267, #271).
+ - Implemented ``Project.__contains__`` check in constant time (#231).
 
 Fixed
 +++++
--- a/changelog.txt
+++ b/changelog.txt
@@ -19,11 +19,7 @@ Fixed
 +++++
 
  - Fixed issues on Windows with ``H5Store``, project import/export, and operations that move files (#264, #266).
-
-Changed
-+++++++
-
- - Workspace directory is created with ``Project`` instance (#267, #271).
+ - Workspace directory is created when ``Project`` is initialized (#267, #271).
 
 
 [1.3.0] -- 2019-12-20
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -163,7 +163,7 @@ class Project(object):
             except OSError:
                 logger.error(
                     "Error occurred while trying to create "
-                    "workspace directory for project {}".format(self.id))
+                    "workspace directory for project {}.".format(self.id))
                 raise 
 
         # Internal caches
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -164,6 +164,7 @@ class Project(object):
                 logger.error(
                     "Error occurred while trying to create "
                     "workspace directory for project {}".format(self.id))
+                raise 
 
         # Internal caches
         self._index_cache = dict()
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -162,7 +162,7 @@ class Project(object):
                 _mkdir_p(self._wd)
             except OSError:
                 logger.error(
-                    "Error occured while trying to create "
+                    "Error occurred while trying to create "
                     "workspace directory for project {}".format(self.id))
 
         # Internal caches
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -164,7 +164,7 @@ class Project(object):
                 logger.error(
                     "Error occurred while trying to create "
                     "workspace directory for project {}.".format(self.id))
-                raise 
+                raise
 
         # Internal caches
         self._index_cache = dict()
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -155,14 +155,15 @@ class Project(object):
         # Prepare project document
         self._fn_doc = os.path.join(self._rd, self.FN_DOCUMENT)
         self._document = None
-        
-        # Prepare Workspace Directory 
+
+        # Prepare Workspace Directory
         if not os.path.isdir(self._wd):
             try:
                 _mkdir_p(self._wd)
             except OSError:
-                logger.error("Error occured while trying to create "
-                         "workspace directory for project {}".format(self.id))
+                logger.error(
+                    "Error occured while trying to create "
+                    "workspace directory for project {}".format(self.id))
 
         # Internal caches
         self._index_cache = dict()
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -155,6 +155,14 @@ class Project(object):
         # Prepare project document
         self._fn_doc = os.path.join(self._rd, self.FN_DOCUMENT)
         self._document = None
+        
+        # Prepare Workspace Directory 
+        if not os.path.isdir(self._wd):
+            try:
+                _mkdir_p(self._wd)
+            except OSError:
+                logger.error("Error occured while trying to create "
+                         "workspace directory for project {}".format(self.id))
 
         # Internal caches
         self._index_cache = dict()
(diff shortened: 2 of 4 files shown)


vishav1771:
@jglaser @csadorf Some tests are failing. Can you help me fix the issues?

bdice:
@vishav1771 Those tests are checking the assumptions of the old behavior. Since you changed the behavior, you will also need to update those tests. If you look at the test outputs, it will help you find the code to change. See the very bottom of this page for the error outputs: https://circleci.com/gh/glotzerlab/signac/4716

```python
======================================================================
ERROR: test_workspace_read_only_path (test_project.CachedProjectTest)
----------------------------------------------------------------------
  File "/home/circleci/repo/tests/test_project.py", line 237, in test_workspace_read_only_path
    with open(os.path.join(self.project.workspace()), 'w'):
IsADirectoryError: [Errno 21] Is a directory: '/tmp/signac_5zyno0sl/wd'

======================================================================
ERROR: test_workspace_read_only_path (test_project.ProjectTest)
----------------------------------------------------------------------
  File "/home/circleci/repo/tests/test_project.py", line 237, in test_workspace_read_only_path
    with open(os.path.join(self.project.workspace()), 'w'):
IsADirectoryError: [Errno 21] Is a directory: '/tmp/signac_qq_kwy8m/wd'

======================================================================
FAIL: test_no_workspace_warn_on_find (test_project.CachedProjectTest)
----------------------------------------------------------------------
  File "/home/circleci/repo/tests/test_project.py", line 217, in test_no_workspace_warn_on_find
    self.assertFalse(os.path.exists(self.project.workspace()))
AssertionError: True is not false

======================================================================
FAIL: test_no_workspace_warn_on_find (test_project.ProjectTest)
----------------------------------------------------------------------
  File "/home/circleci/repo/tests/test_project.py", line 217, in test_no_workspace_warn_on_find
    self.assertFalse(os.path.exists(self.project.workspace()))
AssertionError: True is not false
```

For the first two tests, you'll see a comment right above line 237 in `test_project.py`. It says:
```python
        # Create file where workspace would be, thus preventing the creation
        # of the workspace directory.
```
The purpose of `test_workspace_read_only_path` is to check the behavior if a workspace cannot be created. This test makes a **file** called `workspace`, which prevents the creation of the **directory** `workspace`. To fix this test, you'll need to add a line at the beginning of that test that removes the existing workspace directory before it makes the file.

The purpose of `test_no_workspace_warn_on_find` is to make sure that the user is warned if they attempt to search for jobs and no workspace directory is found. You should do the same thing here, just remove the workspace directory at the beginning of the test.

signac/contrib/project.py
```diff
@@ -156,6 +156,15 @@ def __init__(self, config=None, _ignore_schema_version=False):
         self._fn_doc = os.path.join(self._rd, self.FN_DOCUMENT)
         self._document = None
 
+        # Prepare Workspace Directory
+        if not os.path.isdir(self._wd):
+            try:
+                _mkdir_p(self._wd)
+            except OSError:
+                logger.error(
+                    "Error occured while trying to create "
```
bdice: Spelling fix: occured ➡️ occurred
```suggestion
                    "Error occurred while trying to create "
```

csadorf:
> @vishav1771 Those tests are checking the assumptions of the old behavior. Since you changed the behavior, you will also need to update those tests.

Just because it was not explicitly mentioned, I would like to point out that **changing existing unit tests is generally not appropriate** since this means that the patch introduces a change in behavior that is likely backwards-incompatible.

We should therefore carefully analyze in the review 
- what the exact nature of this change is,
- whether it is avoidable, and
- what possible implications it may have on users who implicitly or explicitly rely on the current behavior, and
- whether we should delay the release of this breaking change until the next major version increase.

@vishav1771 All of your changes are fine, I just want to make sure the reviewers are sensitive to this issue.

vishav1771:
@bdice I have done the changes you asked.

bdice:
@vishav1771 Thanks! I reviewed the PR checklist at the top and there are a couple of next steps:
- Add yourself to the list of contributors (edit `contributors.yaml`)
- Add a **new** test that checks if the workspace directory exists after the project is initialized, so that we guarantee this behavior in the future (you might call it `test_workspace_directory_exists` in `test_project.py`)
- Update the changelog with a short sentence documenting the change in behavior, and include the issue and PR numbers in parentheses (just follow the pattern). Put it under these headers:
```
Version 1
=========

next
----

Added
+++++
(... some stuff ...)

Changed
+++++++
- Your line goes here
```

tests/test_project.py
```diff
@@ -98,6 +98,11 @@ def test_workspace_directory_with_env_variable(self):
         self.project.config['workspace_dir'] = '${SIGNAC_ENV_DIR_TEST}'
         self.assertEqual(self._tmp_wd, self.project.workspace())
 
+    def test_workspace_directory_exists(self):
+        root = self._tmp_dir.name
```
bdice: There is already a project that is initialized by the setup for all the tests in this section. The class `ProjectTest` inherits from `BaseProjectTest`, which in turn inherits from `BaseJobTest` in `test_job.py`. The method `setUp` in that class initializes a project and makes it accessible via `self.project`. https://github.com/glotzerlab/signac/blob/dc870ac06159cae97e1475894913d1def567555e/tests/test_job.py#L71-L81

By using `self.project`, the contents of this test function can be simplified to just one line that uses the existing project:
```python
def test_workspace_directory_exists(self):
    self.assertTrue(os.path.exists(self.project.workspace()))
```

signac/contrib/project.py
```diff
@@ -156,6 +156,15 @@ def __init__(self, config=None, _ignore_schema_version=False):
...
         self._document = None
 
+        # Prepare Workspace Directory
+        if not os.path.isdir(self._wd):
+            try:
+                _mkdir_p(self._wd)
+            except OSError:
+                logger.error(
+                    "Error occurred while trying to create "
+                    "workspace directory for project {}".format(self.id))
```
bdice: After calling `logger.error(...)`, the error still needs to be raised because the project initialization has failed. Just add the statement `raise` after this error is logged.

Adding `raise` would also match the current behavior in the Job class when attempting to create the workspace directory.
https://github.com/glotzerlab/signac/blob/dc870ac06159cae97e1475894913d1def567555e/signac/contrib/job.py#L332-L341

bdice: Add a period at the end of the error message.
```suggestion
                    "workspace directory for project {}.".format(self.id))
```

changelog.txt
```diff
@@ -20,6 +20,11 @@ Fixed
 
  - Fixed issues on Windows with ``H5Store``, project import/export, and operations that move files (#264, #266).
 
+Changed
++++++++
+
+ - Workspace directory is created with ``Project`` instance (#267, #271).
```
bdice: I suggested a slight change in wording. Sometimes users will create signac projects from the command line, too. The command is `signac init PROJECT_NAME` so I think it's a bit clearer to say "initialize."
```suggestion
 - Workspace directory is created when ``Project`` is initialized (#267, #271).
```

bdice: Could you move this "Changed" section above "Fixed" so that it matches the order of the other changelog sections? Typically I think we follow the order "Added, Changed, Fixed, Deprecated, Removed." (There may be a couple exceptions to this.)

signac/contrib/project.py
```diff
@@ -156,6 +156,16 @@ def __init__(self, config=None, _ignore_schema_version=False):
...
 
+        # Prepare Workspace Directory
+        if not os.path.isdir(self._wd):
+            try:
+                _mkdir_p(self._wd)
+            except OSError:
+                logger.error(
+                    "Error occurred while trying to create "
+                    "workspace directory for project {}.".format(self.id))
+                raise
```
lyrivera: Wouldn't it be more informative if the error message was part printed with the `raise` statement?

bdice: I'd like to match the existing behavior for this kind of failure in the Job class. Logging an error and raising the original exception also allows users to catch that exception and handle it as they wish. Either choice would be okay in general, but let's keep it this way for consistency. Thanks for the attention to detail!

bdice:
Thanks again, @vishav1771! I'll merge this pull request. Resolves #267.

Issue #274, opened 2020-01-22
Cannot access project data through signac interface


klywang:
### Description

Accessing arrays from `project.data` using `signac` returns a `ValueError`. The same arrays may be accessed using`h5py` so I believe the data was written correctly. This error does not occur when accessing arrays using `job.data`or when accessing scalars and strings from `project.data`.

### To reproduce

In a project directory:
```bash
$ signac shell
>>> import numpy as np
>>> import h5py
>>> pr.data['a'] = np.zeros([10])
>>> with pr.data:
...     pr.data.a[:]
... 
```
This will return a `ValueError`.

Accessing the same arrays with `h5py` returns the correct values:
```bash
>>> f = h5py.File('signac_data.h5', mode='r')
>>> f['a']
<HDF5 dataset "a": shape (10,), type "<f8">
>>> f['a'][:]
array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])
>>> f.close()
```
### Error output

```bash
Traceback (most recent call last):
  File "<console>", line 2, in <module>
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/Users/kelwang/miniconda3/envs/ENV/lib/python3.8/site-packages/h5py/_hl/dataset.py", line 489, in __getitem__
    if is_empty_dataspace(self.id):
  File "/Users/kelwang/miniconda3/envs/ENV/lib/python3.8/site-packages/h5py/_hl/base.py", line 87, in is_empty_dataspace
    if obj.get_space().get_simple_extent_type() == h5s.NULL:
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5d.pyx", line 289, in h5py.h5d.DatasetID.get_space
ValueError: Not a dataset (not a dataset)
```

### System configuration

Please complete the following information:

 - Operating System [e.g. macOS]: macOS and Linux (Debian)
 - Version of Python [e.g. 3.7]: Python 3.8.1
 - Version of signac [e.g. 1.0]: 1.3.0

Or copy & paste the output of: `python -c 'import platform; print(platform.platform()); import sys; print(sys.version); import signac; print(signac.__version__)'`

bdice:
[...]

PR #275, opened 2020-01-23
Changed testing framework (Pytest)

vishav1771:
## Description
Changed testing framework from `Unittest` to `Pytest`

## Motivation and Context
Fixes #212 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change<sup>1</sup>

<sup>1</sup>The change breaks (or has the potential to break) existing functionality.

## Checklist:

- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
- [x] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [x] I have updated the API documentation as part of the package doc-strings.
- [x] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
- [x] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added all related issue and pull request numbers for future reference (if applicable). See example below.

Example for a changelog entry: `Fix issue with launching rockets to the moon (#101, #212).`

PR #278, opened 2020-01-27
Fix project data interface and add unittests for job and project data interfaces

klywang:
## Description
`project.data` used to return a new `H5StoreManager` each time it was called. This prevented users from reading arrays previously stored using the `project.data` interface.

Unit tests covered scalars and strings, but did not cover arrays or pandas dataframes.

## Motivation and Context
Resolves issue #274. Arrays stored using `project.data` cannot be accessed using the `project.data` interface.
These interfaces were not tested comprehensively in unit tests. 
This pull request is still in progress.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:


klywang:
In the `master` branch. The following code properly returns arrays:
``` bash
$ signac shell
>>> pr.data['a'] = np.zeros([10])
>>> with pr.data as store:
...     store.a[:]
... 
array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])
```
While 
``` bash
$ signac shell
>>> pr.data['a'] = np.zeros([10])
>>> with pr.data:
...     pr.data.a[:]
... 
```
returns an error. The current unit tests, which inherit from `test_h5store`, pass on the `master` branch since they test an equivelant of the first example.

If the goal is to add unit tests which can catch the bug reported in #274, would it be best to test both instances?

bdice:
@klywang We may choose to merge #275 first, which replaces unittest with pytest. I think it would be easier to update this PR to use pytest.

bdice:
@klywang Yes, we should definitely test both cases! Both are valid.

klywang:
@bdice Okay I can add tests for both instances after the testing framework has been updated.

bdice:
@jennyfothergill Good question! The test classes like `ProjectStoreTest` are actually getting all of their test logic from the parent classes. This is a case of [diamond inheritance](https://en.wikipedia.org/wiki/Multiple_inheritance#The_diamond_problem). I talked with @klywang about this in person, so I can re-cap it here. I apologize in advance for the fact that all the class names sound alike, which makes this hard to read.

The `H5StoreManager` itself was not the cause of the bug -- our _implementation_ of that store manager in the `Project` class had a bug (but this bug wasn't in the `Job` class, for some reason). To prevent this from happening again, we want to test `project.stores` and `project.data` directly (also need to test `job.stores` and `job.data` in a similar way before this PR is finalized). That is, we want tests that do everything that the `H5StoreManager` tests do, except with `project.stores` (which is itself an `H5StoreManager`). We solve that by creating a new class:

```python
class ProjectStoreTest(BaseProjectStoreTest, test_h5store.H5StoreTest):
    pass
```

**Everything in this class (the setup and the tests) is defined by inheritance.** It's not actually empty, as it appears. The `BaseProjectStoreTest` implements the core logic of the `BaseH5StoreTest` (e.g. a `setUp` function to create the project and methods for acquiring the data store, by returning `project.data`). The `BaseH5StoreTest` contains logic shared by all `H5Store` related tests. However, it doesn't actually have any tests. In `test_h5store.py`, you'll see that the tests are defined in `H5StoreTest`. However, we want to override the `BaseH5StoreTest` to use our temporary project's `project.data` instead of an `H5StoreManager` created in a temporary folder. So, we create `BaseProjectStoreTest` and re-define those methods. Then in the `ProjectStoreTest` class, the setup logic comes from `BaseProjectStoreTest` and the actual tests come from `test_h5store.H5StoreTest`.

In picture form:
image

FYI, this picture (and the class names in particular) may change when we shift to pytest.

jennyfothergill:
Oh! Wow, thank you so much for the great explanation! I think I understand.

bdice:
@klywang I updated this PR to match pytest. The tests appear to be running correctly with pytest. Next steps:
- [ ] pull the latest version of this branch
- [ ] add the second test instance you were mentioning above to `test_h5store.py` so it will be used by both sets of tests
- [ ] implement the same kind of inherited test classes for the `test_job.py`

If you want to run the tests locally, you'll need to install `pytest` and `pytest-subtests`. Then you can run `pytest test_project.py`, etc.

Use `pytest test_project.py -k "TestProjectStore" -v` to run all the store-related tests.

klywang:
While adding tests, I came across these questions:
1. Why is `H5StoreAlreadyOpen` ignored in `H5StoreManager`? This appears to allow users to embed context managers with different modes in their code, but in practice, only opens the file with the mode specifiecd in the initial context manager.
2. How similar should `H5Store` and `H5StoreManager` be? Tests for more advanced usages of `H5Store` fail when the same operations are performed indirectly through the project interface with `H5StoreManager`.

**Context for question 1 & 2**
As of now, the behavior of H5Stores and H5StoreManager behavior differs for several tests. Examples include tests for multiple instances of a opened file in read/write mode:
```python
with project.data.open(mode='w') as h5p:
    with project.data.open(mode='r') as h5p2:
       pass
```
Will return an `H5StoreAlreadyOpen` error while:
```python
with H5Store(filename, mode='w') as h5p:
    with H5Store(filename, mode='r') as h5p:
        pass
```
Will not return an error.

In `h5store.py`, the following lines are included in the `H5Store` context manager:
https://github.com/glotzerlab/signac/blob/569b8e5957ea07f00528ae3cfd1392f1c10d1043/signac/core/h5store.py#L317-L322

If I understand correctly, the context manager should ignore the arguments of the second context manager. To test this, I swapped line 118 and 119  in following test and ran it for `H5Store` rather than the project interface (since that one will fail regardless of order):
https://github.com/glotzerlab/signac/blob/569b8e5957ea07f00528ae3cfd1392f1c10d1043/tests/test_h5store.py#L117-L127

so that the order looked like this:
```python
with self.open_h5store(mode='r') as h5s_r:
    with self.open_h5store(mode='w') as h5s_w:
                assert 'foo' not in h5s_r
                assert 'foo' not in h5s_w

                h5s_w['foo'] = 'bar'
                assert 'foo' in h5s_r
                self.assertEqual(h5s_r['foo'], 'bar')
                assert 'foo' in h5s_w
                self.assertEqual(h5s_r['foo'], 'bar')
```
This test failed at `h5s_w['foo']  = 'bar'` because the file did not exist. This is likely due ot the fact that the `H5Store` was opened in "read" mode and "write" mode was never triggered.
Why are we allowing this sort of behavior and testing it in H5Store? I could attempt to replicate it in the project data interface (and `H5StoreManager`) but it seems like it may be better to allow the `H5StoreAlreadyOpen` error to occur, rather than passing it.

**Context for question 2**
Additionally, 4/5 of the tests in `TestH5StoreMultiProcessing` fail when tested on `project.data` but pass for `H5Store`. These tests seem to test if `__enter__` works correctly. It seems like calling the `H5StoreManager` class in a context manager does not trigger the same behavior as directly calling the `H5Store` class in a context manager. For example, this test assumes that calling the context manager in "append" mode (the default mode set for the context manager in `H5Store`) will create a file:
https://github.com/glotzerlab/signac/blob/569b8e5957ea07f00528ae3cfd1392f1c10d1043/tests/test_h5store.py#L685-L699

But the same test, where `open_h5store` is the yields the context manager for the project data or stores interface does not create a file. I would guess the difference in behavior is the same reason `H5Store` passes the `H5StoreAlreadyOpen` error but the `H5StoreManager` error.

klywang:
This should be for review now.

tests/test_project.py
```diff
@@ -2103,3 +2114,136 @@ def test_input_args(self):
...
+        with self.get_other_h5store().open(**kwargs) as h5s:
+            yield h5s
+
+
+class TestProjectStore(TestProjectStoreBase, test_h5store.TestH5Store):
+
+    """
+    This test opens multiple instances of H5Store, but
+    the project data interface opens one instance of H5Store.
+    Theses tests will (and should) fail using the project data interface.
```
zhou-pj:
```suggestion
    This test will (and should) fail using the project data interface.
```

tests/test_project.py
```diff
@@ -2103,3 +2114,136 @@ def test_input_args(self):
...
+    def test_assign_valid_types_within_same_file(self):
+        pass
+
+
+class TestProjectStoreOpen(TestProjectStoreBase, test_h5store.TestH5StoreOpen):
+
+    """
+    This test opens multiple instances of H5Store, but
+    the project data interface opens one instance of H5Store.
+    Theses tests will (and should) fail using the project data interface.
```
zhou-pj:
```suggestion
    This test will (and should) fail using the project data interface.
```

tests/test_project.py
```diff
@@ -2103,3 +2114,136 @@ def test_input_args(self):
...
                 # check that call does not fail:
                 tmp_project.detect_schema()
+
+
+class TestProjectStoreBase(test_h5store.TestH5StoreBase):
+
+    project_class = signac.Project
+
+    @pytest.fixture(autouse=True)
+    def setUp_base_h5Store(self, request):
```
zhou-pj: This is a minor questions and not directly introduced by this PR, but I noticed a few functions with the @pytest.fixture that are using a mixture of camelCase and underscore for function name. Is there a reason for that?

tests/test_project.py
```diff
@@ -2103,3 +2114,136 @@ def test_input_args(self):
...
+
+class TestProjectStore(TestProjectStoreBase, test_h5store.TestH5Store):
+
+    """
+    This test opens multiple instances of H5Store, but
+    the project data interface opens one instance of H5Store.
+    Theses tests will (and should) fail using the project data interface.
+    """
+    def test_assign_valid_types_within_same_file(self):
+        pass
```
zhou-pj: just to make sure I understand correctly, the comments above are for this test function and they are overwritten with `pass` because they are expected to fail and thus should be skipped?

klywang: Yes. The original tests opened multiple `H5Stores`. These tests use the `project.data` interface so we cannot test this.

Issue #262, opened 2019-12-19
Permissions on .signac_shell_history give cryptic error


bdice:
### Description

When launching `signac shell`, we got a cryptic error: `Error: [Errno 13] Permission denied`. This happened because I was collaborating with @tommy-waltmann. I ran `signac shell` and the file was created with permissions `600`, meaning that he couldn't read it. We should catch errors when the `.signac_shell_history` file is not readable, throw a warning, and continue to launch the shell with no history saved.

### System configuration

Please complete the following information:

 - Operating System [e.g. macOS]: Linux
 - Version of Python [e.g. 3.7]: Python 3.7
 - Version of signac [e.g. 1.0]: 1.2

bdice:
## Suggested changes to resolve this issue:

Edit this code to show a warning if the `.signac_shell_history` file is not readable/writable, and allow this exception to pass (don't raise). The warning should say something like ".signac_shell_history could not be read/written. The history of this shell will not be saved." Ideally, the warning text should be specific about the permissions it's missing (read or write). https://github.com/glotzerlab/signac/blob/6bf33e097e6545fe7e1d14fddd6e2f35cd23a4bd/signac/__main__.py#L1042-L1051

Test it by creating a `.signac_shell_history` file that is not user-readable/writeable. Make sure to test all combinations of file permissions (read/write, read-only, write-only, no read or write).

ac-optimus:
@bdice can I work on this issue if no one is working on it currently?
Thanks!

csadorf:
@ac-optimus Absolutely!

Btw, might be better to tag [signac-committers](https://github.com/orgs/glotzerlab/teams/signac-committers), anyone of them can assign you to an issue.

ac-optimus:
@csadorf sure,  Thanks !

vyasr:
This issue was resolved by #279.

PR #279, opened 2020-01-28
check conditions for .signac_shell_history permission.

ac-optimus:
## Motivation and Context

This will fix - #262

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:


signac/__main__.py
```diff
@@ -1042,13 +1042,20 @@ def jobs():
             if READLINE:
                 if 'PyPy' not in platform.python_implementation():
                     fn_hist = project.fn('.signac_shell_history')
-                    try:
```
csadorf: I'd like to keep the more *Pythonic* try-except approach to address this issue. The code is Python 3-only, so you can take full advantage of the [ `FileNotFoundError`](https://docs.python.org/3/library/exceptions.html#FileNotFoundError) and [`PermissionError`](https://docs.python.org/3/library/exceptions.html#PermissionError) exceptions.

ac-optimus: @csadorf ya sure, will update it accordingly!

csadorf: Great, please re-request review once you're done. Thx!

signac/__main__.py
```diff
@@ -1042,20 +1042,16 @@ def jobs():
...
-                    else:
+                    try:
                         readline.read_history_file(fn_hist)
                         readline.set_history_length(1000)
+                        with open(fn_hist, 'w') as f: pass
+                        atexit.register(readline.write_history_file, fn_hist)
+                    except FileNotFoundError:
                         atexit.register(readline.write_history_file, fn_hist)
+                    except PermissionError:
+                        print("Warning: .signac_shell_history does not have read/write permission.\
```
bdice: Try breaking the line like this. Python will combine the two strings into one message. Notice there is a space at the end of the first line.
```python
print("Warning: .signac_shell_history does not have read/write permission. "
      "The history of this shell will not be saved.")

Comment (ac-optimus) [reply signac/__main__.py]: @bdice done!

File: signac/__main__.py [hunk 3]
```diff
@@ -1045,10 +1045,13 @@ def jobs():
                     try:
                         readline.read_history_file(fn_hist)
                         readline.set_history_length(1000)
-                    except (IOError, OSError) as error:
-                        if error.errno != errno.ENOENT:
-                            raise
-                    atexit.register(readline.write_history_file, fn_hist)
+                        with open(fn_hist, 'w') as f: pass
+                        atexit.register(readline.write_history_file, fn_hist)
```
csadorf: Should we maybe just open it in append (`a`) mode so that in case that we fail to write the history from memory at program exit it is at least not lost?

signac/__main__.py
```diff
@@ -1045,10 +1045,13 @@ def jobs():
...
+                    except FileNotFoundError:
+                        atexit.register(readline.write_history_file, fn_hist)
+                    except PermissionError:
+                        print("Warning: .signac_shell_history does not have read/write permission. "
```
csadorf: I'd like to replace the literal `.signac_shell_history` with `os.path.relpath(fn_hist)` instead so that we have only one single source of truth with respect to this filename.

signac/__main__.py
ac-optimus: @csadorf yes seems like a good idea. done that!

signac/__main__.py
ac-optimus: @csadorf changed it.

signac/__main__.py
```diff
@@ -1045,10 +1045,13 @@ def jobs():
...
-                    except (IOError, OSError) as error:
-                        if error.errno != errno.ENOENT:
-                            raise
-                    atexit.register(readline.write_history_file, fn_hist)
+                        with open(fn_hist, 'w') as f: pass
+                        atexit.register(readline.write_history_file, fn_hist)
+                    except FileNotFoundError:
+                        atexit.register(readline.write_history_file, fn_hist)
+                    except PermissionError:
+                        print("Warning: " +os.path.relpath(fn_hist) +" does not have read/write permission. "
```
bdice: Try something like:
```suggestion
                        print(("Warning: {} does not have read/write permission. "
                               "The history of this shell will not be saved.").format(os.path.relpath(fn_hist)))
```
You'll probably need to edit the white space / etc. Just use `.format()`.

signac/__main__.py
bdice: @csadorf I'm not sure if I understand. How does using `a` fix the problem if we don't have write access? I thought the primary difference between `w` and `a` was whether it overwrote the existing file or appended to the end.

ac-optimus: @bdice actually the main reason for line-1048 is to check weather the file has a write permission. If it doesn't then we get an exception that is used to throw a warning.

bdice: @ac-optimus @csadorf I would prefer to use the approach of permissions-checking for this specific instance, instead of attempting to write to the file and catching an exception. I believe that opening the file (using `with open(...): pass`) requires more system calls and may have unintended side-effects such as memory allocation (for the file buffer) and updating the "modified" timestamp of the file with no actual modifications. My preferred solution would be to use `os.access` as before.

signac/__main__.py
```diff
@@ -1045,10 +1045,14 @@ def jobs():
...
-                            raise
-                    atexit.register(readline.write_history_file, fn_hist)
+                        with open(fn_hist, 'w') as f: pass
+                        atexit.register(readline.write_history_file, fn_hist)
+                    except FileNotFoundError:
+                        atexit.register(readline.write_history_file, fn_hist)
+                    except PermissionError:
+                        print(("Warning: {} does not have read/write permission. "
+                               "The history of this shell will not be saved.").format(os.path.relpath(fn_hist)))
+                              "The history of this shell will not be saved.")
```
bdice: Remove this line.

signac/__main__.py
csadorf: The [documentation for `os.access`](https://docs.python.org/3/library/os.html#os.access) explicitly recommends to use the [EAFP](https://docs.python.org/3/glossary.html#term-eafp) approach instead. The exact motivation for this (security issue) might not apply here, but your reasoning does not make any sense to me either, since this particular code segment is only executed once per opening of signac shell. The number of system calls should be completely irrelevant. I also don't see any issue with updating the modification time stamp.

IMO, the approach that we should take here, should be the one that is most robust across platforms. With that in mind, attempting to open the file for writing is the most surefire way to figure out whether we have the rights to do that.

The only reason I recommended 'a' over 'w' is that with the former we don't overwrite the file.

csadorf:
@ac-optimus I recommend that you execute the linter locally so that your code follows the code style guidelines.

ac-optimus:
> @ac-optimus I recommend that you execute the linter locally so that your code follows the code style guidelines.

okay sure.

signac/__main__.py
ac-optimus: @bdice @csadorf what to do now?

bdice:
@ac-optimus Your code still isn't passing the linter (flake8). Try installing the [flake8 hook as described here](https://docs.signac.io/projects/core/en/latest/support.html#set-up-a-development-environment), which prevents you from committing code that fails the linter. You can see the errors in the CI output.

```python
signac/__main__.py:1048:52: F841 local variable 'f' is assigned to but never used
                        with open(fn_hist, 'w') as f: pass
                                                   ^
signac/__main__.py:1048:53: E701 multiple statements on one line (colon)
                        with open(fn_hist, 'w') as f: pass
                                                    ^
signac/__main__.py:1054:101: E501 line too long (112 > 100 characters)
                               "The history of this shell will not be saved.").format(os.path.relpath(fn_hist)))
```

signac/__main__.py
bdice: @ac-optimus A few notes:

- Using mode `'a'` would definitely be correct, or else the file will be overwritten with empty contents when we check if it can be opened. I apologize, I should have understood that earlier.
- I agree with @csadorf in general, that EAFP is preferred. However, I think we're going about this in the wrong way. The file is written _at exit time_, and we're checking whether it's writable _during shell launch_. No matter whether you use `os.access` or attempt to write using `with open(...)`, the file permissions could be different at exit time. I believe the best approach would be to create a function `write_history_file` that attempts to save the file and prints a warning if the history could not be saved, and register `write_history_file` to be run at exit time.

My suggestion for this block would be the following:
```python
                    try:
                        readline.read_history_file(fn_hist)
                        readline.set_history_length(1000)
                    except FileNotFoundError:
                        pass
                    except PermissionError:
                        print("Warning: Shell history could not be read from "
                              "{}.".format(os.path.relpath(fn_hist)))
                    def write_history_file():
                        try:
                            readline.write_history_file(fn_hist)
                        except PermissionError:
                            print("Warning: Shell history could not be written to "
                                  "{}.".format(os.path.relpath(fn_hist)))
                    atexit.register(write_history_file)
```

ac-optimus:
> _Edit: This was in response to "what to do now?" -- sorry, I thought that was a general question. I will reply to the comment where you specifically wrote that question._
> 
> @ac-optimus Your code still isn't passing the linter (flake8). Try installing the [flake8 hook as described here](https://docs.signac.io/projects/core/en/latest/support.html#set-up-a-development-environment), which prevents you from committing code that fails the linter. You can see the errors in the CI output.
> 
> ```python
> signac/__main__.py:1048:52: F841 local variable 'f' is assigned to but never used
>                         with open(fn_hist, 'w') as f: pass
>                                                    ^
> signac/__main__.py:1048:53: E701 multiple statements on one line (colon)
>                         with open(fn_hist, 'w') as f: pass
>                                                     ^
> signac/__main__.py:1054:101: E501 line too long (112 > 100 characters)
>                                "The history of this shell will not be saved.").format(os.path.relpath(fn_hist)))
> ```
 @bdice @csadorf  corrected the code style(using linter). Need to learn a lot. Thanks!
Should I merge all the commits, seems like lot mess happened due to me?

bdice:
@ac-optimus Do you mean update this branch with `master`? You don't have to do that yet, we can wait until the end to merge in `master`. Take a look at the code block I suggested on that comment thread and let me know what you think. I think it resolves the questions that @csadorf and I both had about [EAFP](https://docs.python.org/3/glossary.html#term-eafp).

signac/__main__.py
ac-optimus: @bdice  seems a good fix but I want to point to something.  `readline.write_history_file(fn_hist)` does not throw `PermissionError` when we don't have write access but rather changes the permission of the file. And hence whatever be the        case( read/write, read-only, write-only, no read and write) the premission changes once `readline.write_history_file(fn_hist)`  executes. Because of this I added `atexit.register(readline.write_history_file, fn_hist)` when we have write permission and checked the write access using ```with``` statement.
```python
try:
    readline.read_history_file(fn_hist)
    readline.set_history_length(1000)
    with open(fn_hist, 'w'):
        pass
    atexit.register(readline.write_history_file, fn_hist)
except FileNotFoundError:
    atexit.register(readline.write_history_file, fn_hist)
except PermissionError:
    print(("Warning: {} does not have read/write permission. The history "
            "of this shell will not be saved.").format(os.path.relpath(fn_hist)))

```

bdice: @ac-optimus The original issue #262 was about reading the file. If writing the file succeeds and changes the permission of the file, that is fine. I believe that `readline.write_history_file` may be deleting the current file and re-creating it. In my understanding, deleting a file is allowed if the user has write permissions on its containing directory. You may need to see if a `PermissionError` is raise if the user does not have write permission on the containing directory.

ac-optimus: I thought we don't have to disturb the permission of the file. So in that case if we have read and write permission then we append to .signac_shell_history else we delete the current file and create new one to save the current shell history. Because``atexit.register(readline.write_history_file, fn_hist)`` will anyways create a new file if either read or write permission is not there.
And ``PermissionError`` will not be raised  as ``readline.write_history_file`` re-creates the file, when no write or read permission is there. So we might have to add ``with open(...)`` to check the file permission. And even warning might change. Please correct me if I am wrong.
The modified code, but here the write warning will appear after the shell exits --
```python
                     try:
                        readline.read_history_file(fn_hist)
                        readline.set_history_length(1000)
                    except FileNotFoundError:
                        pass
                    except PermissionError:
                        print("Warning: Shell history from "
                              "{}.".format(os.path.relpath(fn_hist)),
                              " would be lost. No read permission.")
                    def write_history_file():
                        try:
                            with open(fn_hist, 'a'):
                                pass
                        except PermissionError:
                            print("Warning: Shell history from "
                                  "{}.".format(os.path.relpath(fn_hist)),
                                  " would be lost. No write permission.")
                        readline.write_history_file(fn_hist)
                    atexit.register(write_history_file)
```
If we want the write warning to show up the moment the shell starts like the read warning then the code would be -
```python

                    try:
                        readline.read_history_file(fn_hist)
                        readline.set_history_length(1000)
                    except FileNotFoundError:
                        pass
                    except PermissionError:
                        print("Warning: Shell history from "
                              "{}.".format(os.path.relpath(fn_hist)),
                              " would be lost. No read premission.")
                    try:
                        with open(fn_hist, 'a'):
                             pass
                    except PermissionError:
                        print("Warning: Shell history from "
                              "{}.".format(os.path.relpath(fn_hist)),
                              " would be lost. No write permission.")
                    atexit.register(readline.write_history_file, fn_hist)

```

ac-optimus: or we can throw common error for both read and write permission like this-
```python
                    try:
                        readline.read_history_file(fn_hist)
                        readline.set_history_length(1000)
                        with open(fn_hist, "a"):
                            pass
                    except FileNotFoundError:
                        pass
                    except PermissionError:
                        print("Warning: Shell history from "
                              "{}".format(os.path.relpath(fn_hist)),
                              "would be lost. Permission error.")
                    atexit.register(readline.write_history_file, fn_hist)


```

csadorf: I'd just go with the last alternative, since I believe that it solves the original issue. @bdice Do you agree?

bdice: I recommend that we adopt the snippet I proposed above: https://github.com/glotzerlab/signac/pull/279#discussion_r373607211
@ac-optimus, can you apply that snippet and test it out manually? (no need to write a unit test)

csadorf: @bdice Sounds good to me!

ac-optimus: @bdice I already tried it manually and it was the reason why I proposed slight modification (1st code spinet above). ```readline.write_history_file(fn_hist)``` does not throw error when we do not have write permission and hence the exception is never called. And the 3rd one is more general. Please have a look and correct me if I am wrong.

csadorf:
@bdice It looks like this PR is waiting on your input. @ac-optimus Is that correct?

ac-optimus:
> @bdice It looks like this PR is waiting on your input. @ac-optimus Is that correct?

@csadorf yes I guess.

ac-optimus:
> After more testing, I found that the `write_history_file` function **will throw** if the file can actually not be overridden by the user. In conclusion, @ac-optimus please adopt @bdice 's proposed solution exactly. Thank you.

okays sure!

csadorf:
@ac-optimus It looks like there is some issue with our CI configuration. This sometimes happens when libraries that we depend on are updated in a non-backwards-compatible way. So please ignore the failing CI for now.

csadorf:
@ac-optimus Please merge master, this might also fix the CI issues.

ac-optimus:
> @ac-optimus Please merge master, this might also fix the CI issues.

@csadorf merged it to master of my fork.

csadorf:
> > @ac-optimus Please merge master, this might also fix the CI issues.
> 
> @csadorf merged it to master of my fork.

Looks like the master branch of your fork is out of date.

ac-optimus:
> > > @ac-optimus Please merge master, this might also fix the CI issues.
> > 
> > 
> > @csadorf merged it to master of my fork.
> 
> Looks like the master branch of your fork is out of date.

corrected that! the master is updated and merged with the permission_issue branch.

csadorf:
@atravitz I fixed the CI issues upstream and merged master here. Please merge this PR at your discretion.

PR #280, opened 2020-01-29
Detect Schema for different type of Project to Sync

### LINKED ISSUE #230: API command `sync` doesn't identify the `schema` correctly
### Description

calling `sync` to synchronize dst project with src, a warning arises, that the schemas differ, although they seem to be the same (statepoint files are attached
[signac_statepoints_src.txt](https://github.com/glotzerlab/signac/files/3641944/signac_statepoints_src.txt)
[signac_statepoints_dst.txt](https://github.com/glotzerlab/signac/files/3641949/signac_statepoints_dst.txt)

). 

### To reproduce

`sync src -v -u`

### Error output

```
WARNING: The detected schemas of the two projects differ! Use --force to ignore.
Keys found only in the source schema: dt, totalSteps, k_bending, period, phi, seed, a, reversalRate, totalMonomers, f_p
Keys found only in the destination schema: dt, totalSteps, k_bending, period, phi, seed, a, reversalRate, totalMonomers, f_p
```

### System configuration

 - Operating System [e.g. macOS]:
 - Version of Python [e.g. 3.7]:
 - Version of signac [e.g. 1.0]:

vishav1771:
## Description

Changed the `detect_schema` function to compare on the bases of data types instead of values.
## Motivation and Context
Fixes #230 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -537,8 +537,7 @@ def main_sync(args):
                 print(MSG_SYNC_STATS.format(stats=stats))
     except SchemaSyncConflict as error:
         _print_err(
-            "WARNING: Merging two projects with different schema requires the -m/--merge option. "
-            "Use --force to ignore all warnings.")
+            "Synchronizing two projects with different schema requires the -m/--merge option.")
         diff_src = error.schema_src.difference(error.schema_dst)
         diff_dst = error.schema_dst.difference(error.schema_src)
         only_in_dst = diff_dst.difference(diff_src)
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -1589,7 +1589,7 @@ job documents."
     parser_sync.add_argument(
         '--force',
         action='store_true',
-        help="Ignore  all warnings, just synchronize.")
+        help="Ignore all warnings, just synchronize.")
     parser_sync.add_argument(
         '--no-check-schema',
         action='store_true',
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -522,7 +522,7 @@ def main_sync(args):
             exclude=args.exclude,
             doc_sync=doc_sync,
             selection=selection,
-            check_schema=not args.no_check_schema,
+            check_schema=not (args.no_check_schema or args.force),
             dry_run=args.dry_run,
             parallel=args.parallel,
             deep=args.deep,
@@ -538,7 +538,8 @@ def main_sync(args):
     except SchemaSyncConflict as error:
         _print_err(
             "WARNING: The detected schemas of the two projects differ! "
-            "Use --no-scema-check to skip the schema check.")
+            "Use --no-scema-check to skip the schema check."
+            "Use --force to ignore all warnings.")
         diff_src = error.schema_src.difference(error.schema_dst)
         diff_dst = error.schema_dst.difference(error.schema_src)
         only_in_dst = diff_dst.difference(diff_src)
@@ -1585,6 +1586,10 @@ job documents."
         action='store_true',
         help="Allow the specification of a workspace (instead of a project) directory "
              "as the destination path.")
+    parser_sync.add_argument(
+        '-f', '--force',
+        action='store_true',
+        help="Ignore  all warnings, just synchronize.")
     parser_sync.add_argument(
         '--no-check-schema',
         action='store_true',
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -553,7 +553,9 @@ def main_sync(args):
                 "Keys found only in the destination schema: {}".format(', '.join(keys_formatted)))
         if diff_value:
             keys_formatted = ('.'.join(k) for k in diff_value)
-            _print_err("Keys having differnt values in source and destination: {}".format(', '.join(keys_formatted)))
+            _print_err(
+                "Keys having differnt values in source and destination: {}"
+                .format(', '.join(keys_formatted)))
     except DocumentSyncConflict as error:
         _print_err(MSG_SYNC_SPECIFY_KEY.format(keys=', '.join(error.keys)))
     except FileSyncConflict as error:
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -553,7 +553,7 @@ def main_sync(args):
         if diff_value:
             keys_formatted = ('.'.join(k) for k in diff_value)
             _print_err(
-                "Keys having differnt values in source and destination: {}"
+                "Keys having different values in source and destination: {}"
                 .format(', '.join(keys_formatted)))
     except DocumentSyncConflict as error:
         _print_err(MSG_SYNC_SPECIFY_KEY.format(keys=', '.join(error.keys)))
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -522,7 +522,7 @@ def main_sync(args):
             exclude=args.exclude,
             doc_sync=doc_sync,
             selection=selection,
-            check_schema=not (args.no_check_schema or args.force),
+            check_schema=not (args.merge or args.force),
             dry_run=args.dry_run,
             parallel=args.parallel,
             deep=args.deep,
@@ -537,8 +537,7 @@ def main_sync(args):
                 print(MSG_SYNC_STATS.format(stats=stats))
     except SchemaSyncConflict as error:
         _print_err(
-            "WARNING: The detected schemas of the two projects differ! "
-            "Use --no-scema-check to skip the schema check."
+            "WARNING: Merging two projects with different schema requires the -m/--merge option. "
             "Use --force to ignore all warnings.")
         diff_src = error.schema_src.difference(error.schema_dst)
         diff_dst = error.schema_dst.difference(error.schema_src)
@@ -1591,9 +1590,9 @@ job documents."
         action='store_true',
         help="Ignore all warnings, just synchronize.")
     parser_sync.add_argument(
-        '--no-check-schema',
+        '-m', '--merge',
         action='store_true',
-        help="Ignore statepoint schema check.")
+        help="Clone all the jobs that are not present in destination from source.")
     parser_sync.add_argument(
         '--parallel',
         type=int,
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -1587,7 +1587,7 @@ job documents."
         help="Allow the specification of a workspace (instead of a project) directory "
              "as the destination path.")
     parser_sync.add_argument(
-        '-f', '--force',
+        '--force',
         action='store_true',
         help="Ignore  all warnings, just synchronize.")
     parser_sync.add_argument(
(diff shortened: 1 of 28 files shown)


vishav1771:
Hi @csadorf  @vyasr, I have added tests to show the behavior of `sync` function with different types of projects. Is this behavior is fine or need to change anything?

vyasr:
@vishav1771 it looks like your tests are failing because the docstring you added to your test has an extra trailing quote (four instead of three). I'll look at the tests now.

tests/test_sync.py
```diff
@@ -587,27 +587,71 @@ def test_doc_sync(self):
             self.project_a.sync(self.project_b)
         self.project_a.sync(self.project_b, doc_sync=sync.DocSync.ByKey('b'))
 
-    def _setup_mixed(self):
+        
+
```
vyasr: Remove the extra lines here, or you'll have issues with our linter

vishav1771:
@vyasr I have not done any changes in the `detect_schema` yet. I just wanted to confirm the intended behavior of sync of function (as shown by the test ) is correct or not?

vyasr:
Yes, you've absolutely done the right thing. To make sure I understand these tests:

- `test_mixed_same_data_type` will fail with current signac, but we believe it should pass.
- `test_mixed_diff_data_type` will fail with current signac and we believe it should fail
- `test_mixed_multiple_data_type` will fail with current signac, and you think it should succeed in one direction but not the other (the direction where one project's key-values are a strict superset of the other).

Is that interpretation correct? One more I might add is where the types don't match but the values match exactly e.g. `float(1)` and `int(1)` as the values. It would also be good to test a case where one has a superset of the other's keys, i.e. both projects have `a={1, 2}`, but one project also has jobs with `b=1`.

Reading these tests makes me think that we are perhaps too stringent with our tests of invalid schema. The existence of the `check_schema` flag perhaps makes that acceptable, however it's not clear to me what a user of this would expect in all of these cases. For instance, I might naively only expect synchronization to fail if two jobs with the same statepoint fail to synchronize, i.e. I'm wondering if maybe the expected default for most users is simply `check_schema=False`. I would like to get some feedback from more users/developers on how they would expect synchronization to work. @csadorf would you be OK with polling a few others before making a decision on the *correct* behavior?

csadorf:
> Reading these tests makes me think that we are perhaps too stringent with our tests of invalid schema. 

I agree. The initial intention when I added the schema check was to prevent cases where one would apply a migration to one project data space (project A) and then sync it with a diverged copy (project B), e.g., one where a key was renamed, and now one would end up with a bunch of additional jobs that were originally supposed to be the same.

So here is my proposal, I think that we should keep the current behavior where sync really only syncs when the schema is **exactly** the same. However, we should clarify in the error message that this does not meany anything bad if you intend to sync two data space that are similar, but not identical in terms of schema. We could consider defusing the warning message and to use a different option to override just that check that sounds less menacing then `-f`/`--force`, something like `-m`/`--merge`  or so.

When I look at the [original issue](https://github.com/glotzerlab/signac/issues/230), the error message that we see is very confusing:
```
Keys found only in the source schema: dt, totalSteps, k_bending, period, phi, seed, a, reversalRate, totalMonomers, f_p
Keys found only in the destination schema: dt, totalSteps, k_bending, period, phi, seed, a, reversalRate, totalMonomers, f_p
```
That just doesn't make any sense since it claims that certain keys were only found in the source or destination schema, but they are actually identical.

This is what needs to be addressed.

vyasr:
I think that's a reasonable approach for now. If we go this route of not changing how schema checking works, long term I would advocate that for 2.0 we change the default behavior to `check_schema=False`. I think that is the behavior people are actually expecting. As long as the per-job synchronization is robust with respect to conflicts, the worst thing people can do is end up with duplicate jobs that can be easily removed or migrated correctly using the `signac shell` interface.

vishav1771:
> I think that's a reasonable approach for now. If we go this route of not changing how schema checking works, long term I would advocate that for 2.0 we change the default behavior to `check_schema=False`. 
I will do that. 
@vyasr Should I add this change to `changelog`  also?

vishav1771:
@vyasr  So, Only need to change the warning message and a different option to override for now?

vyasr:
Yes, I think that for now we would just want to add a `--no-check-schema` option and then change the warning.

signac/__main__.py
```diff
@@ -522,7 +522,7 @@ def _sig(st):
             exclude=args.exclude,
             doc_sync=doc_sync,
             selection=selection,
-            check_schema=not args.force,
+            check_schema=not args.no_check_schema,
```
csadorf: I am sorry, I should have been more clear about this. Please do not replace `-f`/`--force` option, but add an additional one.

Removing an existing option is not a backwards-compatible change and the `-f`/`--force` option is used to override other checks as well.

signac/__main__.py
```diff
@@ -538,7 +538,8 @@ def _sig(st):
     except SchemaSyncConflict as error:
         _print_err(
             "WARNING: The detected schemas of the two projects differ! "
-            "Use --no-scema-check to skip the schema check.")
+            "Use --no-scema-check to skip the schema check."
```
csadorf: There is a typo and you need to add a space after the dot ("."), to properly separate the sentences, however I'd like to actually propose something else:
- [ ] Rename the option to `-m`/`--merge`.
- [ ] Convert the warning to a simple more positive and affirmative message:

> Merging of two projects with different schemas requires the -m/--merge option.

My reasoning for the different option name is that it expresses the same behavior in a more positive way. "--no-schema-check" implies something negative to me, I am skipping a check and are potentially doing something bad, whereas `--merge` implies that I am doing something proper, just explicitly. Similar to how `cp` requires the `-r` option for recursive copying. It's not a bad thing, you just need to give explicit permission, because the consequences are bad if you did not really intend to do that.

See also: https://devguide.python.org/documenting/#affirmative-tone

@vyasr You cool with that?

vyasr: Yes, I think that is a good idea. I would also be clear about exactly what merging means, i.e. that statepoints existing in the source that are not in the destination will be copied over. This clarification avoids any confusion with respect to the merge strategy for individual jobs.

signac/__main__.py
```diff
@@ -537,17 +537,25 @@ def _sig(st):
                 print(MSG_SYNC_STATS.format(stats=stats))
     except SchemaSyncConflict as error:
         _print_err(
-            "WARNING: The detected schemas of the two projects differ! "
-            "Use --force to ignore.")
-        only_in_src = error.schema_src.difference(error.schema_dst)
+            "WARNING: Merging two projects with different schema requires the -m/--merge option. "
```
csadorf:
```suggestion
            "Synchronizing two projects with different schema requires the -m/--merge option. ")
```
This is not a warning, just a message.

signac/__main__.py
```diff
@@ -537,17 +537,25 @@ def _sig(st):
...
+            "Use --force to ignore all warnings.")
```
csadorf:
```suggestion

```

vishav1771:
> @vishav1771 One final question, have you verified that this will solve the specific issue for the data space posted in #230 ?

Yes, I have verified it with statepoints given in #230 and few other cases. However, while testing I came across a test case shown :
 ```
project1:
{
  job1:{
     'a' : 0,
     'b' : 0,
   }
  job2:{
     'a' : 1,
     'b' : 1,
   }
}

project2:
{
  job1:{
     'a' : 0,
     'b' : 1,
   }
  job2:{
     'a' : 1,
     'b' : 0,
   }
}
```
They will generate the same schema and will pass `check_schema`. I don't think that this is the intended behaviour. Can you please verify?

**Comment by csadorf** [comment]
> > @vishav1771 One final question, have you verified that this will solve the specific issue for the data space posted in #230 ?
> 
> Yes, I have verified it with statepoints given in #230 and few other cases. 

Great, thank you.

> They will generate the same schema and will pass `check_schema`. I don't think that this is the intended behaviour. Can you please verify?

While the two projects don't have the identical set of jobs, their implicit schema is indeed the same so IMO it is intended behavior that they would sync without problem. Otherwise we would need to redefine that merging means that one syncs two projects that have the exact same set of jobs. I don't think that's what we want. @vyasr Thoughts?

**Comment by vyasr** [comment]
> > > @vishav1771 One final question, have you verified that this will solve the specific issue for the data space posted in #230 ?
> > They will generate the same schema and will pass `check_schema`. I don't think that this is the intended behaviour. Can you please verify?
> 
> While the two projects don't have the identical set of jobs, their implicit schema is indeed the same so IMO it is intended behavior that they would sync without problem. Otherwise we would need to redefine that merging means that one syncs two projects that have the exact same set of jobs. I don't think that's what we want. @vyasr Thoughts?

I agree. I think this question is revisiting the motivation for the original requirement that schema had to be exactly identical, which was to avoid e.g. renaming or adding keys to schema and then trying to sync. If the keys and the types of the values corresponding to each key are the same between two schema, any jobs that exist in the source but not the destination must be truly distinct statepoints, so I think this behavior is the correct one.

**Comment by csadorf** [comment]
@vyasr Great, please merge at your discretion.

File: signac/__main__.py [hunk 6]
```diff
@@ -537,17 +537,24 @@ def _sig(st):
...
             _print_err("Keys found only in the source schema: {}".format(', '.join(keys_formatted)))
-        only_in_dst = error.schema_dst.difference(error.schema_src)
         if only_in_dst:
             keys_formatted = ('.'.join(k) for k in only_in_dst)
             _print_err(
                 "Keys found only in the destination schema: {}".format(', '.join(keys_formatted)))
+        if diff_value:
+            keys_formatted = ('.'.join(k) for k in diff_value)
+            _print_err(
+                "Keys having differnt values in source and destination: {}"
```
vyasr:
```suggestion
                "Keys having different values in source and destination: {}"
```

vyasr:
@vishav1771 I don't have push permissions to your repo. Can you commit my suggestion directly? I will merge after that.

Also, in general it is helpful for us if you complete the checklist when you make the pull request. That lets us track how far you've gotten.

signac/__main__.py
vishav1771: > @vishav1771 I don't have push permissions to your repo. Can you commit my suggestion directly? I will merge after that.

Sure.
> 
> Also, in general it is helpful for us if you complete the checklist when you make the pull request. That lets us track how far you've gotten.


Okay will do it  next time :)

vishav1771:
> @vishav1771 I don't have push permissions to your repo. Can you commit my suggestion directly? I will merge after that.

Sure
> 
> Also, in general it is helpful for us if you complete the checklist when you make the pull request. That lets us track how far you've gotten.

Sorry about that. I'll keep in mind for next time

csadorf:
@vyasr This is ready, isn't it?

vyasr:
Yes, thanks for the ping.

PR #282, opened 2020-01-30
Refactored internal use of get_statepoints

### LINKED ISSUE #227: Refactor internal use of deprecated `Project.get_statepoint`
The `Project.get_statepoint` function is deprecated but used internally. This should be refactored.

_Originally posted by @csadorf in https://github.com/glotzerlab/signac/pull/220_

vishav1771:
## Description

Refactored the internal use  of a deprecated method `get_statepoints` from `project.py` 
## Motivation and Context
Fixes #227 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:
- [x] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.

### DIFF
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -448,7 +448,7 @@ class Project(object):
                     id = matches[0]
                 elif len(matches) > 1:
                     raise LookupError(id)
-            return self.Job(project=self, statepoint=self.get_statepoint(id), _id=id)
+            return self.Job(project=self, statepoint=self.open_job(id).sp, _id=id)
 
     def _job_dirs(self):
         try:
@@ -778,7 +778,7 @@ class Project(object):
             tmp = dict()
         if statepoints is None:
             job_ids = self._job_dirs()
-            _cache = {_id: self.get_statepoint(_id) for _id in job_ids}
+            _cache = {_id: self.open_job(_id).sp for _id in job_ids}
         else:
             _cache = {calc_id(sp): sp for sp in statepoints}
 
@@ -1171,7 +1171,7 @@ class Project(object):
         logger.info("Checking workspace for corruption...")
         for job_id in self._find_job_ids():
             try:
-                sp = self.get_statepoint(job_id)
+                sp = self.open_job(job_id).sp
                 if calc_id(sp) != job_id:
                     corrupted.append(job_id)
                 else:
@@ -1220,7 +1220,7 @@ class Project(object):
         for job_id in job_ids:
             try:
                 # First, check if we can look up the state point.
-                sp = self.get_statepoint(job_id)
+                sp = self.open_job(job_id).sp
                 # Check if state point and id correspond.
                 correct_id = calc_id(sp)
                 if correct_id != job_id:
@@ -1272,7 +1272,7 @@ class Project(object):
         for _id in to_remove:
             del self._index_cache[_id]
         for _id in to_add:
-            self._index_cache[_id] = dict(statepoint=self.get_statepoint(_id), _id=_id)
+            self._index_cache[_id] = dict(statepoint=self.open_job(_id).sp, _id=_id)
         return self._index_cache.values()
 
     def _build_index(self, include_job_document=False):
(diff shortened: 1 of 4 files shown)


signac/contrib/project.py
```diff
@@ -778,7 +788,7 @@ def write_statepoints(self, statepoints=None, fn=None, indent=2):
             tmp = dict()
         if statepoints is None:
             job_ids = self._job_dirs()
-            _cache = {_id: self.get_statepoint(_id) for _id in job_ids}
+            _cache = {_id: self.open_job(id=_id).sp for _id in job_ids}
```
bdice: Let's change `sp` to `statepoint()`. `sp` and `statepoint` are aliased attributes (they're equivalent), but we need to use `dict`s in the cache instead of synced attribute dictionaries. If you call a synced dict (with `()`), it returns a standard `dict`.
```suggestion
            _cache = {_id: self.open_job(id=_id).statepoint() for _id in job_ids}
```

signac/contrib/project.py
```diff
@@ -1171,7 +1181,7 @@ def check(self):
         logger.info("Checking workspace for corruption...")
         for job_id in self._find_job_ids():
             try:
-                sp = self.get_statepoint(job_id)
+                sp = self.open_job(id=job_id).sp
```
bdice: I think this case is more complicated. If I remember correctly, `open_job(id)` cannot be run successfully if the job is corrupted, which defeats the purpose of this line. I am not sure what to do without further consideration.

signac/contrib/project.py
```diff
@@ -1220,7 +1230,7 @@ def repair(self, fn_statepoints=None, index=None, job_ids=None):
         for job_id in job_ids:
             try:
                 # First, check if we can look up the state point.
-                sp = self.get_statepoint(job_id)
+                sp = self.open_job(id=job_id).sp
```
bdice: Similar case to above -- `open_job` validates the state point. May need to rethink this case.

signac/contrib/project.py
```diff
@@ -1272,7 +1282,7 @@ def _sp_index(self):
         for _id in to_remove:
             del self._index_cache[_id]
         for _id in to_add:
-            self._index_cache[_id] = dict(statepoint=self.get_statepoint(_id), _id=_id)
+            self._index_cache[_id] = dict(statepoint=self.open_job(id=_id).sp, _id=_id)
```
```suggestion
            self._index_cache[_id] = dict(statepoint=self.open_job(id=_id).statepoint(), _id=_id)
```

signac/contrib/project.py
```diff
@@ -1281,7 +1291,7 @@ def _build_index(self, include_job_document=False):
         """
         wd = self.workspace() if self.Job is Job else None
         for _id in self._find_job_ids():
-            doc = dict(_id=_id, statepoint=self.get_statepoint(_id))
+            doc = dict(_id=_id, statepoint=self.open_job(id=_id).sp)
```
```suggestion
            doc = dict(_id=_id, statepoint=self.open_job(id=_id).statepoint())
```

signac/contrib/project.py
vishav1771: Can we use `read_statepoints()[job_id]` here?

bdice:
@vishav1771 I like what @csadorf suggested. Replacing internal uses of `get_statepoint` with `_get_statepoint` is a clean way to solve this for now. However, we need to restore the public API of `get_statepoint`. We adhere to [Semantic Versioning](https://semver.org/), which means we will not remove features from the public API without incrementing the major version number (e.g. to signac 2.0).

Here are the steps to address this:
- [ ] Add back the line in the documentation referring to this method.
- [ ] Add back the `@deprecated` decorator, function `get_statepoint`, and docstring, and replaces its contents with `return _get_statepoint(arguments)`

vishav1771:
@bdice  I have done the changes you asked. Kindly check if something needs to be changed?

bdice:
@csadorf You can merge this after you review. Thanks!

PR #287, opened 2020-02-11
Remove support for rapidjson

### LINKED ISSUE #285: Remove support for rapidjson
### Feature description

The rapidjson library is optionally supported by signac. Currently this support is enabled by default if the rapidjson library is installed. From the discussion in #171, we would like to deprecate support for rapidjson. The rapidjson support is implemented in this file:
https://github.com/glotzerlab/signac/blob/master/signac/core/json.py

### Proposed solution

There should be no API breaking changes as a result of removing rapidjson. I'm not sure if this change requires a proper deprecation cycle -- it may be sufficiently minor that we can just remove rapidjson support.

Some options:
1. Remove rapidjson support, just a clean break. **This is my preference.**
2. Throw a deprecation warning if rapidjson is automatically loaded. Since rapidjson is currently loaded by default, I dislike this approach because it would be noisy. The only easy way to hide the warning would be to ask users to uninstall rapidjson. That seems silly.
3. Make rapidjson opt-in (instead of automatic) with a configuration variable. However, that's a nontrivial amount of work and conflicts with the discussions in #171 about why we want to remove alternative JSON in the first place.

I would like a second opinion from @glotzerlab/signac-committers.

### Additional context

Minor refactoring to allow potential support for other JSON backends in the future may be worthwhile (see [comment from @csadorf](https://github.com/glotzerlab/signac/issues/171#issuecomment-583296803)) but a simple solution is fine for now.

ac-optimus:
This resolves #285.
## Description

Removed support for rapidjson.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


signac/core/json.py
```diff
@@ -12,26 +15,7 @@
...
-        encode = Encoder.__call__
-
-    logger.debug(msg.format('rapidjson'))
-except ImportError:
-    import json
-    from json import JSONEncoder
-    from json.decoder import JSONDecodeError
-
-    logger.debug(msg.format('json'))
+logger.debug(msg.format('json'))
```
bdice: I think this can be removed.

signac/core/json.py
```diff
@@ -2,6 +2,9 @@
 # All rights reserved.
 # This software is licensed under the BSD 3-Clause License.
 import logging
+import json
```
bdice: See if we can use this suggestion and remove the functions defined below.
```suggestion
from json import load, loads
```

signac/core/json.py
```diff
@@ -12,26 +15,7 @@
...
-except ImportError:
-    import json
-    from json import JSONEncoder
-    from json.decoder import JSONDecodeError
-
-    logger.debug(msg.format('json'))
+logger.debug(msg.format('json'))
 
 
 _has_default = hasattr(JSONEncoder, 'default')
```
bdice: This seems like a compatibility feature related to rapidjson, I am not sure if it is still needed with the plain `json` module. See if you can verify this in the documentation for JSONEncoder.

ac-optimus:
hi, below are the changes as per your suggestion. With plain json we don't need [JSONEncode](https://pythonexamples.org/convert-python-class-object-to-json/). Please check if this works. I have renamed the class to `CustomJSONdump()`. But I am confused about the exception?
```python
# Copyright (c) 2018 The Regents of the University of Michigan
# All rights reserved.
# This software is licensed under the BSD 3-Clause License.
import json
from json import load, loads

try:
    import numpy
    NUMPY = True
except ImportError:
    NUMPY = False


class CustomJSONdump():
    """Attempt to JSON-encode objects beyond the default supported types.

    This encoder will attempt to obtain a JSON-serializable representation of
    an object that is otherwise not serializable, by calling the object's
    `_as_dict()` method.
    """
    def default(self, o):
        if NUMPY:
            if isinstance(o, numpy.number):
                return o.item()
            elif isinstance(o, numpy.ndarray):
                return o.tolist()
        try:
            return o._as_dict()
        except AttributeError:
            return super(CustomJSONdump, self).default(o)


def dumps(o, sort_keys=False, indent=None):
    return json.dumps(CustomJSONdump(sort_keys=sort_keys, indent=indent))


__all__ = ['loads', 'load', 'dumps']
```

ac-optimus:
> I have a couple suggestions for further simplification. Also please add yourself to the contributors list, @ac-optimus! Good work, I think this is pretty close to complete.

ya sure!😃

bdice:
@ac-optimus You need to keep the custom encoder class like it was, except for the `_has_default` bit. We must extend the encoder class from the json module to support serialization of NumPy data types. Please change that back. I think the other changes may be fine, but it is hard to tell without a proper diff.

Please make separate commits to address each change that I recommended. It is difficult to review changes in long snippets since I cannot tell what changed. Also, please commit all the changes that you can (and try to make the tests pass) _before_ re-requesting review next time.

ac-optimus:
> @ac-optimus You need to keep the custom encoder class like it was, except for the `_has_default` bit. We must extend the encoder class from the json module to support serialization of NumPy data types. Please change that back. I think the other changes may be fine, but it is hard to tell without a proper diff.
> 
> Please make separate commits to address each change that I recommended. It is difficult to review changes in long snippets since I cannot tell what changed. Also, please commit all the changes that you can (and try to make the tests pass) _before_ re-requesting review next time.

okaie sure!

signac/core/json.py
ac-optimus: I guess we can not use `from json import load, loads` I am getting the following error for the same  
Screenshot from 2020-02-12 02-36-45


Also while testing manually I encountered the following issue when I call `json.loads('["foo", {"bar":["baz", null, 1.0, 2]}]')`(In the previous code where we have load and loads function defined). It is getting stuck in recursion. Can you suggest if there is some flaw in my approach?
Screenshot from 2020-02-12 02-34-34

bdice: @ac-optimus You're seeing this issue because you're running the Python interpreter in the directory containing `json.py`. When you're importing `json`, it's not importing from the standard library. It's importing from `json.py`.

If you run `cd $HOME; python -c "from json import load, loads"` you'll notice it works fine. This isn't an issue you'll see when importing `signac.core.json` in the package.

ac-optimus: @bdice  the first error comes when I add `from json import load, loads` in json.py while removing load and loads function as was suggested by you. I completely understand the point related to `cd $HOME; python -c "from json import load, loads"`. I was mentioning that while replacing `load` and `loads` methods from `json.py` with `from json import load, loads` I was welcomed by the error as in first snippet(If I am not committing some silly mistake then this can not be done). Also could you explain why am I getting this recursion when I just try to call load and loads methds from package's `json.py` when no change is made(In the both the snippets I am trying to import `json` from the package and not the standard library one?

bdice:
Hi @ac-optimus, I have pushed some changes to apply the recommendations I made. Please review these changes and let me know if you have any questions or comments.

signac/core/json.py
bdice: @ac-optimus I still believe the issue is caused by trying to run your Python interpreter in that directory. With the changes I just pushed, I get this output:

```python
~/code/signac/signac/core $ python
Python 3.7.6 | packaged by conda-forge | (default, Dec 27 2019, 00:09:34)
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
  File "/home/bdice/code/signac/signac/core/json.py", line 5, in <module>
    from json import load, loads, JSONEncoder
ImportError: cannot import name 'load' from 'json' (/home/bdice/code/signac/signac/core/json.py)
>>>
```
This is expected, since I'm using `core/json.py` as a standalone module from the `core/` directory instead of as it is intended in the package.

ac-optimus:
> Hi @ac-optimus, I have pushed some changes to apply the recommendations I made. Please review these changes and let me know if you have any questions or comments. If you don't have any changes to make, I will approve this PR and then @atravitz can review.

@bdice ya sure, please go ahead, sorry for creating confusion. Thanks!

Issue #291, opened 2020-02-14
Reduce test noise due to deprecation warnings


csadorf:
### Description

Switching to pytest as the primary testing framework has vastly increased the noise for test outputs due to deprecation warnings. The noise should be reduced analagous to how this was handled for signac-flow, here: https://github.com/glotzerlab/signac-flow/pull/246

### To reproduce

Execute the test framework with `pytest` within the repository root directory.

ac-optimus:
hi @csadorf can i work on this issue?

csadorf:
@ac-optimus Absolutely! Please let me know if you have any questions.

ac-optimus:
> @ac-optimus Absolutely! Please let me know if you have any questions.

yes sure!

vyasr:
Resolved by #292

PR #292, opened 2020-02-17
Reduce test noise due to deprecation warnings

ac-optimus:
fixes - #291
## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/contributors.yaml
+++ b/contributors.yaml
@@ -83,5 +83,5 @@ contributors:
   -
     family-names: Chandra
     given-names: Abhavya
-    affiliation: "India Institute of Technology, Gandhinagar"    
+    affiliation: "Indian Institute of Technology, Gandhinagar"    
 ...
--- a/contributors.yaml
+++ b/contributors.yaml
@@ -83,5 +83,5 @@ contributors:
   -
     family-names: Chandra
     given-names: Abhavya
-    affiliation: "Indian Institute of Technology, Gandhinagar"    
+    affiliation: "Indian Institute of Technology, Gandhinagar"
 ...
--- a/contributors.yaml
+++ b/contributors.yaml
@@ -80,4 +80,8 @@ contributors:
     family-names: Sharma
     given-names: Vishav
     affiliation: "National Institute of Technology, Hamirpur"
+  -
+    family-names: Chandra
+    given-names: Abhavya
+    affiliation: "India Institute of Technology, Gandhinagar"    
 ...
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,3 +7,7 @@ python-tag = py3
 [flake8]
 max-line-length=100
 exclude = configobj,passlib,cite.py,conf.py
+
+[tool:pytest]
+filterwarnings =
+  ignore:. is deprecated.*:DeprecationWarning
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,7 +7,3 @@ python-tag = py3
 [flake8]
 max-line-length=100
 exclude = configobj,passlib,cite.py,conf.py
-
-[tool:pytest]
-filterwarnings =
-  ignore:. is deprecated.*:DeprecationWarning
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,3 +7,7 @@ python-tag = py3
 [flake8]
 max-line-length=100
 exclude = configobj,passlib,cite.py,conf.py
+
+[tool:pytest]
+filterwarnings =
+    ignore: .*[The indexing module | get_statepoint] is deprecated.*: DeprecationWarning
--- a/tests/test_buffered_mode.py
+++ b/tests/test_buffered_mode.py
@@ -38,6 +38,13 @@ with TemporaryDirectory() as tmp_dir:
         os.chmod(path, mode)
 
 
+def deprecated_call(fun):
+    def new_fun(*argv, **kwargs):
+        with pytest.deprecated_call():
+            return fun(*argv, **kwargs)
+    return new_fun
+
+
 @pytest.mark.skipif(PYPY, reason="Buffered mode not supported for PyPy.")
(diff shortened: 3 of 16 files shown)


ac-optimus:
This warning is still showing up. Any suggestion how to get ridge of this?
Screenshot from 2020-02-18 02-20-06

vyasr:
@ac-optimus that warning is because the `TestFS` is not actually a test case (or a set of test cases), but is instead used as part of the tests in `TestIndexingBase`. However, because the class name starts with `Test`, pytest finds it and then chooses to ignore it. Can you make a separate PR where you rename that class to something like `FSTest`? That will remove this warning.

vishav1771:
@ac-optimus I think the warning can be removed by renaming `TestFs` to `_TestFs`. `pytest` considers every class starting with `Test` as a testing class so it should not contain any constructor.

ac-optimus:
> @ac-optimus that warning is because the `TestFS` is not actually a test case (or a set of test cases), but is instead used as part of the tests in `TestIndexingBase`. However, because the class name starts with `Test`, pytest finds it and then chooses to ignore it. Can you make a separate PR where you rename that class to something like `FSTest`? That will remove this warning.

done, thanks!

setup.cfg
```diff
@@ -7,3 +7,7 @@ python-tag = py3
 [flake8]
 max-line-length=100
 exclude = configobj,passlib,cite.py,conf.py
+
+[tool:pytest]
+filterwarnings =
+  ignore:. is deprecated.*:DeprecationWarning
```
csadorf: That pattern is way too broad. It is going to suppress many warnings (even future ones) that could actually be relevant for us. What specific warnings are intended to be suppressed here?

tests/test_buffered_mode.py
```diff
@@ -38,6 +38,13 @@
         os.chmod(path, mode)
 
 
+def deprecated_call(fun):
```
csadorf: While I'm generally a fan of tricks like this, in this case I'd prefer to use the `pytest.deprecated_call` CM explicitly at the exact location where it makes sense. The reason is that marking the complete unit test like this will make it difficult to follow which specific function is expected to trigger an expectation warning.

tests/test_indexing.py
```diff
@@ -62,6 +63,13 @@ def get_crawlers(root):
 """
 
 
+def deprecated_call(fun):
```
csadorf: See comment above.

setup.cfg
ac-optimus: I realised that this practice is not good. I was trying to reduce the deprecation warning that were still appearing(snap short below to give you rough idea, you can reproduce the warning by running pytest in the forked branch, around 131 out of 1508 warning are still there) but the above approach is too broad as you said. Can you suggest what should I do for them?
Screenshot from 2020-02-19 19-12-27

tests/test_buffered_mode.py
ac-optimus: done that! please review if this was required. Seeing the repetitive work the lazy programmer in me motivated me to take the decorator path😅 . But the reason that you made completely makes sense.

setup.cfg
csadorf: I think it's fine to install a general warnings filter with a pattern that specifically matches " The indexing module is deprecated.".

vyasr:
@ac-optimus I've merged the TestFS issue into master now, so merging in master should resolve that problem on this branch and you can move forward with addressing the other open comments here.

@zhou-pj you can review when ready.

setup.cfg
ac-optimus: I am confused! Is there anything that I should be adding to deal with remaining warnings?

ac-optimus:
> @ac-optimus I've merged the TestFS issue into master now, so merging in master should resolve that problem on this branch and you can move forward with addressing the other open comments here.
> 
> @zhou-pj you can review when ready.

yes sure!

setup.cfg
csadorf: It appears to me like every remaining warning would be covered by that?

ac-optimus: Okay, will do that and update you soon.

ac-optimus:
> Please resolve conflicts with master and add the filter for warnings caused by the deprecated indexing module as described in my earlier comment.
> 
> (I am primarily commenting here, because this PR is still in my review queue, although there are still outstanding tasks.)

done, please review.

csadorf:
> > Please resolve conflicts with master and add the filter for warnings caused by the deprecated indexing module as described in my earlier comment.
> > (I am primarily commenting here, because this PR is still in my review queue, although there are still outstanding tasks.)
> 
> done, please review.

You still need to resolve the conflicts with the base branch:
Screenshot from 2020-02-20 11-25-21

There are also still two remaining warnings in the test output. You should be able to see the output by clicking on the little green checkmark (or red x if tests failed) next to a commit id and then selecting one of the tests, e.g., *linux-python-38*, and finally by clicking on "run tests".

I believe that the warning related to `TestFS` should be resolved by properly merging the master branch.
The other warning might be related to one of your changes, I'd appreciate if you could investigate.
Screenshot from 2020-02-20 11-25-40

contributors.yaml
```diff
@@ -83,5 +83,5 @@ contributors:
   -
     family-names: Chandra
     given-names: Abhavya
-    affiliation: "India Institute of Technology, Gandhinagar"
+    affiliation: "Indian Institute of Technology, Gandhinagar"
```
bdice: Remove the trailing whitespace here -- and configure your editor to trim trailing whitespace when saving files!

ac-optimus: removed trailing white-space,  sorry for the inconvenience.  And configured my editor to trim them.😅

bdice: Thanks!

Issue #297, opened 2020-02-20
Implement the CLI with Click instead of argparse


csadorf:
### Summary

Implement the command line interface (CLI) with the [Click library](https://click.palletsprojects.com/) instead of [argparse](https://docs.python.org/3/library/argparse.html).

### Motivation

Using Click for the implementation of a CLI has numerous advantages over using argparse, including a more concise API, the enforcement of a more consistent interface, and built-in support for the implementation of test runners. In addition, the library comes with a collection of helper functions and tools, that are relevant for CLI development. Please see [this page](https://click.palletsprojects.com/en/7.x/why/) for a more comprehensive overview of why one would want to implement a CLI with Click.

### Proposed plan for implementation

Since this is a larger refactoring project, it should be executed in stages. Each stage should be introduced as a separate pull request:

- [x] Ensure that the coverage of `signac/__main__.py` is measured, see instructions [here](https://coverage.readthedocs.io/en/coverage-5.0.3/subprocess.html).
- [x] Review existing unit tests for `signac/__main_.py` and implement missing tests to reach a test coverage of at least 80%.
- [ ] Refactor the CLI implementation, replacing argparse with Click _exclusively_ for the 'find' sub command; mark all unit tests for other sub-commands as expected failures. Once the `find` sub-command was successfully refactored, refactor all other sub-commands. Replicate the unit tests for each command found in `tests/test_shell.py` in a seperate test module called `tests/test_main.py` with Click's[ built-in test runner](https://click.palletsprojects.com/en/7.x/testing/). The original tests may not fail!

### Risks

It is very likely that not all sub-commands can be implemented in the exact same way as before, because Click imposes some restrictions that argparse does not. These restrictions will likely improve the consistency and quality of the CLI, but make it a bit more difficult to create an exact replacement and may require the introduction of some minor backwards incompatibilities.

csadorf:
@vishav1771 Would you be interested in giving this a shot?

vishav1771:
Sure :+1:

csadorf:
> Sure

Awesome, thx! Please feel free to discuss questions about this project directly on this issue. Please apply the "Click" label to any PRs related to this issue so that it is easier for us to keep track of things.

bdice:
@vishav1771 Thanks for taking this on! Eventually we will want to do the same for signac-flow, so it may be helpful to document the steps you take along the way for a contributor to do this again for that package.

vishav1771:
[...]

PR #300, opened 2020-03-01
Added tests and coverage for CLI

vishav1771:
## Description

This PR implements the first two tasks of issue #297. 

- Added tests and coverage for `__main__.py` , and reached the coverage of about 80%.

## Motivation and Context
#297 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -8,3 +8,4 @@ click>=7.0
 ruamel.yaml>=0.15.89
 pytest==5.3.4
 pytest-subtests==0.3.0
+pytest-cov==2.8.1
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,3 +9,4 @@ ruamel.yaml>=0.15.89
 pytest==5.3.4
 pytest-subtests==0.3.0
 pytest-cov==2.8.1
+pymongo==3.10.1
--- a/tests/test_shell.py
+++ b/tests/test_shell.py
@@ -411,7 +411,6 @@ class TestBasicShell():
 
     def test_config_show(self):
         self.call('python -m signac init my_project'.split())
-        project = signac.Project()
         out = self.call('python -m signac config show'.split()).strip()
         cfg = config.load_config()
         expected = config.Config(cfg).write()
--- a/tests/test_shell.py
+++ b/tests/test_shell.py
@@ -67,6 +67,20 @@ class TestBasicShell():
             raise ExitCodeError("STDOUT='{}' STDERR='{}'".format(out, err))
         return out.decode()
 
+    def test_print_usage(self):
+        with pytest.raises(ExitCodeError):
+            out = self.call('python -m signac'.split())
+            assert 'usage:' in out
+
+    def test_version(self):
+        out = self.call('python -m signac --version'.split())
+        assert 'signac {}'.format(signac.__version__) in out
+
+    def test_help(self):
+        out = self.call('python -m signac --help'.split())
+        assert 'positional arguments:' in out
+        assert 'optional arguments:' in out
+
     def test_init_project(self):
         self.call('python -m signac init my_project'.split())
         assert str(signac.get_project()) == 'my_project'
@@ -114,9 +128,13 @@ class TestBasicShell():
         self.call('python -m signac init my_project'.split())
         self.call(['python', '-m', 'signac', 'job', '--create', '{"a": 0}'])
         project = signac.Project()
-        for job in project:
-            sp = self.call('python -m signac statepoint {}'.format(job).split())
-            assert project.open_job(json.loads(sp)) == job
+        job = project.open_job({'a': 0})
+        sp = self.call('python -m signac statepoint {}'.format(job).split())
+        assert project.open_job(json.loads(sp)) == job
+        sp = self.call('python -m signac statepoint'.split())
+        assert project.open_job(json.loads(sp)) == job
+        sp = self.call('python -m signac statepoint --pretty'.split())
+        assert "{'a': 0}" in sp
 
     def test_index(self):
         self.call('python -m signac init my_project'.split())
@@ -128,6 +146,9 @@ class TestBasicShell():
             assert len(list(project.index())) == 1
             assert len(list(signac.index())) == 1
         doc = json.loads(self.call('python -m signac index'.split()))
+        assert 'statepoint' in doc
+        assert doc['statepoint'] == {'a': 0}
+        doc = json.loads(self.call('python -m signac project --index'.split()))
         assert 'statepoint' in doc
         assert doc['statepoint'] == {'a': 0}
         project.open_job({'a': 0}).document['b'] = 0
@@ -151,6 +172,10 @@ class TestBasicShell():
         doc = json.loads(self.call('python -m signac document {}'.format(job_a.id).split()))
         assert 'data' in doc
         assert doc['data'] == 4
+        out = self.call('python -m signac document --pretty'.split())
+        for key, value in doc.items():
+            assert str(key) in out
+            assert str(value) in out
 
     @pytest.mark.skipif(WINDOWS, reason='Symbolic links are unsupported on Windows.')
     def test_view_single(self):
@@ -194,12 +219,26 @@ class TestBasicShell():
             assert set(job_ids) == set(project.find_job_ids())
             assert self.call('python -m signac find'.split() + ['{"a": 0}']).strip() == \
                 list(project.find_job_ids({'a': 0}))[0]
-        out = self.call('python -m signac find a 0 --sp'.split()).strip()
+
         job = project.open_job({'a': 0})
+        out = self.call('python -m signac find a 0 --sp'.split()).strip()
         assert out.strip().split(os.linesep) == [str(job.id), str(job.statepoint())]
+        out = self.call('python -m signac find a 0 --sp a'.split()).strip()
+        assert out.strip().split(os.linesep) == [str(job.id), str(job.statepoint())]
+        out = self.call('python -m signac find a 0 --sp b'.split()).strip()
+        assert out.strip().split(os.linesep) == [str(job.id), '{}']
+
         job.document['a'] = 2
         out = self.call('python -m signac find a 0 --doc'.split()).strip()
         assert out.strip().split(os.linesep) == [str(job.id), str(job.document)]
+        out = self.call('python -m signac find a 0 --doc a'.split()).strip()
+        assert out.strip().split(os.linesep) == [str(job.id), str(job.document)]
+        out = self.call('python -m signac find a 0 --doc b'.split()).strip()
+        assert out.strip().split(os.linesep) == [str(job.id), '{}']
+        out = self.call('python -m signac find a 0 --show --one-line'.split()).strip()
+        assert str(job.id) in out
+        assert '{"a": 0}' in out
+        assert '{"a": 2}' in out
 
         # Test the doc_filter
         for job in project.find_jobs():
@@ -238,6 +277,12 @@ class TestBasicShell():
         assert job in project_a
         assert len(project_b) == 1
         assert job in project_b
+        self.call("python -m signac clone {} {}"
+                  .format(os.path.join(self.tmpdir.name, 'b'), job.id).split())
+        assert len(project_a) == 1
+        assert job in project_a
+        assert len(project_b) == 1
+        assert job in project_b
 
     def test_move(self):
         self.call('python -m signac init ProjectA'.split())
@@ -256,6 +301,13 @@ class TestBasicShell():
         assert job not in project_a
         assert len(project_b) == 1
         assert job in project_b
+        project_a.open_job({'a': 0}).init()
+        self.call("python -m signac move {} {}"
+                  .format(os.path.join(self.tmpdir.name, 'b'), job.id).split())
+        assert len(project_a) == 1
+        assert job in project_a
+        assert len(project_b) == 1
+        assert job in project_b
 
     def test_remove(self):
         self.call('python -m signac init my_project'.split())
@@ -268,12 +320,13 @@ class TestBasicShell():
         assert job_to_remove in project
         assert job_to_remove.doc.a == 0
         assert len(job_to_remove.doc) == 1
-        with pytest.deprecated_call():
-            self.call('python -m signac rm --clear {}'.format(job_to_remove.get_id()).split())
+        self.call('python -m signac rm --clear {}'.format(job_to_remove.id).split())
         assert job_to_remove in project
         assert len(job_to_remove.doc) == 0
-        with pytest.deprecated_call():
-            self.call('python -m signac rm {}'.format(job_to_remove.get_id()).split())
+        self.call('python -m signac -v rm {}'.format(job_to_remove.id).split())
+        assert job_to_remove not in project
+        with pytest.raises(ExitCodeError):
+            self.call('python -m signac -v rm {}'.format(job_to_remove.id).split())
         assert job_to_remove not in project
 
     def test_schema(self):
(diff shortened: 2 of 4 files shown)


csadorf:
This looks really good! Thank you for pushing work-in-progress.

There appears to be some kind of unrelated issue with tests on py36. I've tried to investigate a little bit, but so far am not sure what is going on. I suggest to ignore that for now.

vishav1771:
@csadorf I have added a test for `export` option. Please have a look at the CI error.
I think this can be removed by adding `os.mkdir`. Is that okay? (because we don't need that in `export_to`)

csadorf:
> @csadorf I have added a test for `export` option. Please have a look at the CI error.
> I think this can be removed by adding `os.mkdir`. Is that okay? (because we don't need that in `export_to`)

@vishav1771 It should not be necessary to create the directory that we export to, because it is created automatically. So I am not sure what is causing the issue.

Things to check:
- [ ] Does the `/tmp` directory exist and do we have write access to it?
- [ ] Is there some kind of issue with finding the `python` executable?

csadorf:
@vishav1771 Impressive progress on the testing front! You can use [this page](https://codecov.io/gh/glotzerlab/signac/src/dc12796366edcd93d89c1bd023a68f2ceb3f027f/signac/__main__.py) provided by codecov to identify missing hotspots. As you know, the goal is to reach 80% coverage, but please let us know if you are running into trouble reaching that.

bdice:
@vishav1771 I looked at the lines that are missing coverage. A lot of the coverage misses have to do with deprecated modules, which would probably be a pain to test. I would focus on non-deprecated code paths. Here are a few easy tests that should be near the top of the `__main__` tests. (I haven't checked if these tests pass but they're fairly close.)

```python
    def test_print_usage(self):
        out = self.call('python -m signac'.split())
        assert 'usage:' in out

    def test_version(self):
        out = self.call('python -m signac --version'.split())
        assert 'signac {}'.format(signac.__version__) in out

    def test_help(self):
        out = self.call('python -m signac --help'.split())
        assert 'positional arguments:' in out
        assert 'optional arguments:' in out

**Comment by vishav1771** [comment]
@csadorf @bdice Thanks for the review, and providing with the necessary changes. I'll try to reach the required 80% coverage mark.
While writing test for  `config verify`, I was getting the error for `signac config verify `.
```
Error: validate() got an unexpected keyword argument 'skip_missing'
  File "C:\Users\visha\AppData\Local\Continuum\anaconda3\Scripts\signac-script.py", line 11, in <module>
    load_entry_point('signac==1.3.0', 'console_scripts', 'signac')()
  File "C:\Users\visha\AppData\Local\Continuum\anaconda3\lib\site-packages\signac-1.3.0-py3.7.egg\signac\__main__.py", line 1835, in main
  File "C:\Users\visha\AppData\Local\Continuum\anaconda3\lib\site-packages\signac-1.3.0-py3.7.egg\signac\__main__.py", line 809, in main_config_verify
  File "C:\Users\visha\AppData\Local\Continuum\anaconda3\lib\site-packages\signac-1.3.0-py3.7.egg\signac\__main__.py", line 722, in verify_config
  File "C:\Users\visha\AppData\Local\Continuum\anaconda3\lib\site-packages\signac-1.3.0-py3.7.egg\signac\common\config.py", line 133, in verify
TypeError: validate() got an unexpected keyword argument 'skip_missing'
```
Can you please verify this behaviour. Also, can you provide me with examples where the verification of config file fails.

**Comment by csadorf** [comment]
> @csadorf @bdice Thanks for the review, and providing with the necessary changes. I'll try to reach the required 80% coverage mark.
> While writing test for `config verify`, I was getting the error for `signac config verify `.
> 
> ```
> Error: validate() got an unexpected keyword argument 'skip_missing'
> Traceback (most recent call last):
>   File "C:\Users\visha\AppData\Local\Continuum\anaconda3\Scripts\signac-script.py", line 11, in <module>
>     load_entry_point('signac==1.3.0', 'console_scripts', 'signac')()
>   File "C:\Users\visha\AppData\Local\Continuum\anaconda3\lib\site-packages\signac-1.3.0-py3.7.egg\signac\__main__.py", line 1835, in main
>   File "C:\Users\visha\AppData\Local\Continuum\anaconda3\lib\site-packages\signac-1.3.0-py3.7.egg\signac\__main__.py", line 809, in main_config_verify
>   File "C:\Users\visha\AppData\Local\Continuum\anaconda3\lib\site-packages\signac-1.3.0-py3.7.egg\signac\__main__.py", line 722, in verify_config
>   File "C:\Users\visha\AppData\Local\Continuum\anaconda3\lib\site-packages\signac-1.3.0-py3.7.egg\signac\common\config.py", line 133, in verify
> TypeError: validate() got an unexpected keyword argument 'skip_missing'
> ```
> 
> Can you please verify this behaviour. Also, can you provide me with examples where the verification of config file fails.

Looks like you identified a completely unrelated bug. That's great, because it shows how important testing is and we were able to detect this bug because of your hard work.

I've looked into this locally and found that #302 should resolve the issue. Once #302 is merged, we can rebase this branch and move forward.

**Comment by vishav1771** [comment]
Thanks, @csadorf , I'll continue with this PR after #302 is merged.
Also, while playing around, I think I found another unrelated bug.
While executing `signac sync C:\Users\visha\Desktop\test C:\Users\visha\Desktop\workspace -w`
I am getting the error: 
```
Error: 'schema_version'
```
Can you please verify this?

**Comment by bdice** [comment]
I can see the error as well. This appears to be caused during the explicit instantiation of the Project config here, which doesn't include the mandatory `schema_version` key: https://github.com/glotzerlab/signac/blob/fa2d8c40dd5f0d132dfc9c67a2570277c264f8af/signac/__main__.py#L474-L478

I suppose we have to assume a schema version for the workspace (since there's no project configuration to read from) or disallow the option to synchronize directly to the workspace.

**Comment by vishav1771** [comment]
I think we can pass  ` _ignore_schema_version=True` while creating Project to remove the error.

https://github.com/glotzerlab/signac/blob/fa2d8c40dd5f0d132dfc9c67a2570277c264f8af/signac/__main__.py#L474-L478

**Comment by csadorf** [comment]
> I think we can pass ` _ignore_schema_version=True` while creating Project to remove the error.
> 
> https://github.com/glotzerlab/signac/blob/fa2d8c40dd5f0d132dfc9c67a2570277c264f8af/signac/__main__.py#L474-L478

I would advocate for using `_ignore_schema_version=True` as a work-around, but generally deprecating this option.

@bdice The problem further shows why allowing workspaces outside project root directories is not a good idea. In this case there is absolutely no way for us to know what schema that particular workspace is possibly maintained with. The relationship between project and workspace should be bijective.

**Comment by csadorf** [comment]
> Thanks, @csadorf , I'll continue with this PR after #302 is merged.

You can probably continue working on this PR in parallel, you just need to ignore that particular code section.

**Comment by bdice** [comment]
@vishav1771 #302 has been merged.

File: tests/test_shell.py [hunk 1]
```diff
@@ -113,9 +130,13 @@ def test_statepoint(self):
         self.call('python -m signac init my_project'.split())
         self.call(['python', '-m', 'signac', 'job', '--create', '{"a": 0}'])
         project = signac.Project()
-        for job in project:
-            sp = self.call('python -m signac statepoint {}'.format(job).split())
-            assert project.open_job(json.loads(sp)) == job
+        job = project.open_job({'a': 0})
+        sp = self.call('python -m signac statepoint {}'.format(job).split())
```
bdice: Let's make it explicit that we use the job id here.
```suggestion
        sp = self.call('python -m signac statepoint {}'.format(job.id).split())
```

tests/test_shell.py
```diff
@@ -113,9 +130,13 @@ def test_statepoint(self):
...
+        assert project.open_job(json.loads(sp)) == job
+        sp = self.call('python -m signac statepoint'.split())
+        assert project.open_job(json.loads(sp)) == job
+        sp = self.call('python -m signac statepoint --pretty'.split())
+        assert "{'a': 0}" in sp
```
bdice: I found some inconsistent behavior. In general, I believe the signac CLI should always have JSON-compatible output where possible. In a test project, I ran the following:
```bash
$ signac statepoint
{"a": {"b": 0, "c": 0}}
$ signac statepoint --pretty
{'a': {'b': 0, 'c': 0}}
```
The `--pretty` output is not JSON-compatible because JSON requires double quotes, not single quotes. @csadorf Is this something we've thought about before? I don't know how `--pretty` is implemented.

tests/test_shell.py
```diff
@@ -129,13 +150,35 @@ def test_index(self):
...
         project.open_job({'a': 0}).document['b'] = 0
         doc = json.loads(self.call('python -m signac index'.split()))
         assert 'statepoint' in doc
         assert doc['statepoint'] == {'a': 0}
         assert 'b' in doc
         assert doc['b'] == 0
 
+    def test_document(self):
+        self.call('python -m signac init my_project'.split())
+        self.call('python -m signac project --access'.split())
```
bdice: This `signac project --access` command doesn't seem related to testing job documents, so it probably goes somewhere else. The generated `signac_access.py` also appears to use deprecated functionality. @csadorf Do you know if this is supposed to be deprecated overall? I'm not familiar with this part of the codebase, I didn't know this feature for creating "access modules" existed and am not sure whether it still serves a purpose.

tests/test_shell.py
```diff
@@ -129,13 +150,35 @@ def test_index(self):
...
+        job_a.init()
+        assert len(project) == 1
+        job_a.document['data'] = 4
+        doc = json.loads(self.call('python -m signac document'.split()))
+        assert 'data' in doc
+        assert doc['data'] == 4
+        doc = json.loads(self.call('python -m signac document {}'.format(job_a.id).split()))
+        assert 'data' in doc
+        assert doc['data'] == 4
+        out = self.call('python -m signac document --pretty'.split())
```
bdice: Same issue as above with JSON compatibility and single quotes with `--pretty`, double quotes without `--pretty`.

tests/test_shell.py
```diff
@@ -188,6 +251,68 @@ def test_find(self):
...
 
+    def test_diff(self):
+        self.call('python -m signac init ProjectA'.split())
+        project = signac.Project()
+        job_a = project.open_job({"a": 0, "b": 1})
+        job_a.init()
+        job_b = project.open_job({"a": 0, "b": 0})
+        job_b.init()
+        out = self.call('python -m signac diff {} {}'.format(job_a.id, job_b.id).split())
+        expected = [str(job_a.id), "{'b': 1}", str(job_b.id), "{'b': 0}"]
```
bdice: Same potential issue with single/double quotes.

tests/test_shell.py
```diff
@@ -199,13 +324,234 @@ def test_remove(self):
...
+                      .format(os.path.join(self.tmpdir.name, 'b'), self.tmpdir.name).split())
+        assert job_dst.document != job_src.document
+
+        self.call('python -m signac sync {} {} --no-key'
+                  .format(os.path.join(self.tmpdir.name, 'b'), self.tmpdir.name).split())
+        assert job_dst.document != job_src.document
+        assert job_dst.document['a'] != job_src.document['a']
+        assert job_dst.document['nested'] != job_src.document['nested']
+
+        reset()     # only sync a
```
bdice: Split these comments onto their own lines, above `reset()`. Also add a comment explaining what is happening for all the options (`--no-key`, `--all-key`, etc.).

tests/test_shell.py
```diff
@@ -199,13 +324,234 @@ def test_remove(self):
...
+
+    def test_sync_file(self):
+        self.call('python -m signac init ProjectA'.split())
+        project_a = signac.Project()
+        project_b = signac.init_project('ProjectB', os.path.join(self.tmpdir.name, 'b'))
+        job_src = project_a.open_job({'a': 0})
+        job_dst = project_b.open_job({'a': 0})
+        for i, job in enumerate([job_src, job_dst]):
+            with job:
+                with open('test', 'w') as file:
```
bdice: Could be simplified so that `with job` isn't needed: `with open(job.fn('test'), 'w') as file:`

tests/test_shell.py
```diff
@@ -199,13 +324,234 @@ def test_remove(self):
...
+        assert "Import failed" in err
+        self.call("python -m signac import {} --sync".format(prefix_data).split(), error=True)
+        assert len(project_a) == 4
+        assert 'a' in job_dst.document
+        assert job_dst.document['a'] == 0
+        out = self.call(
+            'python -m signac import {} --sync-interactive'.format(prefix_data),
+            'print(str(tmp_project), len(tmp_project)); exit()', shell=True)
+        assert 'ProjectA' in out
+        assert '4' in out
```
bdice: Combine if possible.
```suggestion
        assert 'ProjectA 4' in out
```

tests/test_shell.py
```diff
@@ -215,7 +561,15 @@ def test_shell(self):
             'print(str(project), job, len(list(jobs))); exit()', shell=True)
         assert out.strip() == '>>> {} None {}'.format(project, len(project))
 
+        out = self.call(
+            'python -m signac shell -c print(str(project),len(list(jobs)))'
+            .split())
```
bdice: This is a little bit of a hack, it looks like you're avoiding spaces in the `-c` argument to prevent splitting that token. I suggest either specifying this as input or using another way to tokenize manually like:
```suggestion
        cmd = 'python -m signac shell -c'.split() + ['print(str(project), len(list(jobs)))']
        out = self.call(cmd)
```

tests/test_shell.py
```diff
@@ -244,9 +598,97 @@ def test_shell_with_jobs_and_selection_only_one_job(self):
...
+        assert out.split(os.linesep) == expected
+
+        out = self.call('python -m signac config --global show'.split()).strip()
+        cfg = config.read_config_file(config.FN_CONFIG)
+        expected = config.Config(cfg).write()
+        assert out.split(os.linesep) == expected
+
+    def test_config_set(self):
+        self.call('python -m signac init my_project'.split())
+        self.call('python -m signac config  set a b'.split())
```
bdice:
```suggestion
        self.call('python -m signac config set a b'.split())
```

tests/test_shell.py
csadorf: We use `pprint` to implement pretty. We should probably have used some JSON prettifying instead. Feel free to classify as bug and fix without deprecation cycle.

tests/test_shell.py
csadorf: This is deprecated.

tests/test_shell.py
csadorf: Same comment as above.

tests/test_shell.py
vyasr: Switching from pprint to something like `json.dumps` would also give us something closer to the output that @phmalek requested recently on Gitter

tests/test_shell.py
bdice: I created #303 to address this.

tests/test_shell.py
bdice: While `json.dumps` offers an `indent`, there is no way to have the equivalent of pprint's `depth` flag with `json.dumps`:

```python
>>> from pprint import pprint
>>> pprint({'a': {'very': {'deep': 'dictionary'}}})
{'a': {'very': {'deep': 'dictionary'}}}
>>> pprint({'a': {'very': {'deep': 'dictionary'}}}, depth=2)
{'a': {'very': {...}}}
```
@csadorf @vyasr Is this "depth" a hard requirement for pretty output?

csadorf: I did find that kind of important in my interactive work, yes.

bdice: In that case, I recommend to make no changes to existing behavior. If "pretty" output was just about indenting it, then that can be solved with `json.dumps`. Since there's not an easy way to solve this, let's leave the current behavior as-is. If you agree, we can resolve the conversations about pretty output.

tests/test_shell.py
vishav1771: I don't we can combine both of them as `tmp_project.id` is combination of ` project.id, str(uuid.uuid4()) `

bdice: Oh, okay. Nevermind. I'll resolve this.

requirements-dev.txt
```diff
@@ -8,3 +8,5 @@ click>=7.0
 ruamel.yaml>=0.15.89
 pytest==5.3.4
 pytest-subtests==0.3.0
+pytest-cov==2.8.1
+pymongo==3.10.1
```
csadorf: We are actually trying to move away from the optional pymongo integration, but I guess we can always drop that at some later point.

tests/test_shell.py
```diff
@@ -64,8 +65,24 @@ def call(self, command, input=None, shell=False):
...
         if p.returncode != 0:
             raise ExitCodeError("STDOUT='{}' STDERR='{}'".format(out, err))
+        if error:
+            return err.decode()
         return out.decode()
 
+    def test_print_usage(self):
+        with pytest.raises(ExitCodeError):
+            out = self.call('python -m signac'.split())
+            assert 'usage:' in out
```
csadorf: Considering that `pytest.raises(..)` is implying that we assume that the `call()` is raising an `ExitCodeError`, would this assert never be tested?

tests/test_shell.py
```diff
@@ -113,9 +130,13 @@ def test_statepoint(self):
         self.call('python -m signac init my_project'.split())
         self.call(['python', '-m', 'signac', 'job', '--create', '{"a": 0}'])
         project = signac.Project()
-        for job in project:
-            sp = self.call('python -m signac statepoint {}'.format(job).split())
-            assert project.open_job(json.loads(sp)) == job
+        job = project.open_job({'a': 0})
+        sp = self.call('python -m signac statepoint {}'.format(job.id).split())
+        assert project.open_job(json.loads(sp)) == job
```
csadorf: Can we add an `assert len(project) == 1` below this line, I was initially confused by the following asserts.

tests/test_shell.py
```diff
@@ -188,6 +250,68 @@ def test_find(self):
...
+    def test_clone(self):
+        self.call('python -m signac init ProjectA'.split())
+        project_a = signac.Project()
+        project_b = signac.init_project('ProjectB', os.path.join(self.tmpdir.name, 'b'))
+        job = project_a.open_job({'a': 0})
+        job.init()
+        assert len(project_a) == 1
+        assert len(project_b) == 0
+        with pytest.raises(ExitCodeError):
+            self.call("python -m signac clone {} 9bfd29df07674bc5"
```
csadorf: Can we add a comment here about the meaning of that hard-coded id? Is it expected to already exist?

tests/test_shell.py
```diff
@@ -188,6 +250,68 @@ def test_find(self):
...
+    def test_move(self):
+        self.call('python -m signac init ProjectA'.split())
+        project_a = signac.Project()
+        project_b = signac.init_project('ProjectB', os.path.join(self.tmpdir.name, 'b'))
+        job = project_a.open_job({'a': 0})
+        job.init()
+        assert len(project_a) == 1
+        assert len(project_b) == 0
+        with pytest.raises(ExitCodeError):
+            self.call("python -m signac move {} 9bfd29df07674bc5"
```
csadorf: Same comment as above.

tests/test_shell.py
```diff
@@ -199,14 +323,233 @@ def test_remove(self):
         assert job_to_remove in project
         assert job_to_remove.doc.a == 0
         assert len(job_to_remove.doc) == 1
-        with pytest.deprecated_call():
```
csadorf: Are those calls not considered deprecated anymore?

tests/test_shell.py
```diff
@@ -199,14 +323,233 @@ def test_remove(self):
...
-            self.call('python -m signac rm --clear {}'.format(job_to_remove.get_id()).split())
+        self.call('python -m signac rm --clear {}'.format(job_to_remove.id).split())
         assert job_to_remove in project
         assert len(job_to_remove.doc) == 0
-        with pytest.deprecated_call():
-            self.call('python -m signac rm {}'.format(job_to_remove.get_id()).split())
+        self.call('python -m signac -v rm {}'.format(job_to_remove.id).split())
+        assert job_to_remove not in project
+        with pytest.raises(ExitCodeError):
```
csadorf: I think some comment (here and in similar locations elsewhere) as to why certain calls are expected to fail would be very helpful.

tests/test_shell.py
vishav1771: The deprecated call was because of `get_id` function. I removed it so that it's not needed anymore.

requirements-dev.txt
vishav1771: `pymongo` is required for the `config host` command. This can be removed after the its deprecation cycle.

tests/test_shell.py
vishav1771: Sorry, I didn't realize that earlier. will change.

tests/test_shell.py
```diff
@@ -269,9 +273,6 @@ def test_clone(self):
         job.init()
         assert len(project_a) == 1
         assert len(project_b) == 0
-        with pytest.raises(ExitCodeError):
-            self.call("python -m signac clone {} 9bfd29df07674bc5"
-                      .format(os.path.join(self.tmpdir.name, 'b')).split())
```
csadorf: Why was this removed? This seems like the negative test corresponding to the positive test below.

vishav1771: This test was not removed, I just changed its position. I was trying to arrange the test in a certain order (like first do all positive tests then negative tests).

csadorf:
@vyasr Merge at your discretion.

PR #308, opened 2020-03-19
Deprecation of create_access_module method

### LINKED ISSUE #303: `signac project --access` should be deprecated
This `signac project --access` generates a file `signac_access.py` that uses deprecated functionality. This feature of the CLI should be deprecated.

Originally discussed with @csadorf in #300.

kidrahahjo:
## Description

I have deprecated the `create_access_module` method but haven't adapted with its deprecation warning yet as no warnings were showing up on my local machine.
## Motivation and Context

fixes #303 
## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -1448,6 +1448,9 @@ class Project(object):
         for doc in docs:
             yield doc
 
+    @deprecated(
+        deprecated_in="1.4", removed_in="2.0", current_version=__version__,
+        details="Uses deprecated functionality")
     def create_access_module(self, filename=None, master=True):
         """Create the access module for indexing
 
--- a/contributors.yaml
+++ b/contributors.yaml
@@ -84,4 +84,8 @@ contributors:
     family-names: Chandra
     given-names: Abhavya
     affiliation: "Indian Institute of Technology, Gandhinagar"
+  -
+    family-names: Ojha
+    given-names: Hardik
+    affiliation: "Indian Institute of Technology Roorkee"
 ...


signac/contrib/project.py
```diff
@@ -1448,6 +1448,9 @@ class Crawler(SignacProjectCrawler):
         for doc in docs:
             yield doc
 
+    @deprecated(
+        deprecated_in="1.4", removed_in="2.0", current_version=__version__,
+        details="Uses deprecated functionality")
```
```suggestion
You have to use the "next" version -- you can't deprecate something in a version that has already been released. 😄
```suggestion
        deprecated_in="1.5", removed_in="2.0", current_version=__version__,
        details="Access modules are deprecated.")
```

bdice:
@kidrahahjo Next steps:
- [ ] Make sure the deprecation warning appears when calling `signac project --access`, in addition to the Python API `project.create_access_module()`. It doesn't look like it's showing in the test output right now, but I'm not sure if it's just untested or if it's not showing for another reason.
- [ ] Hide this deprecation warning wherever it occurs in testing output, use pytest's "deprecated call" feature or something like that.
- [ ] Add access modules to the "Deprecated" section of the changelog for the next release.

signac/contrib/project.py
kidrahahjo: Done :stuck_out_tongue:

kidrahahjo:
> @kidrahahjo Next steps:
> 
> * [ ]  Make sure the deprecation warning appears when calling `signac project --access`, in addition to the Python API `project.create_access_module()`. It doesn't look like it's showing in the test output right now, but I'm not sure if it's just untested or if it's not showing for another reason.
> * [ ]  Hide this deprecation warning wherever it occurs in testing output, use pytest's "deprecated call" feature or something like that.
> * [ ]  Add access modules to the "Deprecated" section of the changelog for the next release.

@bdice Can I ignore the deprecation warning globally instead of using the pytest's "deprecated call" feature?

bdice:
@kidrahahjo Ignoring the warning is not quite the same. Using `pytest.deprecated_call` ensures that the code actually triggers a deprecation warning. That means it is testable behavior.
https://docs.pytest.org/en/3.1.0/warnings.html#ensuring-function-triggers

kidrahahjo:
> @kidrahahjo Ignoring the warning is not quite the same. Using `pytest.deprecated_call` ensures that the code actually triggers a deprecation warning. That means it is testable behavior.
> https://docs.pytest.org/en/3.1.0/warnings.html#ensuring-function-triggers

@bdice I will have a look and then do the changes where required. Thank you for providing the link.

kidrahahjo:
@tommy-waltmann @zhou-pj 
Even after testing the module test_shell.py without using `pytest.deprecated_call():` then too it doesn't show up any warning. So the tests are failing. I don't know why the warning is now showing up. I think that this should not be merged until the deprecation warning starts showing up. Please help me with your views on this comment.

tommy-waltmann:
@kidrahahjo There is nothing that immediately jumps out to me as the reason a warning is not being generated, you may have to experiment a little on your own machine as to why this is happening.

For example, you might run only this unit test and make sure, via pdb or a print statement, that the code does in fact enter the `if args.access` block. You might also want to check and make sure that the print statement right after the `warnings.warn` line is being printed to the console.

If neither of those work, you can try some of your own experiments to determine what is going on.

kidrahahjo:
@tommy-waltmann @zhou-pj @bdice  
I don't think that we should use `pytest.deprecated_call()` in the tests because as we are not calling the function directly rather we're running `python -m signac project --access` in the tests through the `call` method. If a user runs this command then he might get the warning. Please let me know if I'm going in the wrong direction.

bdice:
@kidrahahjo I am going to push some changes. I finally figured out something that had been confusing me. I forgot that the deprecation warnings won't actually appear until `__version__ >= deprecated_in`. So until we release 1.5.0, the deprecation warnings will not appear. I kept thinking I would see deprecations appearing in the output but they didn't show up...

signac/__main__.py
```diff
@@ -115,6 +116,9 @@
 """
 
 
+warnings.simplefilter("default")
```
bdice, line 119: @glotzerlab/signac-maintainers Can I get a second opinion on this one-line fix for issue #317? I put it in this PR because it's related to deprecation warnings.

kidrahahjo:
> @kidrahahjo I am going to push some changes. I finally figured out something that had been confusing me. I forgot that the deprecation warnings won't actually appear until `__version__ >= deprecated_in`. So until we release 1.5.0, the deprecation warnings will not appear. I kept thinking I would see deprecations appearing in the output but they didn't show up...

@bdice I tried with a code which had `deprecated_in<__version`__ too but didn't got the warning.

signac/__main__.py
bdice: I verified that warnings are sent to stderr, not stdout.

kidrahahjo:
I also tried using the print statement with `file=sys.stdout`. That too didn't work out for helping me debug. Can you let me know why that happened?

setup.cfg
```diff
@@ -25,7 +25,7 @@ omit =
 
 [tool:pytest]
 filterwarnings =
-	ignore: .*[The indexing module | get_statepoint] is deprecated.*: DeprecationWarning
+	ignore: .*get_statepoint is deprecated.*: DeprecationWarning
```
bdice, line 28: @kidrahahjo I'm confused. Why did you change this?

tests/test_shell.py
```diff
@@ -148,9 +148,9 @@ def test_index(self):
         project = signac.Project()
         project.open_job({'a': 0}).init()
         assert len(project) == 1
-        with pytest.deprecated_call():
-            assert len(list(project.index())) == 1
-            assert len(list(signac.index())) == 1
+        # with pytest.deprecated_call():
+        assert len(list(project.index())) == 1
+        assert len(list(signac.index())) == 1
```
bdice, line 153: @kidrahahjo I'm confused. Why did you change this?

PR #310, opened 2020-03-21
Improve Project docstrings.

bdice:
## Description
A couple docstrings were incomplete.

## Motivation and Context
I didn't know what the `**kwargs` in the signature would do, so I added documentation for them.

## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -1568,7 +1568,7 @@ class Project(object):
 
     @classmethod
     def get_project(cls, root=None, search=True, **kwargs):
-        """Find a project configuration and return the associated project.
+        r"""Find a project configuration and return the associated project.
 
         :param root:
             The starting point to search for a project, defaults to the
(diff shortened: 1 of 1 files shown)


bdice:
Due to the test failure, I opened #311 and #312.

signac/contrib/project.py
```diff
@@ -1568,7 +1568,7 @@ def init_project(cls, name, root=None, workspace=None, make_dir=True):
 
     @classmethod
     def get_project(cls, root=None, search=True, **kwargs):
-        """Find a project configuration and return the associated project.
+        r"""Find a project configuration and return the associated project.
```
atravitz, line 1571: is using raw strings everywhere a convention even when theyre not necessary?

bdice: tl;dr: Using a raw string here is a non-obvious trick for properly documenting `**kwargs` with Sphinx.

This has `**kwargs` and the only way to make a valid docstring for Sphinx is to use `\*\*kwargs` below, and mark it as a raw string so that the backslashes aren't interpreted as escape characters. If you use `**kwargs` without the backslashes, Sphinx will interpret that as bold formatting. If you use `\*\*kwargs` without the raw string, then it will attempt to interpret those as escape sequences. If you use `\\*\\*kwargs` to escape the slashes without the raw string, I think something else breaks.

Issue #315, opened 2020-03-25
Convert all docstrings to numpy style


vyasr:
As part of our overall docs overhaul (see glotzerlab/signac-docs#64), we want to convert our docstrings to numpy style (as decided in glotzerlab/signac-docs#74). The best automated tool I'm familiar with for this task is [pyment](https://github.com/dadadel/pyment). In addition to converting docstrings, it will also generate docstrings for functions, classes, etc that are missing docstrings entirely. However, the conversion will require significant manual review to ensure that all docstrings are converted correctly.

vyasr:
@ac-optimus could you comment on this issue so that I can assign you?

ac-optimus:
hi, can I work on this issue?

mikemhenry:
List of files that need to be fixed:

- [ ] `benchmark.py`
- [ ] `doc/conf.py`
- [ ] `setup.py`
- [ ] `signac/common/configobj/__init__.py`
- [ ] `signac/common/configobj/validate.py`
- [ ] `signac/common/configobj/_version.py`
- [ ] `signac/common/config.py`
- [ ] `signac/common/connection.py`
- [ ] `signac/common/crypt.py`
- [ ] `signac/common/host.py`
- [ ] `signac/common/__init__.py`
- [ ] `signac/common/validate.py`
- [ ] `signac/contrib/filesystems.py`
- [ ] `signac/contrib/import_export.py`
- [ ] `signac/contrib/indexing.py`
- [ ] `signac/contrib/__init__.py`
- [ ] `signac/contrib/migration/__init__.py`
- [ ] `signac/contrib/migration/v0_to_v1.py`
- [ ] `signac/contrib/mpipool.py`
- [ ] `signac/core/attrdict.py`
- [ ] `signac/core/dict_manager.py`
- [ ] `signac/core/h5store.py`
- [ ] `signac/core/__init__.py`
- [ ] `signac/core/json.py`
- [ ] `signac/core/synceddict.py`
- [ ] `signac/db/database.py`
- [ ] `signac/db/__init__.py`
- [ ] `signac/diff.py`
- [ ] `signac/__init__.py`
- [ ] `signac/__main__.py`
- [ ] `signac/sync.py`
- [ ] `signac/syncutil.py`
- [ ] `signac/testing.py`
- [ ] `signac/version.py`
- [ ] `signac/warnings.py`
- [ ] `.sync-zenodo-metadata.py`

bdice:
[...]

PR #318, opened 2020-04-01
numpydoc style(signac/contrib/schema.py)

ac-optimus:
## Description

issue #315 
## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change<sup>1</sup>

<sup>1</sup>The change breaks (or has the potential to break) existing functionality.

## Checklist:

- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
- [x] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [ ] I have updated the API documentation as part of the package doc-strings.
- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
- [ ] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added all related issue and pull request numbers for future reference (if applicable). See example below.

Change of doc string to `numpydoc` in schema.py

PR #319, opened 2020-04-02
numpydoc style(signac/contrib/job.py) 

ac-optimus:
## Description

issue #315 
## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

Change of doc string to `numpydoc` in schema.py

### DIFF
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -23,8 +23,13 @@ logger = logging.getLogger(__name__)
 
 
 class _sp_save_hook(object):
-    """ """
+    """
 
+    Parameters
+    ----------
+    jobs :
+
+    """
     def __init__(self, *jobs):
         self.jobs = list(jobs)
 
@@ -47,10 +52,11 @@ class Job(object):
 
     Parameters
     ----------
+    project :
 
-    Returns
-    -------
+    statepoint :
 
+    _id :
 
     """
 
@@ -92,14 +98,10 @@ class Job(object):
     def get_id(self):
         """The job's statepoint's unique identifier.
 
-        Parameters
-        ----------
-
         Returns
         -------
         str
             The job id.
-
 
         """
         return self._id
@@ -108,14 +110,10 @@ class Job(object):
     def id(self):
         """The unique identifier for the job's statepoint.
 
-        Parameters
-        ----------
-
         Returns
         -------
         str
             The job id.
-
 
         """
         return self._id
@@ -138,14 +136,10 @@ class Job(object):
     def workspace(self):
         """Each job is associated with a unique workspace directory.
 
-        Parameters
-        ----------
-
         Returns
         -------
         str
             The path to the job's workspace directory.
-
 
         """
         return self._wd
@@ -169,16 +163,12 @@ class Job(object):
         new_statepoint : mapping
             The job's new state point.
 
-        Returns
-        -------
-
         Raises
         ------
         DestinationExistsError
             If a job associated with the new state point is already initialized.
         OSError
             If the move failed due to an unknown system related error.
-
 
         """
         dst = self._project.open_job(new_statepoint)
@@ -221,10 +211,6 @@ class Job(object):
         new_sp :
             (Default value = None)
 
-        Returns
-        -------
-
-
         """
 
         if new_sp is None:
@@ -249,9 +235,6 @@ class Job(object):
             which are currently part of the job's state point.
             Use with caution! (Default value = False)
 
-        Returns
-        -------
-
         Raises
         ------
         KeyError
@@ -261,7 +244,6 @@ class Job(object):
             If a job associated with the new state point is already initialized.
         OSError
             If the move failed due to an unknown system related error.
-
 
         """
         statepoint = self.statepoint()
@@ -286,12 +268,8 @@ class Job(object):
             `sp_dict = job.statepoint()` instead of `sp = job.statepoint`.
             For more information, see :class:`~signac.JSONDict`.
 
-        Parameters
-        ----------
-
         Returns
         -------
-
 
         """
         return self._statepoint
@@ -303,12 +281,7 @@ class Job(object):
         Parameters
         ----------
         new_sp :
-
-
-        Returns
-        -------
-
-
+            new state point to be assign.
         """
         self._reset_sp(new_sp)
 
@@ -319,18 +292,7 @@ class Job(object):
 
     @sp.setter
     def sp(self, new_sp):
-        """Alias for :attr:`Job.statepoint`.
-
-        Parameters
-        ----------
-        new_sp :
-
-
-        Returns
-        -------
-
-
-        """
+        """Alias for :attr:`Job.statepoint`."""
         self.statepoint = new_sp
 
     @property
@@ -344,14 +306,10 @@ class Job(object):
             For more information, see :attr:`Job.statepoint` or
             :class:`~signac.JSONDict`.
 
-        Parameters
-        ----------
-
         Returns
         -------
         class : `~signac.JSONDict`
             The job document handle.
-
 
         """
         if self._document is None:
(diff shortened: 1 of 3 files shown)


signac/contrib/job.py
```diff
@@ -23,14 +23,17 @@
...
         self.jobs = list(jobs)
 
     def load(self):
+        """ """
         pass
 
     def save(self):
+        """ """
         for job in self.jobs:
             job._reset_sp()
```
ac-optimus, line 38: I am not sure what class `_sp_save_hook` does, what should I add in its doc string?

tommy-waltmann: Maybe just say "used to override default save and load behavior for synced data structures"? @vyasr @mikemhenry Any other ideas? The entire class is internal, so the comments could be more oriented towards developers instead of users.

vyasr: Definitely agree that the docs should be more for developers. The goal of the `_sp_save_hook` is to force the job's statepoint resetting to happen whenever the statepoint is changed. Resetting a statepoint requires recomputing the hash and moving the folder, which is outside the scope of just what the synced dict does, so that's what this documentation should indicate.

signac/contrib/job.py
```diff
@@ -23,14 +23,22 @@
 
 
 class _sp_save_hook(object):
+    """
```
mikemhenry, line 26: Maybe something like "Force the job's statepoint to reset to whenever the statepoint is changed. Resetting a statepoint requires recomputing the hash and moving the folder"

signac/contrib/job.py
```diff
@@ -254,23 +319,24 @@ def document(self):
...
     def doc(self):
-        """Alias for :attr:`Job.document`.
-
-        .. warning::
-
-            If you need a deep copy that will not modify the underlying
-            persistent JSON file, use `job.document()` instead of `job.doc`.
-            For more information, see :attr:`Job.statepoint` or
-            :class:`~signac.JSONDict`.
-        """
```
mikemhenry, line 269: We should keep this warning IMHO

signac/contrib/job.py
```diff
@@ -215,20 +267,32 @@ def statepoint(self):
...
         return self._statepoint
 
     @statepoint.setter
     def statepoint(self, new_sp):
+        """Assigns new state point to this Job.
+
+        Parameters
+        ----------
+        new_sp :
+            new state point to be assign.
```
mikemhenry, line 284:
```suggestion
            new state point to be assigned.
```

signac/contrib/job.py
ac-optimus: I think this got deleted by mistake.

signac/contrib/job.py
ac-optimus: thanks for the details, I have added the doc string accordingly.

signac/contrib/job.py
```diff
@@ -405,6 +523,10 @@ def clear(self):
 
         This function will do nothing if the job was not previously
         initialized.
+
+        Raises
+        ------
+
```
ac-optimus, line 529: @mikemhenry @tommy-waltmann @vyasr what should I put in this?

signac/contrib/job.py
```diff
@@ -382,15 +495,20 @@ def init(self, force=False):
...
+        Returns
+        -------
+        class : `~.Job`
             The job handle.
-        :rtype:
-            :class:`~.Job`
+
+        Raises
+        ------
+
```
ac-optimus, line 511: @mikemhenry @tommy-waltmann @vyasr what should I put in this?

signac/contrib/job.py
```diff
@@ -363,7 +471,12 @@ def _init(self, force=False):
             self._check_manifest()
 
     def _check_manifest(self):
-        "Check whether the manifest file, if it exists, is correct."
+        """Check whether the manifest file, if it exists, is correct.
+
+        Raises
+        ------
+
```
ac-optimus, line 478: @mikemhenry @tommy-waltmann @vyasr what should I put in this?

signac/contrib/job.py
```diff
@@ -321,16 +407,38 @@ def data(self):
...
+        """Helper function to initialize the job's workspace directory.
+
+        Parameters
+        ----------
+        force :
+            (Default value = False)
+
+        Raises
+        ------
+
```
ac-optimus, line 440: @mikemhenry @tommy-waltmann @vyasr what should I put in this?

signac/contrib/job.py
ac-optimus, line 529: @mikemhenry @tommy-waltmann @vyasr what should I put in this?

vyasr:
@ac-optimus if you don't see any exceptions raised inside a method, you can remove the `Raises` section of the documentation.

ac-optimus:
> @ac-optimus if you don't see any exceptions raised inside a method, you can remove the `Raises` section of the documentation.

@vyasr yes sure, actually my comments were related to the exceptions that are raised and I am confused with what to write in their descriptions.

vyasr:
Oh sorry @ac-optimus that's my fault, I didn't realize that all of these functions were actually raising exceptions. Although you're asking about specific functions, this issue raises an important question about exactly what exceptions we should document. The [numpydoc guide](https://numpydoc.readthedocs.io/en/latest/format.html) suggests that the `Raises` "section should be used judiciously, i.e., only for errors that are non-obvious or have a large chance of getting raised." I interpret that as also meaning that we should not document exceptions that we are not raising ourselves, i.e. don't add `Exceptions` to the `Raises` section if they're being reraised after catching them. So any function where we do something like `except Exception as e:... raise e`, I would just skip those (which I think is most of the ones you asked about). Does that make sense?

ac-optimus:
> Oh sorry @ac-optimus that's my fault, I didn't realize that all of these functions were actually raising exceptions. Although you're asking about specific functions, this issue raises an important question about exactly what exceptions we should document. The [numpydoc guide](https://numpydoc.readthedocs.io/en/latest/format.html) suggests that the `Raises` "section should be used judiciously, i.e., only for errors that are non-obvious or have a large chance of getting raised." I interpret that as also meaning that we should not document exceptions that we are not raising ourselves, i.e. don't add `Exceptions` to the `Raises` section if they're being reraised after catching them. So any function where we do something like `except Exception as e:... raise e`, I would just skip those (which I think is most of the ones you asked about). Does that make sense?


Yes it does @vyasr . So according to that I should remove all the `Raises` docstrings that I have added other than the below ones?

```
def update_statepoint(self, update, overwrite=False):
    ...
    ...   
    ...
        Raises
        ------
        KeyError 
            If the update contains keys, which are already part of the job's
            state point and overwrite is False.
```

and 
```
def sync(self, other, strategy=None, exclude=None, doc_sync=None, **kwargs):
    ...
    ...   
    ...
        Raises
        ------
        FileSyncConflict
            In case that a file synchronization results in a conflict.
```

bdice:
I took a quick glance at this PR. It seems like some docstrings might be indicating the wrong behavior or have the wrong types (e.g. iterables of jobs vs. a single job). Also I saw some rST errors with `.. warning::` that will prevent the docs from rendering properly, as well as spelling/grammar errors. I am working on a few other documentation pull requests like #318, #320, and #322. @mikemhenry could you take another look? Or maybe @vyasr since you've commented on this PR as well?

signac/contrib/job.py
```diff
@@ -535,24 +639,39 @@ def sync(self, other, strategy=None, exclude=None, doc_sync=None, **kwargs):
...
+            True if file with filename exists in workspace.
+
+        """
         return os.path.isfile(self.fn(filename))
 
     def open(self):
         """Enter the job's workspace directory.
 
-        You can use the :class:`~.Job` class as context manager:
+        You can use the : `Job` class as context manager:
```
vyasr, line 674:
```suggestion
        You can use the `Job` class as context manager:
```

ac-optimus:
> I have a number of changes, but you should be able to just batch commit all my suggestions and then I think this is good. The only thing I'm somewhat uncertain about is how things like referring to `Job` will work with numpydoc. I've looked online for examples because numpydoc's documentation doesn't say anything about it, and it seems like it should properly link with doing the `:class:`-type domain prefix, but I don't know if there are any limitations to this. @ac-optimus it would be great if you could try and build this locally once to just see if the links work so that we know for future PRs.

sure, working on it...

ac-optimus:
@tommy-waltmann I think I have addressed them all. Any specific one, just in case I missed it?

ac-optimus:
> > @tommy-waltmann I think I have addressed them all. Any specific one, just in case I missed it?
> 
> I was referring to trying to build the documentation locally to see if numpydoc works as we expect. Otherwise, I think this PR is in good shape.

Yes, I tried that and even had conversation with @vyasr as well on slack. This is working fine!

PR #323, opened 2020-04-08
Added groupby for variable that are not present in entire project

### LINKED ISSUE #321: groupby() by variables that not present in the entire data space
### Feature description

I am interested in obtaining all the jobs that contain a variable `b` which does not necessarily have to be present in all the jobs.

### Proposed solution
From @bdice in the slack group:

> Bradley Dice  10:43 AM
@Miguel This behavior could be improved, but what would you expect the behavior to be? A group where b is None isn't really an option because b could be set to None in the statepoint. ***We could return only groups where the value is set (and ignore jobs where the key is missing).***

### Additional context
Example of the current problem. First `groupby()` works, second does not.
```python
import signac

sp = { 'a' : 1.0,
       'c' : 2.0
        }
job = project.open_job(sp)
job.init()

sp = { 'a' : 2.0,
       'b' : 1.0,
        }
job = project.open_job(sp)
job.init()

sp = { 'a' : 1.0,
       'b' : 3.0,
        }
job = project.open_job(sp)
job.init()

for a, group in project.groupby('a'):
    print(a, list(group))

for b, group in project.groupby('b'):
    print(b, list(group))
```

vishav1771:
## Motivation and Context

Resolves #321.
## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -681,6 +681,15 @@ class Project(object):
             A default value to be used when a given state point key is not present (must
             be sortable).
         """
+        if default is None:
+            if isinstance(key, str):
+                _filter = {key: {"$exists": True}}
+            elif isinstance(key, Iterable):
+                _filter = {k: {"$exists": True} for k in key}
+            else:
+                _filter = None
+            return self.find_jobs(filter=_filter).groupby(key, default=default)
+
         return self.find_jobs().groupby(key, default=default)
 
     def groupbydoc(self, key=None, default=None):
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -1768,10 +1768,13 @@ class JobsCursor(object):
             A default value to be used when a given state point key is not present (must
             be sortable).
         """
-        _filter = None
+        _filter = self._filter
         if isinstance(key, str):
             if default is None:
-                _filter = {key: {"$exists": True}}
+                if _filter is None:
+                    _filter = {key: {"$exists": True}}
+                elif key not in _filter.keys():
+                    _filter[key] = {"$exists": True}
 
                 def keyfunction(job):
                     return job.sp[key]
@@ -1781,7 +1784,12 @@ class JobsCursor(object):
 
         elif isinstance(key, Iterable):
             if default is None:
-                _filter = {k: {"$exists": True} for k in key}
+                if _filter is None:
+                    _filter = {k: {"$exists": True} for k in key}
+                else:
+                    for k in key:
+                        if k not in _filter.keys()
+                    _filter[k] = {"$exists": True}
 
                 def keyfunction(job):
                     return tuple(job.sp[k] for k in key)
@@ -1797,8 +1805,8 @@ class JobsCursor(object):
         else:
             keyfunction = key
 
-        return groupby(sorted(iter(JobsCursor(self._project, _filter, None)), key=keyfunction),
-                       key=keyfunction)
+        return groupby(sorted(iter(JobsCursor(self._project, _filter, self._doc_filter)),
+                              key=keyfunction), key=keyfunction)
 
     def groupbydoc(self, key=None, default=None):
         """Groups jobs according to one or more document values.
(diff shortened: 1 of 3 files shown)


signac/contrib/project.py
```diff
@@ -681,6 +681,15 @@ def groupby(self, key=None, default=None):
...
             be sortable).
         """
+        if default is None:
+            if isinstance(key, str):
+                _filter = {key: {"$exists": True}}
+            elif isinstance(key, Iterable):
+                _filter = {k: {"$exists": True} for k in key}
+            else:
+                _filter = None
+            return self.find_jobs(filter=_filter).groupby(key, default=default)
```
csadorf, line 691: I'm wondering, would it make more sense to implement this logic directly within the `groupby()` implementation of the `JobsCursor` class? So that it would also work when users call this directly on a result of `find_jobs()` ?

bdice: Yes, that's a good idea. We try to keep as much logic in `JobsCursor` as possible, so that it's useful for operations on subsets of the project.

signac/contrib/project.py
```diff
@@ -1801,7 +1803,8 @@ def keyfunction(job):
         else:
             keyfunction = key
 
-        return groupby(sorted(iter(self), key=keyfunction), key=keyfunction)
+        return groupby(sorted(iter(JobsCursor(self._project, _filter, self._doc_filter)),
```
csadorf, line 1806: Why do we need to instantiate "self" again here?

vishav1771: I need to use a different filter than `self.filter` (to remove Jobs which don't have that key). Instead of modifying the `self.filter`, I choose to use a different instance to

zhou-pj:
LGTM！

zhou-pj:
No necessary directly related to this PR, but I wonder if the following part in docstring (from line 1753) are still supported and tested in the code? I'm not sure I understand how `_doc_filter` is used as groupby key in this function.
```
#Group by job.sp['d'] and job.document['count'] using a lambda.
      for key, group in project.groupby(
                lambda job: (job.sp['d'], job.document['count'])
            ):
                print(key, list(group))```

vyasr:
@zhou-pj that's a good question. The API here is actually a little bit strange. Right now, if you pass a string or group of strings to `groupby` or `groupbydoc`, they will automatically construct functions for grouping jobs by statepoint or document keys, respectively. However, you can also pass a callable. If you pass a callable to `groupby`, then it allows grouping by any job parameters because the calllable is a function of the job. If you pass a callable to `groupbydoc`, on the other hand, the callable must be a function of the _job document_. @bdice I believe you were responsible for the initial implementation of this feature, do you recall why we decided to split the grouping functionality into two functions that have this slightly divergent behavior?

bdice:
@vyasr @zhou-pj I looked at the [original pull request](https://bitbucket.org/glotzer/signac/pull-requests/26/groupby/diff) where I introduced this feature. I don't recall discussing this API choice explicitly. I think it was an arbitrary choice that allowed `groupby` to be very general (a callable of `job` is more flexible than a callable of `job.sp`) while allowing `groupbydoc` to have `job.doc` as the argument for convenience. The most "parallel" solution would be to have `groupby` operate on `job.sp` in all cases but it would not be useful for cases where you want to group by something like `job.isfile('output.txt')`.

vyasr:
In the interest of keeping this PR focused, I'm going to move this discussion to a separate issue.

vyasr:
@lyrivera yes, codecov changes are pretty noisy and frequently fail showing negligible changes in coverage i.e. -0.05%. That number is reported as part of the check, so you can do a sanity check that it isn't actually dropping by >1% or something (which would be a more plausible indication that lots of untested code was added).

@lyrivera @zhou-pj if you're happy, can you actually approve the PR?

@csadorf you had some comments when reviewing this PR, so I'd prefer not to have this merged without your approval. Please let us know if you're happy with it now.

signac/contrib/project.py
csadorf: Thx.

PR #325, opened 2020-04-10
Show a palette emoji 🎨 in signac shell's version info.

bdice:
## Description
This PR adds a palette emoji 🎨 in the version info in `signac shell`. It's _perfectly fine_ if we choose to reject this for any reason. I don't think this will cause any problems since Python 3 supports Unicode and this is just a Unicode character. I am not putting the emoji in the output of `signac --version` so that the version can be easily parsed from that output string if needed. There are so many packages that use emoji in their git commit messages and other parts of the software so I do not anticipate any issues with locales/etc.

## Motivation and Context
Just for fun. I've been thinking about this for a while. 😄

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -1818,7 +1818,7 @@ This feature is still experimental and may be removed in future versions.""")
     # allow to parse only --version without any
     # of the other required arguments.
     if '--version' in sys.argv:
-        print('signac', __version__, '🎨')
+        print('signac', __version__)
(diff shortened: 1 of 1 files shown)


csadorf:
Not a reviewer, but just wanted to say that I love this. :)

PR #330, opened 2020-04-28
Add usecols and flatten to to_dataframe feature.

### LINKED ISSUE #327: Request subset of columns in `JobsCursor.to_dataframe`
### Feature description

My statepoints are often overspecified in the sense that I include more information than is required to make a job unique in a data space because those are still variables that are critical to defining the data point and I anticipate the possibility of varying them. It would be very convenient if there was a simple API to specify which columns should be included in the DataFrame (which would essentially be equivalent to running `df.drop(EXCLUDED_COLUMNS, inplace=True)` after the face), and more importantly to replicate the behavior in `project.detect_schema` where you can `exclude_const`.

### Proposed solution

Add `exclude_const` and/or `exclude_keys` extra arguments to `JobsCursor.to_dataframe` to specify statepoint or document parameters that should not be included in the output dataframe.

bdice:
## Description
Add feature for including/excluding state point and document keys from an exported pandas DataFrame.

## Motivation and Context
Resolves #327.

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -10,7 +10,7 @@ import errno
 import uuid
 import gzip
 import time
-from collections.abc import Container, Iterable
+from collections.abc import Iterable
 from contextlib import contextmanager
 from deprecation import deprecated
 from itertools import groupby
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -1902,30 +1902,23 @@ class JobsCursor(object):
         """
         import pandas
 
-        class _IncludeExcludeContainer(Container):
-            """An abstract container that includes/excludes user-provided lists."""
-
-            def __init__(self, includes=None, excludes=None):
-                self.includes = set(includes) if includes is not None else None
-                self.excludes = set(excludes) if excludes is not None else None
-
-            def __contains__(self, item):
-                if self.excludes is not None and item in self.excludes:
-                    return False
-                elif self.includes is not None and item not in self.includes:
-                    return False
-                else:
-                    return True
-
-        sp_valid_keys = _IncludeExcludeContainer(sp_includes, sp_excludes)
-        doc_valid_keys = _IncludeExcludeContainer(doc_includes, doc_excludes)
+        if sp_includes is not None:
+            sp_includes = set(sp_includes)
+        if sp_excludes is not None:
+            sp_excludes = set(sp_excludes)
+        if doc_includes is not None:
+            doc_includes = set(doc_includes)
+        if doc_excludes is not None:
+            doc_excludes = set(doc_excludes)
 
         def _export_sp_and_doc(job):
             for key, value in job.sp.items():
-                if key in sp_valid_keys:
+                if (sp_includes is None or key in sp_includes) and \
+                        (sp_excludes is None or key not in sp_excludes):
                     yield sp_prefix + key, value
             for key, value in job.doc.items():
-                if key in doc_valid_keys:
+                if (doc_includes is None or key in doc_includes) and \
+                        (doc_excludes is None or key not in doc_excludes):
                     yield doc_prefix + key, value
 
         return pandas.DataFrame.from_dict(
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -1878,68 +1878,63 @@ class JobsCursor(object):
         return dict(export_jobs(jobs=list(self), target=target,
                                 path=path, copytree=copytree))
 
-    def to_dataframe(self, sp_prefix='sp.', doc_prefix='doc.', sp_includes=None,
-                     sp_excludes=None, doc_includes=None, doc_excludes=None):
+    def to_dataframe(self, sp_prefix='sp.', doc_prefix='doc.', usecols=None,
+                     flatten=False):
         """Convert the selection of jobs to a pandas dataframe.
 
-        This function exports the job metadata to a :py:class:`pandas.DataFrame`.
-        All state point and document keys are prefixed by default to be able to distinguish them.
+        This function exports the job metadata to a
+        :py:class:`pandas.DataFrame`. All state point and document keys are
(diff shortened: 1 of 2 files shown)


bdice:
@vyasr This is what I had in mind for solving #327. Let me know what you think.

signac/contrib/project.py
```diff
@@ -1901,11 +1902,31 @@ def to_dataframe(self, sp_prefix='sp.', doc_prefix='doc.'):
         """
         import pandas
 
+        class _IncludeExcludeContainer(Container):
```
csadorf, line 1905: Using a dedicated class -- even if internal -- here seems a little bit like overkill to me. :/

Is a simple check as in `(whitelist is None or key in whitelist) and (blacklist is None or key not in blacklist)` not sufficient?

bdice: I had that at first but decided this might be preferred because that logic is repeated for sp and doc keys, _and_ the inputs need to be converted to sets (the inputs could be iterables). It ends up being around the same number of lines total but I can change it if you're opposed to this.

bdice: Done.

bdice:
@vyasr I think adding this feature is in scope. I disagree about supporting only whitelists or only blacklists. I don't think that "constructing one from another should be easy." The one-liner that you wrote is pretty complicated and I think this is a reasonable convenience feature for users. I tried to address the potential for ambiguity on how the two interact by explicitly stating that "excludes override includes" in the docs.

vyasr:
Since we're dealing with pandas here, we should use pandas as a guide. All of its `read_*` functions provide a `usecols` argument, which is a whitelist. However, this argument is ducktyped to support specifying integer columns, or in the case of wanting a blacklist, a callable that could basically be `lambda key: key not in UNDESIRED_KEYS`. IMO that would be a cleaner solution for producing an exclusion list. I saw your documentation and I think it is good, I just think a less confusing API choice that doesn't require reading the documentation is greatly preferable if possible.

I'm a bit confused since you were originally of the viewpoint that this was not in scope for signac. Was it was my argument about large files and performance that changed your mind? If so, since Simon was also originally against this I'd be curious what his thoughts are now.

bdice:
@vyasr I like your idea to mirror the `usecols` syntax of [read_csv](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html).

You're correct, I changed my mind about scope after considering the poor performance that would result from adding unwanted job document fields. Job documents are frequently large and there's no need to copy all of that into the dataframe if it's not requested.

csadorf:
Also in favor if mirroring the pandas API as much as reasonable/possible. 👍

vyasr:
@bdice I'm taking @csadorf's last comment as tacit approval of moving forward. Since the two of you were the ones who were originally against I think we're good to move forward. Once you make the `usecols` change I can review again.

On the issue, you also raised the question of `exclude_const` and nested statepoints. Now that I'm thinking about it, is there a good reason not to unpack nested statepoints? Obviously with heterogeneous schema many things could end up being null values, but I think that's OK since the user is asking for a tabular representation. What if we added an option `unpack_nested`, and then have `exclude_const` always imply `unpack_nested=True`?

csadorf:
Also in favor of having the option to automatically _flatten_. Working with data that has more than two dimensions in data frames is an absolute pain.

bdice:
@vyasr @csadorf This is ready for another round of review. I changed to the `usecols` syntax and added a `flatten` option for nested dicts.

signac/contrib/project.py
```diff
@@ -1878,32 +1878,66 @@ def export_to(self, target, path=None, copytree=None):
...
+            True. For example, ``lambda x: 'sp.' in x``.
+        :type usecols:
+            list-like or callable, optional
+        :param flatten:
+            Whether nested state points or document keys should be flattened.
+            If True, ``{'a': {'b': 'c'}}`` becomes a column named ``a.b`` with
+            value ``c``. If False, it becomes a column named ``a`` with value
+            ``{'b': 'c'}``.
+        :type flatten:
+            bool, optional
```
csadorf, line 1912: Maybe document the default here?

bdice:
@vyasr You can merge this when approved. 👍

signac/contrib/project.py
```diff
@@ -1878,32 +1878,67 @@ def export_to(self, target, path=None, copytree=None):
...
+        :param flatten:
+            Whether nested state points or document keys should be flattened.
+            If True, ``{'a': {'b': 'c'}}`` becomes a column named ``a.b`` with
+            value ``c``. If False, it becomes a column named ``a`` with value
+            ``{'b': 'c'}``. Defaults to ``False``.
+        :type flatten:
+            bool, optional
         :returns:
-            A pandas dataframe with all job metadata.
+            A pandas DataFrame with all job metadata.
```
vyasr, line 1915:
```suggestion
            A :class:`pandas.DataFrame` with all job metadata.
```

signac/contrib/project.py
```diff
@@ -1878,32 +1878,67 @@ def export_to(self, target, path=None, copytree=None):
...
         :type doc_prefix:
-            str
+            str, optional
+        :param usecols:
+            Used to select a subset of columns. If list-like, must contain
+            strings corresponding to the column names that should be included.
+            For example, ``['sp.a', 'doc.notes']``. If callable, the column
+            will be included if the function called on the column name returns
+            True. For example, ``lambda x: 'sp.' in x``. Defaults to ``None``,
+            which uses all columns from the state point and document.
```
vyasr, line 1904:
```suggestion
            True. For example, ``lambda x: 'sp.' in x``. Defaults to ``None``,
            which uses all columns from the state point and document. Note
            that this filter is applied *after* the doc and sp prefixes are
            added to the column names.
```

PR #331, opened 2020-04-29
Use mongodb as optional index for statepoints and job docs

jglaser:
## Description

Use `mongodb` as an optional backend (*via* pymongo). Store state points and job docs in the database instead of creating a directory for every statepoint. Local workspaces are still created by calling `job.open()` explicitly. Jobs are added to the database only with `job.add()`.

To enable the database backend, configure a database host (`myhost`) as described [here](https://docs.signac.io/en/latest/configuration.html#host-configuration) and put these variables into the configuration (`signac.rc`):

```
index_host = myhost
index_db = signac
```

## Motivation and Context
We aim to overcome a scalability issue in signac for large indices (>100,000), when the file system limits the number of subdirectories, i.e. state points, to be created. By decoupling state points from local workspace directories, distributed workers could e.g. create those directories on node-local flash memory as needed for a subset of the entire workspace, while connecting to the central database to store and retrieve job information.

A reference implementation using `mongodb` as a persistent backend on an OLCF OpenShift cluster is underway (actually, figuring out how to allow network access to the cluster took most of the time, rather than implementing these changes :-), which follow ideas by @csadorf.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -180,14 +180,16 @@ class Project(object):
             'statepoint_cache_miss_warning_threshold', 500)
 
         self.db = None
-        hostname = config['index_host']
+        hostname = None
+        if 'index_host' in config:
+            hostname = config['index_host']
         if hostname is not None:
-            db_name = config['index_db']
-            if db_name is None:
+            if 'index_db' not in config:
                 raise ConfigError("When using a db host for the index, specify " +
                                   "the database, too.")
 
             # open the database connection
+            db_name = config['index_db']
             self.db = database.get_database(db_name, hostname=hostname)
 
             # 'index' is the default name for the collection
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -46,7 +46,7 @@ JOB_ID_REGEX = re.compile('[a-f0-9]{32}')
 ACCESS_MODULE_MINIMAL = """import signac
 
 def get_indexes(root):
-    yield signac.get_project(root).index()
+    yield signac.get_project(root).index_from_workspace()
 """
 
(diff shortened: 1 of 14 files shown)


jglaser:
An obvious bottleneck is now the state point search, which loads all statepoints into the cache first before filtering. We will want to map that onto a `mongodb` query

- [ ] use mongodb query API

signac/contrib/project.py
```diff
@@ -1397,8 +1438,8 @@ def _read_cache(self):
             logger.debug("Read cache in {:.3f} seconds.".format(delta))
             return cache
 
-    def index(self, formats=None, depth=0,
-              skip_errors=False, include_job_document=True):
+    def index_from_workspace(self, formats=None, depth=0,
+                             skip_errors=False, include_job_document=True):
```
csadorf, line 1442: I think we should modify the `index()` method such that it calls both `index_from_workspace()` and `index_from_db()` and make the latter two private methods.

csadorf (changes requested):

signac/contrib/project.py
jglaser: That might be a good option. Are you proposing to use the union of the two sets of statepoints then? There could be statepoints that are only local, statepoints that are in the db, and state points that are in both.. need to make the latter unique.

b-butler:
This is handled in a current GSoC project. A MongoDB is one of the many backends for signac synced collections in #364. When the refactoring of data structures is done, adding support for different backends to job statepoints is planned.

PR #332, opened 2020-04-29
Implement feature to search jobs with a sp- and doc-integrated filter 

vishav1771:
## Description
 This patch enables the search for jobs with a statepoint- and document-
 combined filter. To specify whether a key is a statepoint- or a
 document-key use prefixes 'sp.' and 'doc.' respectively. No prefix is
 equivalent to the 'sp.' prefix.
 
 This patch is backwards compatible with the exception that the index
 scheme was slightly modified.
 
 Related to issue [glotzerlab/signac-dashboard#25](https://github.com/glotzerlab/signac-dashboard/issues/25) .

 ## Minimal example
 This patch enables queries such as this one:
 
```python
 >>> project.find_jobs({'foo': 0, 'doc.bar': True})
 signac.contrib.project.JobsCursor({'project': 'project', 'filter': '{'sp.foo': 0, 'doc.bar': True}'})
 >>> # equivalent to:
 >>> project.find_jobs({'foo': 0}, {'bar': True})
 signac.contrib.project.JobsCursor({'project': 'project', 'filter': '{'sp.foo': 0, 'doc.bar': True}'})
 ```
 
 Equivalently on the command line:
 
 ```shell
 $ signac find foo 0 doc.bar true
 ```
 
All keys without a prefix are automatically interpreted as having the `sp.` prefix.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### CODE REVIEW

**Comment by vishav1771** [comment]
@csadorf  @vyasr  I have updated the PR according to the master. 
I have some doubts about the warning. I think it should warn while setting the root key as `doc` in statepoint and `sp` in document. Please guide.

**Comment by vyasr** [comment]
@vishav1771 could you clarify what you mean? What warning are you referring to?

**Comment by csadorf** [comment]
@vishav1771 I'm also not sure what you are referring to, could you clarify, please?

**Comment by vishav1771** [comment]
Sorry for the misunderstanding. Actually, I was referring to the task defined in the PR #188 
>Add warning about the use of `sp` and `doc` as root-level keys in documents.

**Comment by csadorf** [comment]
> #188

The reasoning was that users should generally avoid those keys in both cases, because otherwise it is difficult to know what they are actually referring to. I don't think it's a huge issue, because if you use `doc` or `sp` as a key in either one of them, you could always refer to them as `sp.doc: ..` or `sp.sp: ...` etc. So it's more of an issue of creating misleading structures, but I think the best thing is to probably just ignore this for now.

**Comment by vishav1771** [comment]
Okay, I understand 👍 .

File: tests/test_project.py [hunk 1]
```diff
@@ -924,8 +924,18 @@ def get_sp(i):
...
+            return {
+                'a': i,
+                'b': i % 2,
+                'c': i % 3
+            }
+
         for i in range(12):
-            self.project.open_job(get_sp(i)).init()
+            job = self.project.open_job(get_sp(i))
+            job.init()
```
csadorf, line 937: Proposal: `job = self.project.open_job(get_sp(i)).init()`

vishav1771:
I've changed all the functions, but still, if there's anything missing please do tell!

tests/test_shell.py
```diff
@@ -142,6 +142,7 @@ def test_statepoint(self):
         assert "{'a': 0}" in sp
         assert len(project) == 1
 
+    @pytest.mark.xfail()
```
csadorf, line 145: Why is this an expected failure?

tests/test_project.py
```diff
@@ -861,6 +970,51 @@ def get_sp(i):
...
         assert group_count == len(list(self.project.find_jobs()))
 
+        # using sp.key and doc.key
+        for k, g in self.project.groupby('sp.a'):
+            assert len(list(g)) == 1
+            for job in list(g):
+                assert job.sp['a'] == k
+
+        assert len(list(self.project.groupby('d'))) == 0
+        for k, g in self.project.groupby('sp.d', default=-1):
```
csadorf, line 980: It is not clear to me how these two lines relate to each other. Is the for-loop even entered at all if the length of that grouping is zero?

tests/test_project.py
```diff
@@ -334,7 +355,16 @@ def test_find_jobs_arithmetic_operators(self):
 
     def test_find_jobs_logical_operators(self):
         for i in range(10):
-            self.project.open_job({'a': i, 'b': {'c': i}}).init()
+            job = self.project.open_job({'a': i, 'b': {'c': i}})
+            job.init()
```
csadorf, line 359:
```suggestion
            job = self.project.open_job({'a': i, 'b': {'c': i}}).init()
```

tests/test_project.py
```diff
@@ -368,6 +398,74 @@ def test_find_jobs_logical_operators(self):
...
+        q = {'$or': [{'$or': [{'sp.b.c': 0}, {'sp.b.c': 1}]}]}
+        assert len(self.project.find_jobs(q)) == 2
+        q = {'$or': [{'$and': [{'sp.b': {'c': 0}}, {'sp.b': {'c': 1}}]}]}
+        assert len(self.project.find_jobs(q)) == 0
+        q = {'$or': [{'$and': [{'sp.b.c': 0}, {'sp.b.c': 1}]}]}
+        assert len(self.project.find_jobs(q)) == 0
+
+        # Mixed filters
+        def assert_result_len(q, num):
+            assert len(self.project.find_jobs(q)) == num
```
csadorf, line 449: I like this. Maybe use that also for the tests above?

tests/test_project.py
```diff
@@ -825,8 +924,18 @@ def get_sp(i):
...
+            return {
+                'a': i,
+                'b': i % 2,
+                'c': i % 3
+            }
+
         for i in range(12):
-            self.project.open_job(get_sp(i)).init()
+            job = self.project.open_job(get_sp(i))
+            job.init()
```
csadorf, line 937:
```suggestion
            job = self.project.open_job(get_sp(i)).init()
```

tests/test_shell.py
```diff
@@ -247,13 +248,20 @@ def test_find(self):
 
         # Test the doc_filter
         for job in project.find_jobs():
-            job.document['a'] = job.statepoint()['a']
+            job.document['a'] = job.statepoint()['a'] + 1
+
         with pytest.deprecated_call():
-            for i in range(3):
+            for i in range(1, 4):
```
csadorf, line 254: What is the motivation for this change?

tests/test_shell.py
vishav1771: This test is for index. As index schema is modified, the test is expected to fail.

tests/test_project.py
vishav1771: As `d` is not present in jobs, `self.project.groupby('sp.d', default=-1)` will return one group having key as `default` (which is -1).

tests/test_project.py
vishav1771: Sure 👍

tests/test_shell.py
vishav1771: I have changed the value `a` in doc and sp so that they are different for a particular job. So to test the key 'a' selected belong to sp or doc.

signac/contrib/filterparse.py
```diff
@@ -90,14 +90,59 @@ def parse_filter_arg(args, file=sys.stderr):
...
+
+
+def _parse_filter(filter):
+    if isinstance(filter, str):
+        yield from parse_simple(filter.split())
+    elif filter:
+        yield from filter.items()
+
+
+def parse_filter(filter, prefix='sp'):
```
csadorf, line 147: Shouldn't the prefix be `sp.`?

signac/contrib/project.py
```diff
@@ -658,6 +668,14 @@ def groupby(self, key=None, default=None):
...
                 print(key, list(group))
 
+            # Group jobs by document value 'a'.
+            for key, group in project.groupby('doc.a'):
+                print(key, list(group))
+
+            # Group jobs by jobs.sp['a'] and job.document['b']
+            for key, group in project.groupby('a', 'doc.b'):
+                print(key, list(group))
+
```
csadorf, line 678: This looks like it's left-over from debugging.

signac/contrib/project.py
```diff
@@ -1753,6 +1769,14 @@ def groupby(self, key=None, default=None):
...
```
csadorf, line 1779: Debugging?

tests/test_project.py
```diff
@@ -565,6 +663,7 @@ def test_index(self):
         assert len(docs) == 2 * len(statepoints)
         assert len(set((doc['_id'] for doc in docs))) == len(docs)
 
+    @pytest.mark.xfail()
```
csadorf, line 666: I think it would be good to add a small comment for all the expected failures describing the reason. Those represent action items.

tests/test_shell.py
csadorf: I'd prefer to just add a new test that looks at that specific behavior. Whenever we modify existing tests, we increase the risk of introducing unintended backwards incompatible changes. It is absolutely ok if we largely duplicate existing tests. Duplication of test code is much preferable over modifying existing tests.

tests/test_project.py
csadorf: Got it, thx.

signac/contrib/filterparse.py
vishav1771: We are adding '.' in `_add_prefix`
https://github.com/glotzerlab/signac/blob/007afcd3033b6e0e150b7309f849b666aa424b46/signac/contrib/filterparse.py#L124

signac/contrib/project.py
vishav1771: I updated the doc-string of the `groupby` function and added examples showing how to use doc key.

signac/contrib/project.py
vishav1771: Same as above.

tests/test_shell.py
vishav1771: Okay, Got it 👍

tests/test_project.py
vishav1771: Done

vishav1771: `self.project.open_job(get_sp(i)).init()` return `None`. So I am reverting the change.

tests/test_project.py
vishav1771: I tried it. But,`self.project.open_job(get_sp(i)).init()` returns `None`.

signac/contrib/filterparse.py
csadorf: But that could be changed, right? It seems to me that not hard-coding the added dot makes the function way more flexible without losing functionality.

tests/test_project.py
csadorf: That should absolutely not be the case:
https://github.com/glotzerlab/signac/blob/cbce7ccace9cc006157845d1128b0619bfac23e9/signac/contrib/job.py#L401

signac/contrib/project.py
csadorf: Got it. That was hard to see in the GitHub diff view.

signac/contrib/filterparse.py
vishav1771: Sure, will do that.

tests/test_project.py
vishav1771: I got the following error.
error

tests/test_project.py
vishav1771: Done 👍

tests/test_project.py
vishav1771: Done 👍

signac/contrib/project.py
```diff
@@ -618,9 +618,9 @@ def _find_job_ids(self, filter=None, doc_filter=None, index=None):
         if filter is None and doc_filter is None and index is None:
             return list(self._job_dirs())
         if index is None:
-            filter = dict(parse_filter(filter, 'sp'))
+            filter = dict(parse_filter(filter, 'sp.'))
```
csadorf, line 621: I think we should use the explicit keyword argument here, because otherwise it is hard to follow without knowing the function signature of `parse_filter()`.
```suggestion
            filter = dict(parse_filter(filter, prefix='sp.'))
```
I'd actually go even a step further and use an even more explicit code structure:
```suggestion
            filter = dict(parse_filter(add_prefix('sp.', filter)))
```
Here I also switched the argument order for `add_prefix()`, because that would logically make more sense to me.

_Any changes proposed above would of course also apply to the other spots in the code with the same pattern._

tests/test_project.py
csadorf: @vishav1771 Could you please investigate the reason for this? Are we wrapping the `Project` class in the tests implementation or so? As I pointed out, obviously `init()` must return an instance of `Job`. If it doesn't something is very wrong.

vishav1771: Sure, will look into that 👍

vishav1771: Done.

vishav1771:
@csadorf I have done the changes.

vyasr:
@csadorf sorry I must have completely missed this message when you first sent it. Should we convert this PR from a draft and assign reviewers so that we can restart the discussion? If you would like to start the discussion with just committers prior to that, I can start a conversation in Slack.

csadorf:
> @csadorf sorry I must have completely missed this message when you first sent it. Should we convert this PR from a draft and assign reviewers so that we can restart the discussion? If you would like to start the discussion with just committers prior to that, I can start a conversation in Slack.

Let's start a conversation on slack.

vishav1771:
Next steps to resolve this incompatibility issue:
The following distinct patches each in a separate  PR but on the same feature branch will be introduces (as suggested by @csadorf ):
- [ ] Introduction of an explicit schema version to the index (starting with “0”). This should be straightforward and backwards compatible.

- [ ] Change of the schema to the new format. Backwards incompatible.

- [ ] An attempt to provide a migration, similar to how this is done [here](https://github.com/glotzerlab/signac/tree/master/signac/contrib/migration) If that turns out to be a non-trivial undertaking, we just drop that.

vyasr:
@glotzerlab/signac-committers I'd like to revisit this PR prior to signac 2.0. We have deprecated most APIs in signac that allow users to pass an index. In my opinion, we should simply deprecate the remainder since users rarely interact with index objects. If we do that, then I think that making this breaking change to the indexing scheme in signac 2.0 without offering migration is acceptable. While introducing an appropriate versioning and attempting migration would be nice, I think it's very low priority and not worth delaying this PR for since it would be a very nice feature to add.

csadorf:
> For clarification, the change to the underlying index is just a change in keys from statepoint->sp. I don't recall why (or if) this change is 100% necessary, other than for brevity. @csadorf may remember since he was much more active in reviewing this PR.

The change is necessary to avoid internal mapping and to maintain consistency I believe. But fine to go ahead with that in 2.0 without migration IMO.

signac/contrib/filterparse.py
```diff
@@ -182,18 +182,41 @@ def _parse_simple(key, value=None):
...
+    tuple
+        A single key-value pair of input tokenized filter.
+
+    """
+    for i in range(0, len(tokens), 2):
+        key = tokens[i]
+        if i + 1 < len(tokens):
+            value = tokens[i + 1]
+        else:
+            value = None
```
mikemhenry, line 218: Can you add a test for this?

signac/contrib/filterparse.py
```diff
@@ -218,14 +241,68 @@ def parse_filter_arg(args, file=sys.stderr):
         if _is_json(args[0]):
             return _parse_json(args[0])
         else:
-            return _with_message(_parse_simple(args[0]), file)
+            key, value = _parse_single(args[0])
+            return _with_message({key: value}, file)
```
mikemhenry, line 245: Can you add a test for this case?

signac/contrib/filterparse.py
```diff
@@ -218,14 +241,68 @@ def parse_filter_arg(args, file=sys.stderr):
...
+
+def _add_prefix(prefix, filter):
+    if filter:
+        for key, value in filter.items():
+            if key in ("$and", "$or"):
+                if isinstance(value, list) or isinstance(value, tuple):
+                    yield key, [dict(_add_prefix(prefix, item)) for item in value]
+                else:
+                    raise ValueError(
+                        "The argument to a logical operator must be a sequence (e.g. a list)!"
```
mikemhenry, line 260:
```suggestion
                        "The argument to a logical operator must be a list or tuple!"
```
Since we are not following the normal [EAFP](https://docs.python.org/3.5/glossary.html#term-eafp) pattern here, we know what types we are looking for. Unless `isinstance` will accept like-like and tuple-like objects.

signac/contrib/filterparse.py
```diff
@@ -218,14 +241,68 @@ def parse_filter_arg(args, file=sys.stderr):
...
+
+    Yields
+    ------
+    tuple
+        A key value pair to be used as a filter.
+
+    """
+    filter = list(filter)
+
+    if isinstance(filter, str):
```
mikemhenry, line 303: Will the filter ever be a `str` since we cast it as a list in line 301?

signac/contrib/filterparse.py
```diff
@@ -218,14 +241,68 @@ def parse_filter_arg(args, file=sys.stderr):
...
+        yield from parse_simple(filter.split())
+    elif isinstance(filter, dict):
```
mikemhenry, line 305: Same question, will we ever have a dict since we cast it as a list on line 301?

signac/contrib/project.py
```diff
@@ -1811,7 +1834,7 @@ def repair(self, fn_statepoints=None, index=None, job_ids=None):
                 raise
         if index is not None:
             for doc in index:
-                self._sp_cache[doc["signac_id"]] = doc["statepoint"]
+                self._sp_cache[doc["signac_id"]] = doc["sp"]
```
mikemhenry, line 1837: Is it worth adding a test for this? Not sure how often `repair` is really used in the wild, but it would be nice if it was tested.

signac/contrib/project.py
```diff
@@ -2552,42 +2574,71 @@ def groupby(self, key=None, default=None):
                 else:
                     _filter = {"$and": [{key: {"$exists": True}}, _filter]}
 
-                def keyfunction(job):
-                    """Return job's state point value corresponding to the key.
+                if "." in key and key.split(".", 1)[0] == "doc":
```
mikemhenry, line 2577: Okay this next section feels a bit wild, lots of logic to determine what `keyfunction` to use, off the top of my head I can't think of a really easy way to refactor this to make the logic more simple, but could you add a few developer comments to guide the reader what is going on here? Some of the string manipulation happening here is a bit opaque.

mikemhenry (changes requested):
Looks pretty good, mostly just need a few more tests and I just had some minor points to improve clarity!

vyasr:
@mikemhenry is going to work on adding a few more of these tests.

signac/contrib/filterparse.py
vyasr: This is true. Maybe we should relax this... but I don't want to put in the effort to change this now, so I'm going to accept your suggestion and maybe we can improve this at a later point.

signac/contrib/filterparse.py
vyasr: Nope, good catch!

signac/contrib/project.py
vyasr: I think our plan for 2.0 is to remove most index-related logic, so I don't think we need to test this behavior.

signac/contrib/project.py
vyasr: Oh yeah, I remember hating this part of the merge. Let me see what I can do.

signac/contrib/filterparse.py
```diff
@@ -182,18 +182,41 @@ def _parse_simple(key, value=None):
...
     """
     if value is None or value == "!":
-        return {key: {"$exists": True}}
+        return key, {"$exists": True}
     elif _is_json(value):
-        return {key: _parse_json(value)}
+        return key, _parse_json(value)
     elif _is_regex(value):
-        return {key: {"$regex": value[1:-1]}}
+        return key, {"$regex": value[1:-1]}
```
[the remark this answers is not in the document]
vyasr: @mikemhenry if you could handle the regex test that would be great, I've done all the others.

signac/contrib/project.py
vyasr: Hopefully this logic is clearer now.

mikemhenry: Amazingly clearer!

mikemhenry (approved):

PR #343, opened 2020-06-29
Apply NumPy docstyle to remaining files in "signac" root.

bdice:
## Description
This PR applies NumPy docstyle to remaining files in the `signac` directory (i.e. the level of `__init__.py`).

## Motivation and Context
Progress on #315. After this, remaining PRs should remove `core`, `common`, and `contrib` from the "skipped" directories in `match-dir` in `setup.cfg`.

## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/sync.py
+++ b/signac/sync.py
@@ -102,35 +102,37 @@ __all__ = [
 # Definition of default sync strategies
 
 class FileSync(object):
-    "Collection of file synchronization strategies."
+    """Collection of file synchronization strategies."""
 
     @classmethod
     def keys(cls):
+        """Return keys."""
         return ('always', 'never', 'update', 'Ask')
 
     @staticmethod
     def always(src, dst, fn):
-        "Always overwrite files on conflict."
+        """Resolve sync conflicts by always overwriting."""
         return True
 
     @staticmethod
     def never(src, dst, fn):
-        "Never overwrite files on conflict."
+        """Resolve sync conflicts by never overwriting."""
         return False
 
     @staticmethod
     def update(src, dst, fn):
-        "Overwrite a file if the source file was modified last (based on timestamp)."
+        """Resolve sync conflicts based on newest modified timestamp."""
         return os.path.getmtime(src.fn(fn)) > os.path.getmtime(dst.fn(fn))
 
     class Ask(object):
-        "Ask whether a file should be overwritten interactively."
+        """Resolve sync conflicts by asking whether a file should be overwritten interactively."""
 
         def __init__(self):
             self.yes = set()
             self.no = set()
 
         def __call__(self, src, dst, fn):
+            """Ask user if a file should be overwritten."""
             if fn in self.yes:
                 return True
             elif fn in self.no:
@@ -146,7 +148,7 @@ class FileSync(object):
 
 
 class DocSync(object):
-    "Collection of document synchronization functions."
+    """Collection of document synchronization functions."""
 
     NO_SYNC = False
     "Do not synchronize documents."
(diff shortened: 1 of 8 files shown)


signac/sync.py
```diff
@@ -254,68 +259,58 @@ def sync_jobs(src, dst, strategy=None, exclude=None, doc_sync=None, recursive=Fa
...
+    dst : :class:`~signac.contrib.job.Job`
+        The dst job, data will be copied to this job's workspace.
+    strategy :
+        A synchronization strategy for file conflicts. If no strategy is
+        provided, a :class:`.errors.SyncConflict` exception will be raised upon
+        conflict. (Default value = None)
+    exclude : str
+        A filename exclusion pattern. All files matching this pattern will be
+        excluded from the synchronization process. (Default value = None)
+    doc_sync :
```
cbkerr, line 290: Do `doc_sync` and `strategy` need a type?

signac/sync.py
```diff
@@ -393,67 +388,68 @@ def sync_projects(source, destination, strategy=None, exclude=None, doc_sync=Non
...
-    :param check_schema:
+    strategy :
+        A file synchronization strategy. (Default value = None)
+    exclude :
+        Files with names matching the given pattern will be excluded from the
+        synchronization. (Default value = None)
+    doc_sync :
+        The function applied for synchronizing documents. (Default value =
+        None)
+    selection :
```
cbkerr, line 405: Do the optional arguments need types listed for their non-default values?

signac/syncutil.py
```diff
@@ -15,14 +20,30 @@
...
-    "Implementation adapted from https://docs.python.org/3/library/shutil.html#copytree-example'."
+    """Recursively copy a directory tree from src to dst, using a custom copy function.
+
+    Implementation adapted from https://docs.python.org/3/library/shutil.html#copytree-example.
+
+    Parameters
+    ----------
+    src :
+        Source directory tree.
+    dst :
```
cbkerr, line 39: if I'm reading the code correctly:
```suggestion
    dst : str
```

signac/syncutil.py
```diff
@@ -15,14 +20,30 @@
...
 
 def copytree(src, dst, copy_function=shutil.copy2, symlinks=False):
-    "Implementation adapted from https://docs.python.org/3/library/shutil.html#copytree-example'."
+    """Recursively copy a directory tree from src to dst, using a custom copy function.
+
+    Implementation adapted from https://docs.python.org/3/library/shutil.html#copytree-example.
+
+    Parameters
+    ----------
+    src :
```
cbkerr, line 37:
```suggestion
    src : str
```

signac/sync.py
bdice: I have no idea what to put for the type of `doc_sync`. It supports `None`, `False`, `'copy'`, and callables. Users are supposed to provide one of the properties/functors from the `DocSync` class, but I'm not sure how to express that as a type. @csadorf Do you have thoughts? Maybe this?
```suggestion
    doc_sync : attribute or callable provided by :py:class:`~signac.sync.DocSync`
```

bdice: I went ahead and committed a slightly edited version of that suggestion above.

bdice:
@cbkerr Excellent suggestions, thanks for catching the things I missed. I agree we should adopt the use of the "optional" keyword, but I realize that it is missing in many many places. I will create a separate issue for that so that we remember to apply it everywhere once our docstrings have all been converted.

signac/sync.py
csadorf: I think that's fine, but maybe you can add an example on how to use this function as part of the general description? I think that would make it much easier for users to understand how to use this.

Issue #344, opened 2020-07-06
NumPy docstyle should denote "optional" arguments


bdice:
### Description

@cbkerr [noted in a PR review](https://github.com/glotzerlab/signac/pull/343/#pullrequestreview-439196468) that I didn't use the `optional` flag for optional arguments. I think this problem exists all throughout the new NumPy-style docs and should be dealt with before finishing #315. I'm creating a separate issue because I would perform this as a "last step" after all docstrings have been converted.

### Proposed solution

It should be relatively easy to detect optional arguments because they should all have something like "(Default value: x)" in the docstring. Of course, it's possible some docstrings are missing that and should be updated. While this is happening, it would be good to standardize on a few formatting things. We have some mixture of:
1. Periods before and after the parentheses (I prefer after).
2. Code font for the default value (I prefer code font for anything except False, True, or None, including strings with single quotes like `'default_string'`).
3. Wording: `Defaults to x.` vs. `(Default value = x)`

I think my preferences match the majority of the code base already:
```
    parameter_name : bool, optional
        Description of parameter (Default value = True).
    another_parameter : str, optional
        Description of parameter (Default value = ``'hello world'``).
```

Also, I find this ambiguous and don't know what to do: we have many places in the code where we define a default value as `None` but then use a function or [other non-mutable argument](https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/) like `[]` or `{}` if the provided value is `None`. What is the right way to document this? Do we say the default is actually `None` or do we say the default is whatever replaces `None`? In my recollection, we're specifying the replacement value for functions and strings but not for empty collections (which may be accepted as an implementation detail).
```
    function_parameter : callable, optional
        Function used to copy data (Default value = :func:`shutil.copytree`).
    list_parameter: sequence of :class:`~signac.contrib.job.Job`, optional
        Sequence of jobs to use (Default value = ``[]``)
```

vyasr:
In my opinion, the default should be the actual default (e.g. `None`), but the docstring should describe what happens in that case. For example: "If `None`, files are copied using :func:`shutil.copytree`"

csadorf:
I agree with Vyas. The formalized default value should be the one that is actually matching the function signature, but the description should mention what the code logic will then default to. I think it's important to mention the "actual" default value, so that users have the option to explicitly provide it if they want to.

kidrahahjo:
Can I work on this one after #354 is merged to `numpy_docs`?

[...]

PR #354, opened 2020-07-09
Enforce pydocstyle in core, common, contrib.

bdice:
## Description
This PR enforces pydocstyle in the `core`, `common`, and `contrib` submodules. It does _not_ yet adhere to NumPy docstring style. 

After this PR, we could merge `numpy_docs` into the main branch. This would allow us to begin enforcing `pydocstyle` on all pull requests as we continue to review the last bits of #315 and finalize the transition to NumPy-style docstrings.

## Motivation and Context
Resolves some of the last steps of #315. See comments on #315 for further discussion.

## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/common/config.py
+++ b/signac/common/config.py
@@ -95,7 +95,7 @@ def check_and_fix_permissions(filename):
 
 
 def read_config_file(filename):
-    """Read signac configuration file."""
+    """Read a configuration file."""
     logger.debug("Reading config file '{}'.".format(filename))
     try:
         config = Config(filename, configspec=cfg.split('\n'))
--- a/signac/common/config.py
+++ b/signac/common/config.py
@@ -1,6 +1,8 @@
 # Copyright (c) 2017 The Regents of the University of Michigan
 # All rights reserved.
 # This software is licensed under the BSD 3-Clause License.
+"""Parses signac config files."""
+
 import os
 import stat
 import logging
@@ -19,10 +21,26 @@ FN_CONFIG = os.path.expanduser('~/.signacrc')
 
 
 class PermissionsError(ConfigError):
+    """Indicates an error in file permissions."""
     pass
 
 
+def _search_local(root):
+    for fn in CONFIG_FILENAMES:
+        fn_ = os.path.abspath(os.path.join(root, fn))
+        if os.path.isfile(fn_):
+            yield fn_
+
+
 def search_tree(root=None):
+    """Locates signac configuration files in a directory hierarchy.
+
+    Parameters
+    ----------
+    root : str
+        Path to search. Uses ``os.getcwd()`` if None (Default value = None).
+
+    """
     if root is None:
         root = os.getcwd()
     while(True):
@@ -38,12 +56,14 @@ def search_tree(root=None):
 
 
 def search_standard_dirs():
+    """Locates signac configuration files in standard directories."""
     for path in CONFIG_PATH:
         for fn in _search_local(path):
             yield fn
 
 
 def check_permissions(filename):
+    """Verify that saved passwords are only readable by the current user."""
     st = os.stat(filename)
     if st.st_mode & stat.S_IROTH or st.st_mode & stat.S_IRGRP:
         raise PermissionsError("Permissions of configuration file '{fn}'"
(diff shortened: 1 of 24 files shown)


signac/common/config.py
```diff
@@ -73,6 +95,7 @@ def check_and_fix_permissions(filename):
 
 
 def read_config_file(filename):
+    """Read signac configuration file."""
```
bdice, line 98:
```suggestion
    """Read a configuration file."""
```

signac/common/validate.py
```diff
@@ -11,6 +13,7 @@
 
 
 def mongodb_uri(value, *args, **kwargs):
+    """Return a mongodb URI."""
```
bdice, line 16:
```suggestion
    """Return a MongoDB URI."""
```

mikemhenry:
For those reviewing, don't worry about the failing pypy tests.

signac/common/config.py
```diff
@@ -19,10 +21,26 @@
...
+        if os.path.isfile(fn_):
+            yield fn_
+
+
 def search_tree(root=None):
+    """Locates signac configuration files in a directory hierarchy.
+
+    Parameters
+    ----------
+    root : str
```
kidrahahjo, line 40:
```suggestion
    root : str, optional
```
If it is not necessary to specify a keyword argument, we should use optional.
I read this in the block Sections/Parameters on this [website](https://numpydoc.readthedocs.io/en/latest/format.html_)

signac/common/config.py
```diff
@@ -53,10 +73,12 @@ def check_permissions(filename):
 
 
 def fix_permissions(filename):
+    """Set file permissions to be strictly user-readable and user-writable."""
     os.chmod(filename, stat.S_IRUSR | stat.S_IWUSR)
 
 
 def check_and_fix_permissions(filename):
+    """Verify file permissions and fix problems if needed."""
```
kidrahahjo, line 81: We should probably describe what `filename` is.

signac/common/config.py
```diff
@@ -88,20 +111,15 @@ def read_config_file(filename):
     return config
 
 
-def get_config(infile=None, configspec=None, * args, **kwargs):
+def get_config(infile=None, configspec=None, *args, **kwargs):
+    """Get configuration from a file."""
```
kidrahahjo, line 115: Since I'm not too much friendly with the code for this repository, I'm not sure whether we should describe these params or not.

signac/common/config.py
```diff
@@ -38,12 +56,14 @@ def search_tree(root=None):
 
 
 def search_standard_dirs():
+    """Locates signac configuration files in standard directories."""
```
kidrahahjo, line 59:
```suggestion
    """Locates signac configuration files in standard directories.
    
    """
```

[One line summary](https://numpydoc.readthedocs.io/en/latest/format.html#sections) section shows that the inverted commas should be in a different line. (Though I'm not sure)

signac/common/config.py
kidrahahjo: Okay I just noticed that there's an issue(#344)  open regarding it.

bdice:
@kidrahahjo Thanks for the review! You raised good questions. To clarify the intention of this PR, I'm limiting the scope to "making pydocstyle pass," not "making perfect numpydoc compatible docstrings." Most of the other PRs related to #315 have aimed to create "good" numpydoc compatible docstrings, but this one is a little different because it doesn't touch much of the public, documented API.

For functions in our public documentation, we definitely want to have parameter names and types. However, I'm not sure all of the functions I edited in this PR should belong in the public API. I did confirm that they don't appear in our published documentation, so it is unlikely that users will interface with them. There were a small number of functions that I had no idea how to document, so I just ignored the `pydocstyle` check entirely with `# noqa: D103`.

With that said, I'm going to leave the imperfect docstrings you commented on as they are.

@b-butler @glotzerlab/signac-committers If you disagree with this approach, we can go further on this PR, but I would prefer to have something imperfect and done. Once this PR is merged into `numpy_docs`, we could merge `numpy_docs` into the main branch and enable `pydocstyle` checks for all future PRs, which I think is important so that we don't lose momentum and stall out on this ongoing effort.

signac/common/config.py
bdice: Interesting. I hadn't noticed that in the examples before. I think that we're pretty consistent with using a single line here (no blank lines), and `pydocstyle` doesn't seem to complain, so I think I would prefer to leave this as it is.

PR #358, opened 2020-07-19
Support for pre-commit hook using `pre-commit`.

### LINKED ISSUE #355: The flake8 pre-commit hook is deprecated
### Description

Contributors are advised to install the flake8 pre-commit as described [here](https://docs.signac.io/en/latest/community.html#contributions).
However, this particular hook is [deprecated](https://gitlab.com/pycqa/flake8/-/issues/568).

### To reproduce

Try to install the hook with:
```
flake8 --install-hook git
```
Which will trigger a deprecation warning.

### Error output

```
WARNING: flake8 vcs hooks integration is deprecated and scheduled for removal in 4.x.  For more information, see https://gitlab.com/pycqa/flake8/issues/568
```

### Proposed solution

Switch to [pre-commit](https://pre-commit.com/).

kidrahahjo:
## Description

Enabling the support of [pre-commit](https://pre-commit.com/) for pre-commit hook.
Developers will now have to perform these steps in order to enable the pre-commit hook.
```
pip install -r requirements-precommit.txt
pre-commit install
```
This will add the hook to `git`.

## Motivation and Context

Resolves #355 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,21 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
```
kidrahahjo, line 2: This repo provides lots of hooks.
I think we should adopt the hooks like `requirements-txt-fixer`, `file-contents-sorter`, `double-quote-string-fixer`. `check-yaml`, `pretty-format-json`, and `check-json`.
There are more of them but I think these were relatively more relevant.

kidrahahjo:
This is the output when I run `pre-commit run --all-files`.

> Fix End of Files.........................................................Failed
> - hook id: end-of-file-fixer
> - exit code: 1
> - files were modified by this hook
> 
> Fixing .zenodo.json
> 
> Trim Trailing Whitespace.................................................Passed
> pydocstyle...............................................................Passed
> mypy.....................................................................Failed
> - hook id: mypy
> - exit code: 1
> 
> doc/conf.py:237: error: Need type annotation for 'latex_elements' (hint: "latex_elements: Dict[<type>, <type>] = ...")
> .sync-zenodo-metadata.py:27: error: Incompatible types in assignment (expression has type "None", variable has type "str")
> tests/test_indexing.py:24: error: Incompatible types in assignment (expression has type "None", variable has type "str")
> tests/test_indexing.py:76: error: Need type annotation for 'files' (hint: "files: Dict[<type>, <type>] = ...")
> tests/test_db.py:17: error: Incompatible types in assignment (expression has type "None", variable has type "str")
> Found 5 errors in 4 files (checked 70 source files)
> 
> flake8...................................................................Failed
> - hook id: flake8
> - exit code: 1
> 
> signac/common/configobj/__init__.py:2007:52: E741 ambiguous variable name 'l'
> tests/test_project.py:993:25: E741 ambiguous variable name 'l'
> tests/test_project.py:1010:25: E741 ambiguous variable name 'l'
> tests/test_project.py:1022:25: E741 ambiguous variable name 'l'
> tests/test_project.py:1039:25: E741 ambiguous variable name 'l'
> tests/test_project.py:1051:25: E741 ambiguous variable name 'l'
> tests/test_project.py:1067:25: E741 ambiguous variable name 'l'
> tests/test_project.py:1078:25: E741 ambiguous variable name 'l'
> tests/test_project.py:1095:25: E741 ambiguous variable name 'l'
> signac/contrib/utility.py:258:18: E741 ambiguous variable name 'l'
> signac/contrib/filesystems.py:200:1: E302 expected 2 blank lines, found 1
> signac/contrib/filesystems.py:240:1: E302 expected 2 blank lines, found 1
> signac/contrib/filterparse.py:21:35: E741 ambiguous variable name 'l'
> signac/__main__.py:188:35: E741 ambiguous variable name 'l'

bdice:
@kidrahahjo Nice work! Those failures are expected right now. You're testing with a newer version of `flake8` than we've been using, which enforces the variable naming conventions. Also I am not surprised that mypy fails on the files you mentioned. Feel free to propose changes to those files in this PR, or I can take a look if you don't have time.

kidrahahjo:
@bdice I'll make the changes to resove all the failing tests

kidrahahjo:
@bdice I've pushed the changes. I think we're now ready to use flake8 version 3.8.3 for `signac`. Should I include that in `requirements-dev.txt`?

.pre-commit-config.yaml
tommy-waltmann: I like your list and the hooks you've already added. I think a couple more hooks of potential interest are `name-tests-test` and `debug-statements`

kidrahahjo: @tommy-waltmann Thanks for the review.
I'll try to add them locally and check whether it catches any error or not.

mikemhenry:
Before we merge this, I think we should at least have a PR open in our docs repo which describes how to install the `pre-commit` hooks. This is the file that needs to be updated: https://github.com/glotzerlab/signac-docs/blob/master/docs/source/community.rst

kidrahahjo:
@mikemhenry will make an PR today describing the steps to enable the pre-commit hook.

kidrahahjo:
@mikemhenry @bdice @tommy-waltmann 
I think we should create a separate file `requirements-commit.txt` which will get installed before every commit or maybe similar so that in case we update the library versions, developers won't have to install the versions every time we introduce a new version.

.pre-commit-config.yaml
kidrahahjo: @tommy-waltmann, the `name-tests-test` hook checks whether the name of all the files in the folder `tests/` starts with `test_` or not.
Given this I think it may restrict the development as in signac-flow, we have python modules named `define_test_project.py`, `extract_templates.py` in the `tests/` folder.

kidrahahjo: @tommy-waltmann, the `name-tests-test` hook checks whether the name of all the files in the folder `tests/` starts with `test_` or not.
Given this I think it may restrict the development as in signac-flow, we have python modules named `define_test_project.py`, `extract_templates.py` in the `tests/` folder.

tommy-waltmann: Yeah, on second thought this one may not be the best idea. The unit testing frameworks also enforce this by themselves to some extent by requiring the files be named a certain way to be run as tests.

mikemhenry:
> @mikemhenry @bdice @tommy-waltmann
> I think we should create a separate file `requirements-commit.txt` which will get installed before every commit or maybe similar so that in case we update the library versions, developers won't have to install the versions every time we introduce a new version.

I don't think I quite understand what would go in `requirements-commit.txt`

signac/__main__.py
```diff
@@ -185,7 +185,7 @@ def _read_index(project, fn_index=None):
     if fn_index is not None:
         _print_err("Reading index from file '{}'...".format(fn_index))
         fd = open(fn_index)
-        return (json.loads(l) for l in fd)
+        return (json.loads(i) for i in fd)
```
mikemhenry, line 188: This is fixed in a different PR, but swapping `l` for `i` to get the tool to be happy isn't a great reason to change it. I'm guessing it allows `i` due to the prolific use of that variable name (and most of the time it is clear what it means). Instead, use `_` to indicate that we have a temp variable that we don't care about.

mikemhenry, line 188:
```suggestion
        return (json.loads(_) for _ in fd)
```

signac/contrib/filterparse.py
```diff
@@ -18,7 +18,7 @@ def _read_index(project, fn_index=None):
     if fn_index is not None:
         _print_err("Reading index from file '{}'...".format(fn_index))
         fd = open(fn_index)
-        return (json.loads(l) for l in fd)
+        return (json.loads(i) for i in fd)
```
mikemhenry, line 21:
```suggestion
        return (json.loads(_) for _ in fd)
```

tests/test_project.py
```diff
@@ -990,8 +990,8 @@ def test_export_custom_path_string_modify_tree_flat(self):
         for i in range(10):
             for j in range(2):
                 for k in range(2):
-                    for l in range(2):
-                        self.project.open_job(dict(a=i, b=j, c=k, d=l)).init()
+                    for m in range(2):
```
mikemhenry, line 993: I know these are just tests, but IMHO going from `l` to `m` here is confusing since we are going `i, j, k, l`. My preference would be to add some sort of #noqa to make flake8 happy.

signac/__main__.py
bdice: Or even better, a descriptive name like `line`.
```suggestion
        return (json.loads(line) for line in fd)
```

signac/contrib/filterparse.py
bdice: Or `line`, see previous comment.
```suggestion
        return (json.loads(line) for line in fd)
```

tests/test_project.py
bdice: Or descriptive names like `a_value`, `b_value`, `c_value`, `d_value` that correspond to the state point.

signac/__main__.py
kidrahahjo: Sometimes we use `_` to just ignore the variable and here we're using that particular variable. 
I think replacing it with `line` should be fine.

signac/contrib/filterparse.py
kidrahahjo: @mikemhenry Since, for a similar method in `__main__.py`, we have used `line`. Hence, I think I should use `line` here too.

tests/test_project.py
kidrahahjo: There are several places where this is repeated. Hence, instead of using #noqa everywhere, I think I should go with changing all the names.

kidrahahjo:
@mikemhenry We manually mention the libraries for CI testing and we currently use the command mentioned below.
`pip install --progress-bar off --user -U flake8==3.7.1 pydocstyle==5.0.2 mypy==0.770`.

In requirements-commit.txt, we'll be adding these libraries:
> pre-commit==2.6.0
> flake8==3.7.1 
> pydocstyle==5.0.2 
> mypy==0.770

After this we should edit the command for CI testing -> `pip install --progress-bar off --user -r requirements-commit.txt`.
And now we'd simply have to update this file whenever we change the versions.

The user will be advised to install this file for setting up a pre-commit hook now.

kidrahahjo:
> Is it possible to have the commit hooks use the `setup.cfg` file to figure out what arguments to use for the different tools? I believe this happens by default (I removed the args line and it seemed to still pass, so I think it using the `setup.cfg` but if you could double check that, that would be great.
> 
> By using `setup.cfg` then we have a single point of truth with what options we want `pydocstyle` to use and this will reduce the maintenance burden of copying changes to a bunch of different files.

@mikemhenry it's not working for me locally. I don't know if it's a bug or not but it's not reading the `setup.cfg` file at all.
I looked on the internet but didn't found anything useful. Can you please have a look?

signac/contrib/filterparse.py
mikemhenry: Yes `line` is great! My suggestion for `_` was to use it as a throw away variable instead of something like `i` or `l` BUT always always use something descriptive if it makes sense!

bdice: 😊 I distinguish "temporary" variables like `line` that are re-used by name at least once from "throwaway" variables that are never used by name. I use `_` when dealing with functions like [NumPy histograms](https://numpy.org/doc/stable/reference/generated/numpy.histogram.html) that return a tuple like `histogram, bin_edges`. If I don't plan to use bin edges, then I call it like
```python
histogram, _ = np.histogram(...)
```

kidrahahjo:
@mikemhenry 
pre-commit just passes arguments to pycodestyle, so probably pycodestyle  is ignoring the regular expressions.
Please have a look at it [here](https://pre-commit.com/#arguments-pattern-in-hooks)

Hence we'd explicitly have to use pre-commit's `files` option to let it know which files we want to check.

kidrahahjo:
@bdice 
I think this is ready to get merged now

PR #187, opened 2019-05-11
Ensure that collection file is closed. Resolves #186.

bdice:
## Motivation and Context
Resolves unclosed file error in #186.

## Types of Changes
- [x] Bug fix

## Checklist:
- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.txt).
- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
- [x] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [ ] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt).

PR #359, opened 2020-07-28
improve cli docs

atravitz:
#187  Description

Adding a section for CLI in the API docs.

## Motivation and Context

A trend I've noticed is that many people (myself included) aren't aware of much of the command line functionality of signac. Making clearer and more visible documentation might help this.

## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:


cbkerr:
I got these building locally, and I like the way that this is shaping up!

I wonder if we can format the CLI examples in such a way that mirrors what users would type to reproduce the output. For instance, in practice I would run `signac job -h` not `python -m signac job -h`. Maybe this is an insignificant difference.
You have the note at the top about running directly with the signac command, but if possible I think we should run the examples with the signac command and put the longer way of running the examples as a note.

It looks like we can configure the [option](https://ipython.readthedocs.io/en/stable/sphinxext.html#ipython-sphinx-directive-module) to import certain packages by changing the setting `ipython_execlines`. We cannot use the setting `ipython_promptin` to make the inputs `In[1]` look like a `$ ` to signify being a command line because that option expects a placeholder `%d` for the line number.

cbkerr:
I found a [package](https://github.com/NextThought/sphinxcontrib-programoutput) ([on conda](https://anaconda.org/conda-forge/sphinxcontrib-programoutput)) that enables embedding the output of any command line command. I tried it on the signac help commands and it works as expected.

After adding it to extensions, we [invoke](https://sphinxcontrib-programoutput.readthedocs.io/en/latest/) it with:
```
.. program-output:: signac project -h
```

What do you think of this approach?

atravitz:
@cbkerr, yes I like that approach but I think @bdice wanted to avoid adding new dependencies if we can. I agree that the `program-output` method is clearer. @bdice, thoughts?

bdice:
@atravitz @cbkerr This package seems useful and stable (as a dependency). I'm glad we tried ipython first, just to know if it suited our needs, but I agree the programoutput package is worth including if it solves the problem and provides better functionality. Give it a try!

bdice:
We may want to add our documentation dependencies to the `requirements-dev.txt` list. (Possibly a separate issue/PR.)

doc/cli.rst
```diff
@@ -0,0 +1,146 @@
...
+The following core **signac** functions are---in addition to the Python interface---accessible
+directly via the ``$ signac`` command.
+
+
+init
+====
+
+.. ipython::
+
+   In [1]: !python -m signac init -h
```
bdice, line 18: Let's use `--help` instead of `-h` for clarity.

doc/index.rst
```diff
@@ -21,6 +21,7 @@ Contents
    :maxdepth: 2
 
    installation
+   cli
```
bdice, line 24: Personally I'd put this below API but I don't have a strong opinion.

cbkerr: When I started using signac, I think I used the command line more for daily tasks before using signac in scripts, so based on "order of learning" maybe CLI can be first.

atravitz:
can we merge this for now, and i'll make more PRs to add more detail? also, I'd like to link to this from `signac-docs` but I'm not sure how to go about that.

doc/cli.rst
```diff
@@ -0,0 +1,112 @@
...
+import
+======
+
+.. program-output:: signac import --help
+
+
+export
+======
+
+.. program-output:: signac export --help
```
cbkerr, line 112:
```suggestion
.. program-output:: signac export --help


config
======

.. program-output:: signac config --help

```

bdice:
@atravitz @vyasr Here is an example of how to link from signac-docs to a package's docs. *Don't hardcode links!*

Dashboard page in signac-docs has a reference ``:ref:`dashboard-security`` `: https://github.com/glotzerlab/signac-docs/blob/master/docs/source/dashboard.rst

... which links to the section `.. _dashboard-security:` here: https://raw.githubusercontent.com/glotzerlab/signac-dashboard/master/doc/security.rst

In short, just make sure you have a section header like `.. _signac-cli-reference` *above the heading* for this section of the docs and then add ``:ref:`signac-cli-reference``` in `signac-docs` (or elsewhere in this package's documentation) where desired. The important part is to make sure that `signac-cli-reference` is a *unique name* and won't be used elsewhere. The more descriptive, the better. Ideally it matches the name of the (hopefully unique) section header.

This cross-referencing works because we have an intersphinx linkage between signac-docs and signac: https://github.com/glotzerlab/signac-docs/blob/8b8e1138a82ad6b95e5b6aa2ab00e8a2ba786695/docs/source/conf.py#L184-L191

vyasr:
@bdice good catch, thanks. I didn't think we had intersphinx set up since the first thing I checked was the packages page on signac-docs, [which currently hardcodes links](https://raw.githubusercontent.com/glotzerlab/signac-docs/master/docs/source/packages.rst). Probably worth fixing that at some point as well.

bdice:
@vyasr @atravitz @cbkerr I created a `signac-docs` wiki entry describing this process for the future. I've had to write this description several times, so it will be nice to have a reference to demonstrate it. https://github.com/glotzerlab/signac-docs/wiki/How-to-cross-reference-links-between-package-docs-and-signac-docs

csadorf:
The build for this commit failed: https://readthedocs.org/projects/signac-core/builds/11587075/

PR #364, opened 2020-07-31
Added backends to synced_collection

vishav1771:
## Description

Implemented Redis, Zarr, and MongoDB backends for SyncedCollection API.
## Motivation and Context

Related to #249.
## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


tests/test_synced_collection.py
```diff
@@ -279,7 +349,7 @@ def test_call(self, synced_dict, testdata):
     def test_reopen(self, synced_dict, testdata):
         key = 'reopen'
         synced_dict[key] = testdata
-        synced_dict2 = deepcopy(synced_dict)
+        synced_dict2 = copy(synced_dict)
```
vishav1771, line 352: `deepcopy` try to copy 'redis client' which raises error so I changed it to `copy`.

signac/core/mongodbcollection.py
```diff
@@ -0,0 +1,147 @@
+# Copyright (c) 2020 The Regents of the University of Michigan
```
csadorf, line 1: Please rename module to `collection_mongodb.py`.

signac/core/mongodbcollection.py
```diff
@@ -0,0 +1,147 @@
...
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+"""Implements Mongo-backend.
+
+This implements the Mongo-backend for SyncedCollection API by
```
csadorf, line 6:
```suggestion
This implements the MongoDB-backend for SyncedCollection API by
```
"MongoDB" is a name, please make sure to use it consistently correctly.

signac/core/mongodbcollection.py
```diff
@@ -0,0 +1,147 @@
...
+implementing sync and load methods.
+"""
+from .synced_collection import SyncedCollection
+from .syncedattrdict import SyncedAttrDict
+from .synced_list import SyncedList
+
+
+class MongoCollection(SyncedCollection):
```
csadorf, line 14:
```suggestion
class MongoDBCollection(SyncedCollection):
```
Please use the name "MongoDB" and its translation into code objects correctly and consistently.

signac/core/mongodbcollection.py
```diff
@@ -0,0 +1,147 @@
...
+    """Implement sync and load using a Mongo backend."""
```
csadorf, line 15:
```suggestion
    """Implement sync and load using a MongoDB backend."""
```

signac/core/mongodbcollection.py
```diff
@@ -0,0 +1,147 @@
...
+            self._client = client
+        self._db = self._client.get_database(database)
+        self._collection = self._db.get_collection(collection)
+        self._name = name
+        self._key = type(self).__name__ + '::name'
+        super().__init__(**kwargs)
+        if (name is None) == (self._parent is None):
+            raise ValueError(
+                "Illegal argument combination, one of the two arguments, "
+                "parent or name must be None, but not both.")
```
csadorf, line 36: We can massively simplify this constructor implementation and improve the API by only accepting an already instantiated instance of [`pymongo.Collection`](https://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection) here. This has the benefit that we don't need to sanitize and validate the input arguments, deal with establishing the connection and related issues, and we don't need to import `pymongo` at all within this module. Finally, this API would be much more flexible for users and developers and decouples it from potential breaking upstream changes in the `pymongo` package.
```suggestion
    def __init__(self, collection, **kwargs):
        self._collection = collection
        super().__init__(**kwargs)
```

signac/core/mongodbcollection.py
```diff
@@ -0,0 +1,147 @@
...
+    database : string
+        Name of database (Default value = 'signac_db').
+    collection : string
+        Name of collection (Default value = 'collection')
+    data: mapping, optional
+        The intial data pass to MongoDict. Defaults to `dict()`
+    parent: object, optional
+        A parent instance of MongoDict or None (Default value = None).
+    """
+    pass
```
csadorf, line 99: Looks like A+ abstraction here. 👍 
Do we even need the `pass` statement here?

signac/core/mongodbcollection.py
```diff
@@ -0,0 +1,147 @@
...
+        Name of database (Default value = 'signac_db').
+    collection : string
+        Name of collection (Default value = 'collection')
+    data: mapping, optional
+        The intial data pass to MongoList. Defaults to `list()`
+    parent: object, optional
+        A parent instance of MongoList or None (Default value = None).
+    """
+
+    pass
```
csadorf, line 144: Not sure if we even need the `pass` statement, but there is inconsistent spacing here compared to the previous class.

signac/core/rediscollection.py
```diff
@@ -0,0 +1,141 @@
...
+            redis_kwargs = redis_kwargs if redis_kwargs is not None else {}
+            self._client = redis.Redis(**redis_kwargs)
+        else:
+            self._client = client
+        self._name = name
+        super().__init__(**kwargs)
+        if (name is None) == (self._parent is None):
+            raise ValueError(
+                "Illegal argument combination, one of the two arguments, "
+                "parent or name must be None, but not both.")
```
csadorf, line 34: Similar to my arguments for MongoDB, I would try to move the establishment of connections out of the scope of this constructor as much as possible for a simpler and clearer API:
```suggestion
    def __init__(self, name=None, client=None, **kwargs):
        self._client = client
        self._name = name
        super().__init__(**kwargs)
```

signac/core/rediscollection.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def _load(self):
+        """Load the data from a Redis-database."""
+        blob = self._client.get(self._name)
+        return json.loads(blob) if blob is not None else blob
```
csadorf, line 39:
```suggestion
        return None if blob is None else json.loads(blob)
```
Slightly easier to interpret I think.

signac/core/rediscollection.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def _sync(self):
+        """Write the data from Redis-database."""
+        data = self.to_base()
+        # Serialize data:
+        blob = json.dumps(data).encode()
+
+        self._client.set(self._name, blob)
```
csadorf, line 47:
```suggestion
        self._client.set(self._name, json.dumps(self.to_base()).encode())
```

signac/core/zarrcollection.py
```diff
@@ -0,0 +1,136 @@
...
+        import numcodecs  # zarr depends on numcodecs
+
+        self._root = zarr.group(store=store)
+        self._name = name
+        self._object_codec = numcodecs.JSON()
+        super().__init__(**kwargs)
+        if (name is None) == (self._parent is None):
+            raise ValueError(
+                "Illegal argument combination, one of the two arguments, "
+                "parent or name must be None, but not both.")
```
csadorf, line 30: We keep repeating this particular check in all constructors, maybe we can move that into the parent constructor?

signac/core/zarrcollection.py
```diff
@@ -0,0 +1,136 @@
...
+class ZarrCollection(SyncedCollection):
+    """Implement sync and load using a Zarr backend."""
+
+    backend = __name__  # type: ignore
+
+    def __init__(self, name=None, store=None, **kwargs):
+        import zarr
+        import numcodecs  # zarr depends on numcodecs
+
+        self._root = zarr.group(store=store)
```
csadorf, line 23: Similar to the MongoDB and Redis backends, I suggest a slight shift of the API here. Since we are not using the `store` argument anywhere else, it would probably make sense to directly pass the `root` group to the constructor.

signac/core/zarrcollection.py
```diff
@@ -0,0 +1,136 @@
...
+                "parent or name must be None, but not both.")
+
+    def _load(self):
+        """Load the data from a Radis-database."""
+        try:
+            dataset = self._root[self._name]
+            data = dataset[0]
+        except KeyError:
+            data = None
+        return data
```
csadorf, line 39:
```suggestion
        try:
            return self._root[self.name][0]
        except KeyError:
            return None
```

signac/core/zarrcollection.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def _sync(self):
+        """Write the data from Radis-database."""
+        data = self.to_base()
+        # Serialize data:
```
csadorf, line 44: ?

tests/test_synced_collection.py
```diff
@@ -8,11 +8,17 @@
...
 
 from signac.core.synced_list import SyncedCollection
 from signac.core.jsoncollection import JSONDict
 from signac.core.jsoncollection import JSONList
+from signac.core.zarrcollection import ZarrDict
+from signac.core.zarrcollection import ZarrList
+from signac.core.rediscollection import RedisDict
+from signac.core.rediscollection import RedisList
+from signac.core.mongodbcollection import MongoDict
+from signac.core.mongodbcollection import MongoList
```
csadorf, line 21: I would very much prefer to have the backend specific tests in their own respective test modules.

tests/test_synced_collection.py
csadorf: That changes the test fundamentally, not a good idea. Instead try to implement the `RedisDict.__deepcopy__()` function. Please see [here](https://docs.python.org/3.7/library/copy.html) for more information on how to do this.

tests/test_synced_collection.py
```diff
@@ -602,3 +672,112 @@ def synced_list(self):
...
+
+        key = 'write_invalid_type'
+        synced_dict[key] = testdata
+        assert len(synced_dict) == 1
+        assert synced_dict[key] == testdata
+        d2 = Foo()
+        with pytest.raises(bson.errors.InvalidDocument):
+            synced_dict[key + '2'] = d2
+        assert len(synced_dict) == 1
+        assert synced_dict[key] == testdata
```
csadorf, line 767: Please catch the exception in the backend implementation and raise the correct exception type so that we can remove this backend-specific test.

signac/core/mongodbcollection.py
csadorf: And apply similar naming scheme to the other modules.

tests/test_synced_collection.py
vishav1771: Okay, will add that.

signac/core/mongodbcollection.py
vishav1771: No, we don't need that. I have removed it.

signac/core/zarrcollection.py
vishav1771: I'll remove that.

signac/core/collection_redis.py
```diff
@@ -18,33 +19,22 @@ class RedisCollection(SyncedCollection):
...
     def _sync(self):
         """Write the data from Redis-database."""
-        data = self.to_base()
-        # Serialize data:
-        blob = json.dumps(data).encode()
+        self._client.set(self._name, json.dumps(self.to_base()).encode())
 
-        self._client.set(self._name, blob)
+    def __deepcopy__(self, memo):
+        return type(self)(client=self._client, name=self._name, data=self.to_base(),
```
csadorf, line 36: For a proper deepcopy you have to deepcopy `self._client` and `self._name` as well.

signac/core/collection_zarr.py
```diff
@@ -16,36 +18,31 @@ class ZarrCollection(SyncedCollection):
...
         self._object_codec = numcodecs.JSON()
         super().__init__(**kwargs)
-        if (name is None) == (self._parent is None):
-            raise ValueError(
-                "Illegal argument combination, one of the two arguments, "
-                "parent or name must be None, but not both.")
 
     def _load(self):
-        """Load the data from a Radis-database."""
+        """Load the data."""
```
csadorf, line 29: I did not notice this before, but the word "data" is typically assumed to be plural in this context:
```suggestion
        """Load data."""
```
Furthermore, I think it would make sense to be more explicit here:
```suggestion
        """Load data from zarr-file."""
```
Please apply everywhere else as well.

signac/core/collection_zarr.py
```diff
@@ -16,36 +18,31 @@ class ZarrCollection(SyncedCollection):
...
-        """Write the data from Radis-database."""
+        """Write the data."""
         data = self.to_base()
-        # Serialize data:
         dataset = self._root.require_dataset(
             self._name, overwrite=True, shape=1, dtype='object', object_codec=self._object_codec)
         dataset[0] = data
 
+    def __deepcopy__(self, memo):
+        return type(self)(group=self._root, name=self._name, data=self.to_base(),
```
csadorf, line 43: For a proper deepcopy, you would need to deepcopy `self._root` and `self._name` as well.

signac/core/collection_mongodb.py
```diff
@@ -44,10 +35,17 @@ def _sync(self):
...
         data = self.to_base()
         data_to_insert = {self._key: self._name, 'data': data}
-        self._collection.replace_one({self._key: self._name}, data_to_insert, True)
+        try:
+            self._collection.replace_one({self._key: self._name}, data_to_insert, True)
+        except self._errors.InvalidDocument as err:
+            raise TypeError(str(err))
 
+    def __deepcopy__(self, memo):
+        return type(self)(collection=self._collection, name=self._name, data=self.to_base(),
```
csadorf, line 44: For a proper deepcopy, you have to deepcopy `self._collection` and `self._name` as well.

signac/core/collection_zarr.py
bdice: Minor clarification: Dropping the article “the” is a stylistic choice, not one that depends on data being plural or singular. (Otherwise it’d be a debate of “a data” for singular vs “the data” for plural, not a question of dropping the article.) I am fine with the suggested changes, regardless.

signac/core/collection_redis.py
vishav1771: I tried to do that but `deepcopy(self._client)` raise `TypeError`.

signac/core/collection_mongodb.py
vishav1771: I tried to do that but `deepcopy(self._collection)` raise `TypeError`.

signac/core/collection_zarr.py
vishav1771: I have changed it with `Load the data from zarr-store`.

signac/core/collection_redis.py
csadorf: In cases like this, do not silently **introduce a bug into the code**, but instead point out the problem in the discussion here and propose an alternative solution if you have any. I anticipated this as a potential issue, but I did not want to make any assumptions before we actually test this out.

To address this issue, I suggest that we rename `__deepcopy__` to `_pseudo_deepcopy` and use that function in the tests instead. In this way we ensure that tests are as closely similar as possible, without implementing a flawed `__deepcopy__()` function. So in the test, it should look like this:
```python
try:
    synced_dict2 = deepcopy(synced_dict)
except TypeError:
    # Use fallback implementation, deepcopy not supported by backend.
    synced_dict2 = synced_dict._pseudo_deepcopy()
```
Does this make sense? If not, please ask.

signac/core/collection_mongodb.py
csadorf: See comment above.

signac/core/collection_redis.py
vishav1771: Yes, it makes sense.

signac/core/collection_mongodb.py
```diff
@@ -40,7 +40,7 @@ def _sync(self):
         except self._errors.InvalidDocument as err:
             raise TypeError(str(err))
 
-    def __deepcopy__(self, memo):
+    def _pseudo_deepcopy(self, memo=None):
```
csadorf, line 43: - [ ] Please add a doc-string that describes the purpose of this function and the motivation for its implementation.
- [ ] Please remove the memo argument, because its use should be restricted to the `__deepcopy__` signature.

csadorf:
@bdice @tommy-waltmann This appears almost ready to merge.

signac/core/collection_zarr.py
```diff
@@ -0,0 +1,129 @@
...
+    Parameters
+    ----------
+    name: str, optional
+        The name of the  collection (Default value = None).
+    data: mapping, optional
+        The intial data pass to ZarrDict. Defaults to `dict()`.
+    group: object
+        A zarr.hierarchy.Group instance
+    parent: object, optional
+        A parent instance of ZarrDict or None (Default value = None).
```
tommy-waltmann, line 88: Should the `group` argument be documented as optional here? Perhaps it should also be the first parameter documented, since `ZarrCollection` explicitly lists `group` as the first argument and not as a part of `**kwargs`.

signac/core/collection_zarr.py
```diff
@@ -0,0 +1,129 @@
...
+    Parameters
+    ----------
+    name: str
+        The name of the  collection.
+    data: mapping, optional
+        The intial data pass to ZarrList. Defaults to `list()`.
+    group: object
+        A zarr.hierarchy.Group instance
+    parent: object, optional
+        A parent instance of ZarrList or None (Default value = None).
```
tommy-waltmann, line 125: Should `group` be documented as optional here? If not, I think it would be a good idea to list it before `data`.

csadorf:
@bdice Do you still want to review?

bdice:
> @bdice Do you still want to review?

@csadorf I’m not able to review for a few days. I am confident that the approvals here are enough to merge.

Issue #316, opened 2020-03-27
Unexpected errors when using ints as keys for JSONDict


tcmoore3:
### Description

I'm trying to go through a project and save calculation results into a `JSONDict`. The basic schema is just 2 statepoint parameters, say `p1` and `p2`. I create a `JSONDict` `dd`, and call `dd.setdefault(p1, {})`. I then get a `KeyError` when I try to access `dd[p1]`. This error only happens with `p1` is an int or float; I get the expected behavior when it is a string.

It seems like there is an issue saving the dict when I have a non-str key.

### To reproduce
This example show

```shell
>>> import signac
>>> import os; os.system('rm test.json')
>>> x = signac.JSONDict('test.json')
>>> y = x.setdefault(1, {})  # note: key is int
>>> z = y.setdefault(2, {})
>>> z['c'] = 1
>>> print(x)
{'1': {}}

>>> import signac
>>> import os; os.system('rm test.json')
>>> x = signac.JSONDict('test.json')
>>> y = x.setdefault('1', {})  # note: key is str
>>> z = y.setdefault('2', {})
>>> z['c'] = 1
>>> print(x)
{'1': {'2': {'c': 1}}}
```

In the first case, `json.dict` has the contents `{'1': {}}`, so it doesn't look like the syncing is actually happening in that case.

### System configuration

```
$ python -c 'import platform; print(platform.platform()); import sys; print(sys.version); import signac; print(signac.__version__)'
Darwin-17.7.0-x86_64-i386-64bit
3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 14:38:56) 
[Clang 4.0.1 (tags/RELEASE_401/final)]
1.2.0
```
and
```
$ python -c 'import platform; print(platform.platform()); import sys; print(sys.version); import signac; print(signac.__version__)'
Darwin-17.7.0-x86_64-i386-64bit
3.7.4 (default, Aug 13 2019, 15:17:50) 
[Clang 4.0.1 (tags/RELEASE_401/final)]
1.3.0
```

mikemhenry:
We will throw a warning if a user tries to use an `int` as a key since `int` keys are not supported in JSON.

PR #362, opened 2020-07-31
DOC: Autoreformat doc to conform to numpydoc.

Carreau:
One thing which is misparsed by numpydoc is

         Returns
         -------
    -    NoneType or :class:`~signac.sync.FileTransferStats`
    +    NoneType or : class:`~signac.sync.FileTransferStats`
             Returns stats if ``collect_stats`` is ``True``, else ``None``.

Which I of course did not add; So that might be a bug in NumPyDoc.

Partial work toward #344

## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change<sup>1</sup>

## Checklist:

- [ ] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
- [ ] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
- [ ] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
- [ ] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
- [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.

PR #378, opened 2020-08-20
Added validation layer to SyncedCollection

vishav1771:
## Description

This PR adds the validation layer to the `SyncedCollection`.
Currently, we only validate the key of `SyncedAttrDict`. This PR extends the behavior to validate the data input to `SyncedCollection`. It adds `AttrFormatValidator` for `SyncedAttrDict` and `JSONFormatValidator` for `JSONCollection`. 
## Motivation and Context

Related to #249 and #316.
## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/core/collection_json.py
+++ b/signac/core/collection_json.py
@@ -11,11 +11,33 @@ import os
 import json
 import errno
 import uuid
+import warnings
 
 from .synced_collection import SyncedCollection
 from .syncedattrdict import SyncedAttrDict
 from .synced_list import SyncedList
-from .validators import JSONFormatValidator
+from .validators import JSON_format_validator
+
+
+def _convert_key_to_str(data):
+    """Recursively convert non-string keys to strings for (potentially nested) input collections.
+
+    This retains compatibility with auto-casting keys to strings, and will be
+    removed in signac 2.0.
+    Input collections must be of "base type" (dict or list).
+    """
+    if isinstance(data, dict):
+        def _str_key(key):
+            if not isinstance(key, str):
+                warnings.warn(f"Use of {type(key).__name__} as key is deprecated "
+                              "and will be removed in version 2.0",
+                              DeprecationWarning)
+                key = str(key)
+            return key
+        return {_str_key(key): _convert_key_to_str(value) for key, value in data.items()}
+    elif isinstance(data, list):
+        return [_convert_key_to_str(value) for value in data]
+    return data
 
 
 class JSONCollection(SyncedCollection):
@@ -30,7 +52,7 @@ class JSONCollection(SyncedCollection):
         super().__init__(**kwargs)
 
     def _load(self):
-        """Load the data from a JSON-file."""
+        """Load the data from a JSON file."""
         try:
             with open(self._filename, 'rb') as file:
                 blob = file.read()
@@ -39,23 +61,11 @@ class JSONCollection(SyncedCollection):
             if error.errno == errno.ENOENT:
                 return None
 
-    def _convert_non_str_key_to_str(self, data):
-        if isinstance(data, dict):
-            def _str_key(key):
-                if not isinstance(key, str):
-                    # raise DeprecationWarning here
-                    key = str(key)
-                return key
-            return {_str_key(key): self._convert_non_str_key_to_str(value) for key, value in data.items()}
-        elif isinstance(data, list):
-            return [self._convert_non_str_key_to_str(value) for value in data]
-        return data
-
     def _sync(self):
         """Write the data to JSON file."""
         data = self.to_base()
         # Converting non-string keys to string
-        data = self._convert_non_str_key_to_str(data)
+        data = _convert_key_to_str(data)
         # Serialize data
         blob = json.dumps(data).encode()
         # When write_concern flag is set, we write the data into dummy file and then
@@ -72,7 +82,7 @@ class JSONCollection(SyncedCollection):
                 file.write(blob)
 
 
-JSONCollection.add_validator(JSONFormatValidator)
+JSONCollection.add_validator(JSON_format_validator)
 
 
 class JSONDict(JSONCollection, SyncedAttrDict):
--- a/signac/core/collection_json.py
+++ b/signac/core/collection_json.py
@@ -86,14 +86,11 @@ class JSONCollection(SyncedCollection):
     backend = __name__  # type: ignore
 
     def __init__(self, filename=None, write_concern=False, **kwargs):
-        self._filename = os.path.realpath(filename) if filename is not None else None
+        self._filename = None if filename is None else os.path.realpath(filename)
         self._write_concern = write_concern
+        kwargs['name'] = filename
         super().__init__(**kwargs)
         self._validators.append(JSONFormatValidator)
-        if (filename is None) == (self._parent is None):
-            raise ValueError(
-                "Illegal argument combination, one of the two arguments, "
-                "parent or filename must be None, but not both.")
 
     def _load(self):
         """Load the data from a JSON-file."""
@@ -160,22 +157,20 @@ class JSONDict(SyncedAttrDict, JSONCollection):
     ----------
     filename: str, optional
         The filename of the associated JSON file on disk (Default value = None).
+    write_concern: bool, optional
+        Ensure file consistency by writing changes back to a temporary file
+        first, before replacing the original file (Default value = None).
     data: mapping, optional
         The intial data pass to JSONDict. Defaults to `list()`
     parent: object, optional
         A parent instance of JSONDict or None (Default value = None).
-    write_concern: bool, optional
-        Ensure file consistency by writing changes back to a temporary file
-        first, before replacing the original file (Default value = None).
     """
-
-    pass
 
 
 class JSONList(SyncedList, JSONCollection):
     """A non-string sequence interface to a persistent JSON file.
 
-    The JSONDict inherits from :class:`~core.collection_api.SyncedCollection`
+    The JSONList inherits from :class:`~core.collection_api.SyncedCollection`
     and :class:`~core.syncedlist.SyncedList`.
 
     .. code-block:: python
@@ -198,18 +193,16 @@ class JSONList(SyncedList, JSONCollection):
 
     Parameters
     ----------
-    filename: str
+    filename: str, optional
         The filename of the associated JSON file on disk (Default value = None).
-    data: non-str Sequence
-        The intial data pass to JSONDict
-    parent: object
-        A parent instance of JSONDict or None (Default value = None).
-    write_concern: bool
+    write_concern: bool, optional
         Ensure file consistency by writing changes back to a temporary file
         first, before replacing the original file (Default value = None).
+    data: non-str Sequence, optional
+        The intial data pass to JSONList
+    parent: object, optional
+        A parent instance of JSONList or None (Default value = None).
     """
-
-    pass
 
 
 SyncedCollection.register(JSONDict, JSONList)
--- a/signac/core/collection_json.py
+++ b/signac/core/collection_json.py
@@ -41,17 +41,18 @@ class JSONCollection(SyncedCollection):
 
     def _convert_non_str_key_to_str(self, data):
         if isinstance(data, dict):
-            ret = {}
-            for key, value in data.items():
-                ret[str(key)] = self._convert_non_str_key_to_str(value)
-            return ret
+            def _str_key(key):
+                if not isinstance(key, str):
+                    # raise DeprecationWarning here
+                    key = str(key)
+                return key
+            return {_str_key(key): self._convert_non_str_key_to_str(value) for key, value in data.items()}
         elif isinstance(data, list):
-            for i in range(len(data)):
-                data[i] = self._convert_non_str_key_to_str(data[i])
+            return [self._convert_non_str_key_to_str(value) for value in data]
         return data
 
     def _sync(self):
-        """Write the data to json file."""
+        """Write the data to JSON file."""
         data = self.to_base()
         # Converting non-string keys to string
         data = self._convert_non_str_key_to_str(data)
--- a/signac/core/collection_json.py
+++ b/signac/core/collection_json.py
@@ -16,7 +16,7 @@ import warnings
 from .synced_collection import SyncedCollection
 from .syncedattrdict import SyncedAttrDict
 from .synced_list import SyncedList
-from .validators import JSON_format_validator
+from .validators import json_format_validator
 
 
 def _convert_key_to_str(data):
@@ -82,7 +82,7 @@ class JSONCollection(SyncedCollection):
                 file.write(blob)
 
 
-JSONCollection.add_validator(JSON_format_validator)
+JSONCollection.add_validator(json_format_validator)
 
 
 class JSONDict(JSONCollection, SyncedAttrDict):
--- a/signac/core/collection_json.py
+++ b/signac/core/collection_json.py
@@ -21,7 +21,7 @@ from .validators import JSONFormatValidator
 class JSONCollection(SyncedCollection):
     """Implement sync and load using a JSON back end."""
 
-    backend = __name__  # type: ignore
+    _backend = __name__  # type: ignore
 
     def __init__(self, filename=None, write_concern=False, **kwargs):
         self._filename = None if filename is None else os.path.realpath(filename)
--- a/signac/core/collection_json.py
+++ b/signac/core/collection_json.py
@@ -58,7 +58,7 @@ class JSONCollection(SyncedCollection):
                 file.write(blob)
 
 
-JSONCollection.add_validator(JSONFormatValidator())
+JSONCollection.add_validator(JSONFormatValidator)
 
 
 class JSONDict(JSONCollection, SyncedAttrDict):
--- a/signac/core/collection_json.py
+++ b/signac/core/collection_json.py
@@ -88,7 +88,7 @@ JSONCollection.add_validator(json_format_validator)
 class JSONDict(JSONCollection, SyncedAttrDict):
     """A dict-like mapping interface to a persistent JSON file.
 
-    The JSONDict inherits from :class:`~core.collection_api.SyncedCollection`
+    The JSONDict inherits from :class:`~core.synced_collection.SyncedCollection`
     and :class:`~core.syncedattrdict.SyncedAttrDict`.
 
     .. code-block:: python
@@ -134,7 +134,7 @@ class JSONDict(JSONCollection, SyncedAttrDict):
 class JSONList(JSONCollection, SyncedList):
     """A non-string sequence interface to a persistent JSON file.
 
-    The JSONList inherits from :class:`~core.collection_api.SyncedCollection`
+    The JSONList inherits from :class:`~core.synced_collection.SyncedCollection`
     and :class:`~core.syncedlist.SyncedList`.
 
     .. code-block:: python
--- a/signac/core/collection_json.py
+++ b/signac/core/collection_json.py
@@ -11,73 +11,11 @@ import os
 import json
 import errno
 import uuid
-import warnings
 
 from .synced_collection import SyncedCollection
 from .syncedattrdict import SyncedAttrDict
 from .synced_list import SyncedList
-from collections.abc import Mapping
-from collections.abc import Sequence
-from .synced_collection import NUMPY
-from ..errors import KeyTypeError
-
-if NUMPY:
-    import numpy
-
-
-class JSONFormatValidator:
-    """Implement the validation for :class:`JSONCollection`."""
-
-    @classmethod
-    def validate(cls, data):
-        """Emit a warning or raise an exception if data is invalid.
-
-        Parameters
-        ----------
-        data:
-            Data to validate.
-
-        Returns
-        -------
-        data
-
-        Raises
-        ------
-        KeyTypeError:
-            If keys of mapping have unsupported data type.
-        TypeError
-            If data type is not supported.
-        """
-
-        if isinstance(data, (str, int, float, bool, bytes, type(None))):
-            return data
-        elif isinstance(data, Mapping):
-            ret = {}
-            for key, value in data.items():
-                if isinstance(key, (int, bool, type(None))):
-                    warnings.warn(
-                        "Use of {} as key is deprecated and will be removed in version 2.0"
-                        .format(type(key)), DeprecationWarning)
-                    new_key = str(key)
-                elif isinstance(key, str):
-                    new_key = key
-                else:
-                    raise KeyTypeError(
-                        "Keys must be str, int, bool or None, not {}".format(type(key).__name__))
-                new_value = cls.validate(value)
-                ret[new_key] = new_value
-            return ret
-        elif isinstance(data, Sequence):
-            for i in range(len(data)):
-                data[i] = cls.validate(data[i])
-            return data
-        elif NUMPY:
-            if isinstance(data, numpy.ndarray):
-                data = data.tolist()
-                return cls.validate(data)
-            if isinstance(data, numpy.number):
-                return data.item()
-        raise TypeError("Object of {} is not JSON-serializable".format(type(data)))
+from .validators import JSONFormatValidator
 
 
 class JSONCollection(SyncedCollection):
@@ -90,7 +28,6 @@ class JSONCollection(SyncedCollection):
         self._write_concern = write_concern
         kwargs['name'] = filename
         super().__init__(**kwargs)
-        self._validators.append(JSONFormatValidator)
 
     def _load(self):
         """Load the data from a JSON-file."""
@@ -121,7 +58,10 @@ class JSONCollection(SyncedCollection):
                 file.write(blob)
 
 
-class JSONDict(SyncedAttrDict, JSONCollection):
+JSONCollection.add_validator(JSONFormatValidator())
+
+
+class JSONDict(JSONCollection, SyncedAttrDict):
     """A dict-like mapping interface to a persistent JSON file.
 
     The JSONDict inherits from :class:`~core.collection_api.SyncedCollection`
@@ -167,7 +107,7 @@ class JSONDict(SyncedAttrDict, JSONCollection):
     """
 
 
-class JSONList(SyncedList, JSONCollection):
+class JSONList(JSONCollection, SyncedList):
     """A non-string sequence interface to a persistent JSON file.
 
     The JSONList inherits from :class:`~core.collection_api.SyncedCollection`
--- a/signac/core/collection_json.py
+++ b/signac/core/collection_json.py
@@ -39,10 +39,23 @@ class JSONCollection(SyncedCollection):
             if error.errno == errno.ENOENT:
                 return None
 
+    def _convert_non_str_key_to_str(self, data):
+        if isinstance(data, dict):
+            ret = {}
+            for key, value in data.items():
+                ret[str(key)] = self._convert_non_str_key_to_str(value)
+            return ret
+        elif isinstance(data, list):
+            for i in range(len(data)):
+                data[i] = self._convert_non_str_key_to_str(data[i])
+        return data
+
     def _sync(self):
         """Write the data to json file."""
         data = self.to_base()
-        # Serialize data:
+        # Converting non-string keys to string
+        data = self._convert_non_str_key_to_str(data)
+        # Serialize data
         blob = json.dumps(data).encode()
         # When write_concern flag is set, we write the data into dummy file and then
         # replace that file with original file.
--- a/signac/core/collection_mongodb.py
+++ b/signac/core/collection_mongodb.py
@@ -1,0 +1,135 @@
+# Copyright (c) 2020 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+"""Implements MongoDB-backend.
+
+This implements the MongoDB-backend for SyncedCollection API by
+implementing sync and load methods.
+"""
+from copy import deepcopy
+
+from .synced_collection import SyncedCollection
+from .syncedattrdict import SyncedAttrDict
+from .synced_list import SyncedList
+
+
+class MongoDBCollection(SyncedCollection):
+    """Implement sync and load using a MongoDB backend."""
+
+    backend = __name__  # type: ignore
+
+    def __init__(self, collection=None, **kwargs):
+        import bson  # for InvalidDocument
+
+        self._collection = collection
+        self._errors = bson.errors
+        self._key = type(self).__name__ + '::name'
+        super().__init__(**kwargs)
+
+    def _load(self):
+        """Load the data from a MongoDB."""
+        blob = self._collection.find_one({self._key: self._name})
+        return blob['data'] if blob is not None else None
+
+    def _sync(self):
+        """Write the data from MongoDB."""
+        data = self.to_base()
+        data_to_insert = {self._key: self._name, 'data': data}
+        try:
+            self._collection.replace_one({self._key: self._name}, data_to_insert, True)
+        except self._errors.InvalidDocument as err:
+            raise TypeError(str(err))
+
+    def _pseudo_deepcopy(self):
+        """Return a copy of instance.
+
+        It is a psuedo implementation for `deepcopy` because
+        `pymongo.Collection` does not support `deepcopy` method.
+        """
+        return type(self)(collection=self._collection, name=self._name, data=self.to_base(),
+                          parent=deepcopy(self._parent))
+
+
+class MongoDBDict(MongoDBCollection, SyncedAttrDict):
+    """A dict-like mapping interface to a persistent Mongo-database.
+
+    The MongoDBDict inherits from :class:`~core.collection_api.MongoCollection`
+    and :class:`~core.syncedattrdict.SyncedAttrDict`.
+
+    .. code-block:: python
+
+        doc = MongoDBDict('data')
+        doc['foo'] = "bar"
+        assert doc.foo == doc['foo'] == "bar"
+        assert 'foo' in doc
+        del doc['foo']
+
+    .. code-block:: python
+
+        >>> doc['foo'] = dict(bar=True)
+        >>> doc
+        {'foo': {'bar': True}}
+        >>> doc.foo.bar = False
+        {'foo': {'bar': False}}
+
+    .. warning::
+
+        While the MongoDBDict object behaves like a dictionary, there are
+        important distinctions to remember. In particular, because operations
+        are reflected as changes to an underlying database, copying (even deep
+        copying) a MongoDBDict instance may exhibit unexpected behavior. If a
+        true copy is required, you should use the `to_base()` method to get a
+        dictionary representation, and if necessary construct a new MongoDBDict
+        instance: `new_dict = MongoDBDict(old_dict.to_base())`.
+
+    Parameters
+    ----------
+    collection : object, optional
+        A pymongo.Collection instance.
+    data: mapping, optional
+        The intial data pass to MongoDBDict. Defaults to `dict()`.
+    name: str, optional
+        The name of the  collection (Default value = None).
+    parent: object, optional
+        A parent instance of MongoDBDict (Default value = None).
+    """
+
+
+class MongoDBList(MongoDBCollection, SyncedList):
+    """A non-string sequence interface to a persistent Mongo file.
+
+    The MongoDBList inherits from :class:`~core.synced_collection.SyncedCollection`
+    and :class:`~core.syncedlist.SyncedList`.
+
+    .. code-block:: python
+
+        synced_list = MongoDBList('data')
+        synced_list.append("bar")
+        assert synced_list[0] == "bar"
+        assert len(synced_list) == 1
+        del synced_list[0]
+
+    .. warning::
+
+        While the MongoDBList object behaves like a list, there are
+        important distinctions to remember. In particular, because operations
+        are reflected as changes to an underlying database, copying (even deep
+        copying) a MongoDBList instance may exhibit unexpected behavior. If a
+        true copy is required, you should use the `to_base()` method to get a
+        dictionary representation, and if necessary construct a new MongoDBList
+        instance: `new_list = MongoDBList(old_list.to_base())`.
+
+    Parameters
+    ----------
+    collection : object, optional
+        A pymongo.Collection instance (Default value = None).
+    data: non-str Sequence, optional
+        The intial data pass to MongoDBList. Defaults to `list()`.
+    name: str, optional
+        The name of the  collection (Default value = None).
+    parent: object, optional
+        A parent instance of MongoDBList (Default value = None).
+    """
+
+
+SyncedCollection.register(MongoDBDict, MongoDBList)
--- a/signac/core/collection_mongodb.py
+++ b/signac/core/collection_mongodb.py
@@ -53,7 +53,7 @@ class MongoDBCollection(SyncedCollection):
 class MongoDBDict(MongoDBCollection, SyncedAttrDict):
     """A dict-like mapping interface to a persistent Mongo-database.
 
-    The MongoDBDict inherits from :class:`~core.collection_api.MongoCollection`
+    The MongoDBDict inherits from :class:`~core.synced_collection.MongoCollection`
     and :class:`~core.syncedattrdict.SyncedAttrDict`.
 
     .. code-block:: python
--- a/signac/core/collection_mongodb.py
+++ b/signac/core/collection_mongodb.py
@@ -11,6 +11,7 @@ from copy import deepcopy
 from .synced_collection import SyncedCollection
 from .syncedattrdict import SyncedAttrDict
 from .synced_list import SyncedList
+from .collection_json import JSONFormatValidator
 
 
 class MongoDBCollection(SyncedCollection):
@@ -19,12 +20,10 @@ class MongoDBCollection(SyncedCollection):
     backend = __name__  # type: ignore
 
     def __init__(self, collection=None, **kwargs):
-        import bson  # for InvalidDocument
-
         self._collection = collection
-        self._errors = bson.errors
         self._key = type(self).__name__ + '::name'
         super().__init__(**kwargs)
+        self._validators.append(JSONFormatValidator)
 
     def _load(self):
         """Load the data from a MongoDB."""
@@ -35,10 +34,7 @@ class MongoDBCollection(SyncedCollection):
         """Write the data from MongoDB."""
         data = self.to_base()
         data_to_insert = {self._key: self._name, 'data': data}
-        try:
-            self._collection.replace_one({self._key: self._name}, data_to_insert, True)
-        except self._errors.InvalidDocument as err:
-            raise TypeError(str(err))
+        self._collection.replace_one({self._key: self._name}, data_to_insert, True)
 
     def _pseudo_deepcopy(self):
         """Return a copy of instance.
--- a/signac/core/collection_mongodb.py
+++ b/signac/core/collection_mongodb.py
@@ -16,7 +16,7 @@ from .synced_list import SyncedList
 class MongoDBCollection(SyncedCollection):
     """Implement sync and load using a MongoDB backend."""
 
-    backend = __name__  # type: ignore
+    _backend = __name__  # type: ignore
 
     def __init__(self, collection=None, **kwargs):
         import bson  # for InvalidDocument error
--- a/signac/core/collection_mongodb.py
+++ b/signac/core/collection_mongodb.py
@@ -11,7 +11,6 @@ from copy import deepcopy
 from .synced_collection import SyncedCollection
 from .syncedattrdict import SyncedAttrDict
 from .synced_list import SyncedList
-from .collection_json import JSONFormatValidator
 
 
 class MongoDBCollection(SyncedCollection):
@@ -20,10 +19,12 @@ class MongoDBCollection(SyncedCollection):
     backend = __name__  # type: ignore
 
     def __init__(self, collection=None, **kwargs):
+        import bson  # for InvalidDocument error
+
         self._collection = collection
+        self._errors = bson.errors
         self._key = type(self).__name__ + '::name'
         super().__init__(**kwargs)
-        self._validators.append(JSONFormatValidator)
 
     def _load(self):
         """Load the data from a MongoDB."""
@@ -34,7 +35,10 @@ class MongoDBCollection(SyncedCollection):
         """Write the data from MongoDB."""
         data = self.to_base()
         data_to_insert = {self._key: self._name, 'data': data}
-        self._collection.replace_one({self._key: self._name}, data_to_insert, True)
+        try:
+            self._collection.replace_one({self._key: self._name}, data_to_insert, True)
+        except self._errors.InvalidDocument as err:
+            raise TypeError(str(err))
 
     def _pseudo_deepcopy(self):
         """Return a copy of instance.
--- a/signac/core/synced_collection.py
+++ b/signac/core/synced_collection.py
@@ -30,7 +30,7 @@ class SyncedCollection(Collection):
     """
 
     backend = None
-    _validators = list()
+    _validators: List[object] = list()  # list of callable objects
 
     def __init__(self, name=None, parent=None):
         self._data = None
--- a/signac/core/synced_collection.py
+++ b/signac/core/synced_collection.py
@@ -9,7 +9,7 @@ backend with different data-structures or vice-versa. It declares as abstract
 methods the methods that must be implemented by any subclass to match the API.
 """
 
-from typing import List, Any
+from typing import List, Callable
 from contextlib import contextmanager
 from abc import abstractmethod
 from collections import defaultdict
@@ -31,7 +31,7 @@ class SyncedCollection(Collection):
     """
 
     _backend = None
-    _validators: List[Any] = list()  # list of callable objects
+    _validators: List[Callable] = []
 
     def __init__(self, name=None, parent=None):
         self._data = None
@@ -51,7 +51,7 @@ class SyncedCollection(Collection):
         Every subclass must have a separate list so that distinct sets of validators can
         be registered to each of them.
         """
-        cls._validators = list()
+        cls._validators = []
 
     @classmethod
     def register(cls, *args):
@@ -79,11 +79,9 @@ class SyncedCollection(Collection):
         """Return the list of validators applied to the instance."""
         validators = []
         # Classes inherit the validators of their parent classes.
-        for _cls in type(self).__mro__:
-            if hasattr(_cls, '_validators'):
-                for validator in _cls._validators:
-                    if validator not in validators:
-                        validators.append(validator)
+        for base_cls in type(self).__mro__:
+            if hasattr(base_cls, '_validators'):
+                validators.extend([v for v in base_cls._validators if v not in validators])
         return validators
 
     @classmethod
@@ -93,11 +91,9 @@ class SyncedCollection(Collection):
         Parameters
         ----------
         \*args
-            Validator to register
+            Validator(s) to register.
         """
-        for validator in args:
-            if validator not in cls._validators:
-                cls._validators.append(validator)
+        cls._validators.extend([v for v in args if v not in cls._validators])
 
     @classmethod
     def from_base(cls, data, backend=None, **kwargs):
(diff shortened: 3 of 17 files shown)


signac/core/collection_json.py
```diff
@@ -27,6 +90,7 @@ def __init__(self, filename=None, write_concern=False, **kwargs):
         self._write_concern = write_concern
         kwargs['name'] = filename
         super().__init__(**kwargs)
+        self._validators.append(JSONFormatValidator)
```
vyasr, line 93: @vishav1771 as per my comment on the docstring of the `JSONFormatValidator` class, that format validator is not specific to just this collection but could be used for any collection. This also means that `SyncedCollection` should expose an API to register validators to specific subclasses (like `JSONCollection`). However, I don't think that the validation should be controlled by our infrastructure, i.e. a user of the `SyncedCollection` family of classes should be able to freely mix and match validators and backends. I think the best way to enable this is for `SyncedCollection` to define a method `add_validator` or accept validators as a constructor argument. So the prototype in signac I'm imagining would look something like

```
# Need to import from wherever these actually live.
from synced_collection... import JSONDict, JSONFormatValidator

class Job:
    def __init__(self, ...):
        ...
        # Option 1
        self._sp = JSONDict(..., validator=JSONFormatValidator)
        # Option 2
        self._sp = JSONDict(...)
        self._sp.add_validator(JSONFormatValidator)

```

An alternative would be to do this on a class level, i.e. to define `add_validator` as a `classmethod` of `SyncedCollection`. This approach introduces complexities such as whether you should inherit validators from parent classes. For instance, if you do `JSONCollection.add_validator(...)` and `JSONDict.add_validator(...)`, I would probably expect `JSONDict` to use the validators added to `JSONCollection` irrespective of the order in which those two calls occurred. On the plus side, this approach is more concise since you don't have to reassign the validator to every instance. @csadorf thoughts?

signac/core/collection_json.py
```diff
@@ -57,7 +121,7 @@ def _sync(self):
                 file.write(blob)
 
 
-class JSONDict(JSONCollection, SyncedAttrDict):
+class JSONDict(SyncedAttrDict, JSONCollection):
```
vyasr, line 124: Is there a reason this inheritance order needed to change? The design of the classes was intended to avoid any collisions that might require nonintuitive method resolution order.

signac/core/collection_mongodb.py
```diff
@@ -19,12 +20,10 @@ class MongoDBCollection(SyncedCollection):
     backend = __name__  # type: ignore
 
     def __init__(self, collection=None, **kwargs):
-        import bson  # for InvalidDocument
-
         self._collection = collection
-        self._errors = bson.errors
         self._key = type(self).__name__ + '::name'
         super().__init__(**kwargs)
+        self._validators.append(JSONFormatValidator)
```
vyasr, line 26: Same as above, shouldn't have to hardcode.

signac/core/collection_mongodb.py
```diff
@@ -35,10 +34,7 @@ def _sync(self):
         """Write the data from MongoDB."""
         data = self.to_base()
         data_to_insert = {self._key: self._name, 'data': data}
-        try:
-            self._collection.replace_one({self._key: self._name}, data_to_insert, True)
-        except self._errors.InvalidDocument as err:
-            raise TypeError(str(err))
```
vyasr, line 41: Was the try-except just performing the JSON validation before?

signac/core/syncedattrdict.py
```diff
@@ -17,10 +17,48 @@
...
+                        "SyncedAttrDict keys must be str, int, bool or None, not {}"
+                        .format(type(key).__name__))
+        return data
+
+
 class SyncedAttrDict(SyncedCollection, MutableMapping):
     """Implement the dict data structure along with values access through attributes named as keys.
 
-    The SyncedAttrDict inherits from :class:`~core.collection_api.SyncedCollection`
+    The SyncedAttrDict inherits from :class:`~core.synced_colletion.SyncedCollection`
```
vyasr, line 61:
```suggestion
    The SyncedAttrDict inherits from :class:`~core.synced_collection.SyncedCollection`
```

signac/core/collection_json.py
vyasr: Having read the whole PR, I assume the reason has to do with the validators associated with the `SyncedAttrDict` class. I think this change can be reverted, see my comment below.

signac/core/syncedattrdict.py
```diff
@@ -17,10 +17,48 @@
 from ..errors import KeyTypeError
 
 
+class AttrDictFormatValidator:
```
vyasr, line 20: I don't think this class needs to exist. There's no case (at least right now) where the `JSONFormatValidator` wouldn't be what we wanted to use in all cases. Also in practice (although we don't have to impose this restriction anywhere in the API) I expect that validation is something that will be used for compatibility with backend specifications, not for filtering specific data types (although that's certainly possible). Also, as I said in a previous comment the validators and collections should be more independent, so the `SyncedAttrDict` doesn't need any validators hardcoded into its initializer anyway.

signac/core/collection_mongodb.py
vishav1771: This was changing the exception type (`InvalidDocument` to `TypeError`).  Now, this exception is raised by validator .

signac/core/syncedattrdict.py
vishav1771: Sorry, I will change that.

signac/core/syncedattrdict.py
vishav1771: I will change this to the old implementation (`_valiadate_key`).

signac/core/syncedattrdict.py
vyasr: No problem, you can directly apply this suggestion.

signac/core/syncedattrdict.py
vyasr: I'm not suggesting that, I'm suggesting to use the JSONFormatValidator here. I don't see any reason that you can't. The JSONFormatValidator is a superset of this one that also works for non-mapping data types, but is otherwise doing the same thing, right?

signac/core/collection_json.py
vyasr: I need to walk back one part of this: certain backends may _require_ some validation, in which case those validators should be applied automatically. In particular, all `JSONCollection` instances need to have the `JSONFormatValidator` applied (even outside the signac context), whereas in signac we plan to apply that validator irrespective of backend because we require the data to be valid JSON no matter what backend is used. So you probably still need to add the validator here.

Since there need to be some default validators for certain classes, that might lead us to prefer the `classmethod` approach I described, but it also wouldn't be hard to implement either of the instance-based approaches and simply have a separate class constant that just stores the default validators used for instances of a class. As a result, I think both approaches are still feasible even with this requirement.

signac/core/collection_mongodb.py
vishav1771: @vyasr What should we done regarding this? Should I revert the change? (as this was removed because I added `JSONFormatValidator` in `MongoDBCollection`)

signac/core/synced_collection.py
```diff
@@ -58,6 +60,17 @@ def register(cls, *args):
         for _cls in args:
             cls.registry[_cls.backend].append(_cls)
 
+    @classmethod
+    def add_validator(cls, *args):
```
vyasr, line 64: Since you chose the classmethod approach, we need to consider how to handle the inheritance of validators. For example, have you tested whether a `JSONDict` will throw an error if you try to use a dotted key? With the current implementation I don't think that it will, because those validators only belong to the parent class but won't be inherited by the child.

signac/core/synced_collection.py
```diff
@@ -30,6 +31,7 @@ class SyncedCollection(Collection):
     """
 
     backend = None
+    _validators: List[object] = list()  # list of callable objects
```
vyasr, line 34: Have we been typing in other places in the recent PRs?

signac/core/validators.py
```diff
@@ -0,0 +1,105 @@
...
+
+from ..errors import InvalidKeyError
+from ..errors import KeyTypeError
+from .synced_collection import NUMPY
+
+if NUMPY:
+    import numpy
+
+
+class NoDotInKey:
```
vyasr, line 17: I don't think either of the validators in this file need to be classes, I would implement them as functions.

signac/core/synced_collection.py
vishav1771: No, I got an [error](https://app.circleci.com/pipelines/github/glotzerlab/signac/1300/workflows/64d8599a-d2c3-4f0e-b5ad-63dbb194d158/jobs/8088) while testing on CI for this case.

signac/core/synced_collection.py
vishav1771: Yes, I have tested it. We also have a test for that. https://github.com/glotzerlab/signac/blob/1906db8dc0f91c3950de06ad9e370c46275b07e9/tests/test_synced_collection.py#L344-L346

vyasr: Ah OK I see the issue with the current design. The `_validators` list is defined for the `SyncedCollection` class, so when any subclass (for instance, `JSONCollection`) calls `add_validator`, it's actually just extending the same list. As a result, adding a validator to any of the classes adds it to all of them. This example illustrates the problem:

```
class A:
    _validators = list()

    @classmethod
    def add(cls, *val):
        cls._validators.extend(val)


class B(A):
    pass


class C(B):
    pass


A.add(1)
B.add(2)
C.add(3)

print(A._validators)
print(B._validators)
print(C._validators)

print(A._validators is B._validators)
print(A._validators is C._validators)

# Output 
[1, 2, 3]
[1, 2, 3]
[1, 2, 3]
True
True
```

The easiest solution that maintains the class-based approach to validators and still ensures that all instances of `SyncedCollection` get their own `_validators` list is using the `__init_subclass__` hook:

```
class A:
    @classmethod
    def __init_subclass__(cls):
        cls._validators = list()

    @classmethod
    def add(cls, *val):
        cls._validators.extend(val)


class B(A):
    pass


class C(B):
    pass


print(hasattr(A, '_validators'))
print(hasattr(B, '_validators'))
print(hasattr(C, '_validators'))
print(B._validators is C._validators)

# Output 
False
True
True
False
```

Note that this method is called whenever subclasses are created, but not for the creation of the original class, so you will still need to define the `_validators` variable in the `SyncedCollection` class. *Please document why we are using this hook when you write this method.*

signac/core/synced_collection.py
vyasr: Got it, I don't recall when exactly we added mypy validation but makes sense.

vishav1771:
@vyasr We are using JSON encoding in `Zarr` and `Redis`  backends. Should I add `JSONFormatValidator` to their default validators?

signac/core/synced_collection.py
vyasr: However, this reintroduces my original question in [this comment](https://github.com/glotzerlab/signac/pull/378#discussion_r475798622): if we do the class-based approach, we need to handle inheritance of validators from parent classes. @bdice I noticed your response here, what are your thoughts?  I think it would be very confusing to register a `JSONFormatValidator` to the `JSONCollection`, and then have to re-register it to the `JSONDict` and `JSONList` classes. On the other hand, as a user of this code, how would you expect this to behave if you instantiated your own synced collections? Would you expect that modifying a parent of a class changed the validators of that class? I think that is the most consistent behavior we could define.

Also, from an implementation point of view, the only way I can think of to enable this is to define `SyncedCollection.validators` as a property that walks the MRO tree and finds the validators registered to all parents. Is there a simpler alternative?

vyasr:
> @vyasr We are using JSON encoding in `Zarr` and `Redis` backends. Should I add `JSONFormatValidator` to their default validators?

No, the `SyncedCollection` classes should only have default validation if the back end itself requires that validation to work. All of these classes should be developed to work seamlessly with `signac`, but I think we should development them as a completely independent module to the greatest degree possible. The requirement for `Zarr` and `Redis` backends to have JSON formatting is not a requirement of these types of collections, but rather a requirement of `signac`, which requires valid JSON no matter what back end is being used. Therefore, `signac` should take care of registering the appropriate additional validators for all back ends.

signac/core/synced_collection.py
bdice: @vyasr I read the comments above and in the linked thread. I like the class-based approach you suggested. I would like to register validators with `JSONCollection` and have those apply automatically to `JSONList` and `JSONDict`.

I haven't tried this myself so I don't know the pros/cons of the functionality of `__init_subclass__` (which I wasn't aware of) vs. walking the MRO entries. (Also in my understanding, it's not really an "MRO tree" due to C3 linearization, but that's subtle.)

signac/core/validators.py
```diff
@@ -0,0 +1,99 @@
...
+if NUMPY:
+    import numpy
+
+
+def NoDotInKey(data):
+    """Raise an exception if there is dot in key.
+
+    Parameters
+    ----------
+    data:
```
bdice, line 22: See #362 - I think the colon is unnecessary unless a data type is specified.
```suggestion
    data
```

signac/core/validators.py
```diff
@@ -0,0 +1,99 @@
...
+        Data to validate.
+
+    Returns
+    -------
+    data
```
bdice, line 27:
```suggestion
    Validated data.
```

signac/core/validators.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Raises
+    ------
+    KeyTypeError:
```
bdice, line 31:
```suggestion
    KeyTypeError
```

signac/core/validators.py
```diff
@@ -0,0 +1,99 @@
...
+                    .format(type(key).__name__))
+    return data
+
+
+def JSONFormatValidator(data):
+    """Implement the validation for JSON serializable data
+
+    Parameters
+    ----------
+    data:
```
bdice, line 56:
```suggestion
    data
```

signac/core/validators.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Raises
+    ------
+    KeyTypeError:
+        If keys of mapping have unsupported data type.
+    TypeError
+        If data type is not supported.
+    """
+
+    if isinstance(data, (str, int, float, bool, bytes, type(None))):
```
bdice, line 71: I don't think `bytes` is JSON-serializable -- this could be a bug. I wonder if this is a holdover from Python 2, when `str` and `bytes` were handled differently?

```pytb
>>> some_bytes = b'123'
>>> import json
>>> json.dumps(some_bytes)
  File "/home/bdice/miniconda3/envs/dice/lib/python3.8/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
    return _iterencode(o, 0)
  File "/home/bdice/miniconda3/envs/dice/lib/python3.8/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type bytes is not JSON serializable
>
```

signac/core/validators.py
```diff
@@ -0,0 +1,99 @@
...
+                    "Keys must be str, int, bool or None, not {}".format(type(key).__name__))
+            new_value = JSONFormatValidator(value)
+            ret[new_key] = new_value
+        return ret
+    elif isinstance(data, Sequence):
+        for i in range(len(data)):
+            data[i] = JSONFormatValidator(data[i])
+        return data
+    elif NUMPY:
+        if isinstance(data, numpy.ndarray):
```
bdice, line 94: I think we can support any array-like object here with a simple change, but I don't want to complicate this PR. I created a separate issue #380 that can be handled later (not in scope for GSoC).

bdice:
Also @vishav1771 it looks like the Zarr backend is failing some tests, because of something with JSON encoding of `str` and `int` keys. We are planning to disable the automatic conversion of `int` keys to `str` in signac 2.0 but I would be fine with disabling support for that with non-JSON backends and skipping some tests of that behavior (rather than enabling the automatic conversion for other backends). Be sure to add comments in the test files explaining the reason if you skip tests.

signac/core/synced_collection.py
```diff
@@ -133,6 +146,12 @@ def load(self):
             else:
                 self._parent.load()
 
+    def _validate(self, data):
+        """Validate the input data"""
+        for validator in self._validators:
+            data = validator(data)
```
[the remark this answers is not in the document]
vyasr: I discuss this exact question in [this comment](https://github.com/glotzerlab/signac/pull/378#discussion_r475785546).

signac/core/collection_json.py
```diff
@@ -11,10 +11,73 @@
...
+
+if NUMPY:
+    import numpy
+
+
+class JSONFormatValidator:
+    """Implement the validation for :class:`JSONCollection`."""
+
+    @classmethod
+    def validate(cls, data):
```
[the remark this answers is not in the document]
vyasr: @bdice now that the validators are just arbitrary callables and not classes, I think we should perhaps move

signac/core/synced_collection.py
vyasr: The two aren't mutually exclusive, but rather one implies the other. If validators are stored per class, then you _have_ to traverse the MRO (you're right, it's not a tree with new-style classes but I recycled old terminology there) to find the validators added to parent classes. The only way I can see to avoid this is to either 1) not share validators with subclasses, which I don't think makes sense, or 2) to use the instance-based where validators are added to every instance of a class. Option 2 is simpler from an implementation perspective than the class-based approach, but I think it's a poorer design since validation really should apply to all instances of the class, so I would prefer the class approach. What I'm asking is whether you can think of any way to avoid having to traverse the list of parent classes to find this; if not, then I think we just need to implement a `SyncedCollection.validators` `classmethod` property that collects this in the same we handle operations in the FlowProject in signac-flow.

signac/core/validators.py
vyasr: Good catch. That certainly seems likely, and I'm fine just disallowing bytes.

signac/core/synced_collection.py
vishav1771: @vyasr I don't think we can use `classmethod` as property. I have implemented it in `_validate`. https://github.com/glotzerlab/signac/blob/39d630d5032bed09d7d08b412e726ee5067a4af8/signac/core/synced_collection.py#L160-L168

signac/core/collection_json.py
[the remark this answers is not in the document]
vishav1771: I think we can remove converting NumPy arrays to lists from validators as `SyncedList` handle NumPy arrays. I moved that to validation because I think the serializing of arrays is done to store to JSON backend. In the current implementation, we handle (b) internally so can be removed from validation.

tests/test_synced_collection.py
```diff
@@ -345,14 +345,25 @@ def test_keys_with_dots(self, synced_dict):
...
             pass
-        for key in ('key', MyStr('key'), 0, None, True):
+        for key in ('key', MyStr('key')):
             synced_dict[key] = testdata
-            assert str(key) in synced_dict
-            assert synced_dict[str(key)] == testdata
+            assert key in synced_dict
+            assert synced_dict[key] == testdata
+
+    def test_keys_non_str_valid_type(self, synced_dict, testdata):
```
vishav1771, line 357: @bdice In #336, we decided not to add backend specific tests in this class. Should I remove this test or not?

signac/core/collection_json.py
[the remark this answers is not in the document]
vishav1771: Done.

tests/test_synced_collection.py
bdice: @vishav1771 If possible, I would only implement this test for the JSON backend (instead of defining it for all backends and skipping on zarr/redis). I would add a comment to the test that says it should be removed in signac 2.0 and mention issue #316.

signac/core/synced_collection.py
bdice: @vishav1771 I agree with your statement that `@property` and `@classmethod` don't work together. It's possible to create a custom `@classproperty` decorator but that's too complicated in my opinion. This is an alternative using `@property` with `type(self)` instead of `cls`:

```python
class AwesomeClass:
    @property
    def name(self):
        return type(self).__name__

awesome = AwesomeClass()
print(awesome.name)
```

signac/core/collection_json.py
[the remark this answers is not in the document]
bdice: @vyasr @vishav1771 I am in favor of **1) change validators to be either raise exceptions or be no-ops.** We shouldn't make forward-thinking design decisions based on deprecated behavior. As noted above, that would require us to break deprecated behavior if we just dropped the non-str keys. I would consider inserting a "workaround" in `JSONCollection`, which would be separate from (perhaps before?) validation. Its implementation would be similar to the previous approach to handling this. There should be a clear comment to remove the workaround in signac 2.0, since it only affects the JSON backend with specific key types.

signac/core/synced_collection.py
```diff
@@ -41,34 +43,56 @@ def __init__(self, name=None, parent=None):
                 "Illegal argument combination, one of the two arguments, "
                 "parent or name must be None, but not both.")
 
+    @classmethod
+    def __init_subclass__(cls):
+        """Add  `_validator` attribute to every subsclass.
```
bdice, line 48:
```suggestion
        """Add  `_validator` attribute to every subclass.
```

signac/core/synced_collection.py
```diff
@@ -41,34 +43,56 @@ def __init__(self, name=None, parent=None):
...
+
+        Every subclass contain list of validators that are applied
+        to the input of the instance of subclass.
```
bdice, line 51:
```suggestion
        Subclasses contain a list of validators that are applied to collection input data.
```

signac/core/synced_collection.py
```diff
@@ -133,6 +157,17 @@ def load(self):
             else:
                 self._parent.load()
 
+    @classmethod
+    def _validate(cls, data):
+        """Validate the input data."""
+        # Validate for every parent class which have _validators
+        for _cls in cls.mro():
```
bdice, line 164: I _think_ we want to use `cls.__mro__` instead, because it's the cached result of `cls.mro()` at the time the class is instantiated (not an _instance_ of the class, in my understanding). (The `mro` method is defined by `type`.) [https://docs.python.org/3/library/stdtypes.html#class.\_\_mro\_\_](https://docs.python.org/3/library/stdtypes.html#class.__mro__)
```suggestion
        for _cls in cls.__mro__:
```

signac/core/validators.py
```diff
@@ -0,0 +1,96 @@
+# Copyright (c) 2017 The Regents of the University of Michigan
```
bdice, line 1: New file, new year. 😄
```suggestion
# Copyright (c) 2020 The Regents of the University of Michigan
```

signac/core/validators.py
```diff
@@ -0,0 +1,96 @@
...
+from ..errors import InvalidKeyError
+from ..errors import KeyTypeError
+from .synced_collection import NUMPY
+
+if NUMPY:
+    import numpy
+
+
+def NoDotInKey(data):
+    """Raise an exception if there is dot in key.
```
bdice, line 18:
```suggestion
    """Raise an exception if there is a dot (``.``) in a mapping's key.
```

signac/core/validators.py
```diff
@@ -0,0 +1,96 @@
...
+                        "Mapping keys may not contain dots ('.'): {}".format(key))
+            elif not isinstance(key, VALID_KEY_TYPES):
+                raise KeyTypeError(
+                    "Mapping keys must be str, int, bool or None, not {}"
+                    .format(type(key).__name__))
+    return data
+
+
+def JSONFormatValidator(data):
+    """Implement the validation for JSON serializable data
```
bdice, line 52:
```suggestion
    """Implement the validation for JSON serializable data.
```

signac/core/validators.py
```diff
@@ -0,0 +1,96 @@
...
+        If key is not supported.
+    """
+    VALID_KEY_TYPES = (str, int, bool, type(None))
+
+    if isinstance(data, Mapping):
+        for key in data.keys():
+            if isinstance(key, str):
+                if '.' in key:
+                    raise InvalidKeyError(
+                        "Mapping keys may not contain dots ('.'): {}".format(key))
```
bdice, line 43: Use f-strings.
```suggestion
                        f"Mapping keys may not contain dots ('.'): {key}"
```

signac/core/validators.py
```diff
@@ -0,0 +1,96 @@
...
+            elif not isinstance(key, VALID_KEY_TYPES):
+                raise KeyTypeError(
+                    "Mapping keys must be str, int, bool or None, not {}"
+                    .format(type(key).__name__))
```
bdice, line 47:
```suggestion
                    f"Mapping keys must be str, int, bool or None, not {type(key).__name__}")
```

signac/core/validators.py
```diff
@@ -0,0 +1,96 @@
...
+
+    if isinstance(data, (str, int, float, bool, type(None))):
+        return data
+    elif isinstance(data, Mapping):
+        ret = {}
+        for key, value in data.items():
+            if isinstance(key, (int, bool, type(None))):
+                warnings.warn(
+                    "Use of {} as key is deprecated and will be removed in version 2.0"
+                    .format(type(key)), DeprecationWarning)
```
bdice, line 79:
```suggestion
                    f"Use of {type(key)} as key is deprecated and will be removed in version 2.0",
                    DeprecationWarning)
```

signac/core/validators.py
```diff
@@ -0,0 +1,96 @@
...
+                new_key = str(key)
+            elif isinstance(key, str):
+                new_key = key
+            else:
+                raise KeyTypeError(
+                    "Keys must be str, int, bool or None, not {}".format(type(key).__name__))
```
bdice, line 85:
```suggestion
                    f"Keys must be str, int, bool or None, not {type(key).__name__}")
```

signac/core/validators.py
```diff
@@ -0,0 +1,96 @@
...
+            ret[new_key] = new_value
+        return ret
+    elif isinstance(data, Sequence):
+        for i in range(len(data)):
+            data[i] = JSONFormatValidator(data[i])
+        return data
+    elif NUMPY:
+        if isinstance(data, (numpy.ndarray, numpy.number)):
+            return data
+    raise TypeError("Object of {} is not JSON-serializable".format(type(data)))
```
bdice, line 96:
```suggestion
    raise TypeError(f"Object of {type(data)} is not JSON-serializable")
```

signac/core/synced_list.py
```diff
@@ -136,6 +139,7 @@ def reset(self, data=None):
 
     def __setitem__(self, key, value):
         self.load()
+        value = self._validate(value)
```
bdice, line 142: Should the item be validated before loading and suspending synchronization? Seems like it should be the first step. Also for all the other methods below.

signac/core/validators.py
```diff
@@ -0,0 +1,95 @@
...
+import warnings
+from collections.abc import Mapping
+from collections.abc import Sequence
+
+from ..errors import InvalidKeyError
+from ..errors import KeyTypeError
+from .synced_collection import NUMPY
+
+if NUMPY:
+    import numpy
```
bdice, line 14: I would rather copy the try/except snippet into this file, rather than import `synced_collection`.
```suggestion

try:
    import numpy
    NUMPY = True
except ImportError:
    NUMPY = False
```

tests/test_synced_collection.py
```diff
@@ -345,14 +345,24 @@ def test_keys_with_dots(self, synced_dict):
...
-            assert str(key) in synced_dict
-            assert synced_dict[str(key)] == testdata
+            assert key in synced_dict
+            assert synced_dict[key] == testdata
+
+    # The following test tests the support for non-str keys
+    # for JSON backend which will be removed in version 2.0.
+    # see issue #316.
+    def test_keys_non_str_valid_type(self, synced_dict, testdata):
+        if synced_dict.backend == 'signac.core.collection_json':
```
bdice, line 361: Why does the `synced_dict.backend` property exist? A type check seems more appropriate, especially since `JSONDict` has already been imported.
```suggestion
        if isinstance(synced_dict, JSONDict):
```

bdice: Upon further investigation, I'm almost positive that `backend` should be a private property, and it may be possible to eliminate it altogether. I think you can handle the registry using the knowledge you have about synced collection classes without needing to define a string for the backend name.

vishav1771: `synced_dict.backend` is not a property. It is a class attribute. Will change it to private. But I don't think we can remove this. This also used to find type of child class (JSONDict child should be JSONDict and JSONList ).

signac/core/synced_collection.py
```diff
@@ -41,42 +43,75 @@ def __init__(self, name=None, parent=None):
...
             cls.registry = defaultdict(list)
         for _cls in args:
-            cls.registry[_cls.backend].append(_cls)
+            cls.registry[_cls._backend].append(_cls)
+
+    @property
+    def validators(self):
+        """Return the list of validators applied to the instance."""
+        validators = []
+        # Instance also inherite the validators of parent class.
```
vyasr, line 75:
```suggestion
        # Classes inherit the validators of their parent classes.
```

signac/core/validators.py
```diff
@@ -0,0 +1,97 @@
...
+    """Raise an exception if there is a dot (``.``) in a mapping's key.
+
+    Parameters
+    ----------
+    data
+        Data to validate.
+
+    Returns
+    -------
+    Validated data.
```
vyasr, line 29: You need to change all validators to not return the data anymore. I didn't comment on this in the other files since it's everywhere and will be easier for you to just go through and delete all instances of `data = validator.validate(data)` or whatever. Removing the returns from these two validators should automatically cause your tests to fail until you fix this everywhere, so that will be easy enough to check.

signac/core/synced_collection.py
```diff
@@ -29,7 +30,8 @@ class SyncedCollection(Collection):
     in the underlying backend. The backend name wil be same as the module name.
     """
 
-    backend = None
+    _backend = None
+    _validators: List[Any] = list()  # list of callable objects
```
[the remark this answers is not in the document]
vishav1771: Thanks for the information, Will use it.

signac/core/synced_collection.py
```diff
@@ -41,42 +43,81 @@ def __init__(self, name=None, parent=None):
...
         for _cls in args:
-            cls.registry[_cls.backend].append(_cls)
+            cls.registry[_cls._backend].append(_cls)
+
+    @property
+    def validators(self):
+        """Return the list of validators applied to the instance."""
+        validators = []
+        # Classes inherit the validators of their parent classes.
+        for _cls in type(self).__mro__:
```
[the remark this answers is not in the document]
vishav1771: Thank you for the information. I will keep in mind these points while using variable names in the future.

signac/core/syncedattrdict.py
```diff
@@ -147,13 +133,11 @@ def reset(self, data=None):
         """
         if data is None:
             data = {}
-
         if isinstance(data, Mapping):
-            self.load()
```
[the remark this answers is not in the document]
vishav1771: This `self.load()` is removed from function `reset`. We create a new instance for `self._data` regardless of data in the disk in `reset` so we do not need `self._data` to be synchronized with disk.

signac/core/validators.py
```diff
@@ -0,0 +1,83 @@
...
+        Data to validate.
+
+    Raises
+    ------
+    KeyTypeError
+        If keys have unsupported data type.
+    InvalidKeyError
+        If key is not supported.
+    """
+    VALID_KEY_TYPES = (str, int, bool, type(None))
```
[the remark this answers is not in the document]
vishav1771: Validation is done before we pass the data to the backend. So, I think we need to support all the non-string key types here.
> (we aren't going to support non-string keys in new backends)

I think we are only removing support for non-string keys in JSON backend.

signac/core/syncedattrdict.py
```diff
@@ -194,17 +178,18 @@ def clear(self):
         self.sync()
 
     def update(self, mapping):
```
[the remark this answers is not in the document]
vishav1771: I have also found this on link
>update() accepts either another dictionary object or an iterable of key/value pairs (as tuples or other iterables of length two). If keyword arguments are specified, the dictionary is then updated with those key/value pairs: d.update(red=1, blue=2).

Currently, we support only mapping as input. Should we add iterable of key/value pairs or `kwargs` also?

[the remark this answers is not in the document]
bdice: @vishav1771 I'll make a code suggestion.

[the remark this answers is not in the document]
bdice: @vishav1771 I can't make a code suggestion inline because it would alter deleted lines, but here's the snippet I would try. You may need to add `from collections.abc import Mapping`.

```python
    def update(self, other=None, **kwargs):
        self.load()
        if other is not None:
            # Convert sequence of key, value pairs to dict before validation
            if not isinstance(other, Mapping):
                other = dict(other)
            self._validate(other)
        if kwargs:
            self._validate(kwargs)
        with self._suspend_sync():
            for key, value in other.items():
                self._data[key] = self.from_base(data=value, parent=self)
            for key, value in kwargs.items():
                self._data[key] = self.from_base(data=value, parent=self)
        self.sync()
```

vishav1771:
@bdice  @vyasr  I have done the changes except changing the `_validator` to `set`.

signac/core/validators.py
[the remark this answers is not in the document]
vyasr: @bdice @vishav1771 this change is not necessary and the current implementation is in line with the design that I proposed. The `SyncedCollection` layer is to completely support the use cases in signac while remaining entirely independent of signac. Any validator can be registered with any class prior to its use, which that signac can (and should) apply the `JSON_format_validator` to _all_ back ends. However, a `SyncedAttrDict` used outside the context of signac (which is completely feasible) still needs to avoid having dots in keys because that breaks the attribute-based access to stored items, which is why this validator is shipped with the framework and registered by default to this class. In the context of signac, we'll need both validators.

One thing that we haven't discussed but we might need to at some point is, if applying both validators becomes a noticeable performance drag we may need to add a method to clear validators and write a single validator that performs both functions. However, at the moment this would be premature optimization.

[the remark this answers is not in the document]
vyasr: I came here straight from a comment on Slack and I think I missed the point here. @vishav1771 I agree with @bdice that we shouldn't support non-str keys. I actually think we should remove the automatic casting to str altogether. That's deprecated behavior, and I think when synced collections get merged we should go ahead and break it, with the intention of releasing synced collections as part of the 2.x line of signac.

signac/core/validators.py
```diff
@@ -0,0 +1,91 @@
...
+        If keys of mapping have unsupported data type.
+    TypeError
+        If data type is not supported.
+    """
+
+    if isinstance(data, (str, int, float, bool, type(None))):
+        return
+    if isinstance(data, Mapping):
+        for key, value in data.items():
+            if not isinstance(key, (str, int, bool, type(None))):
```
bdice, line 77:
```suggestion
            # Support for non-str keys will be removed in version 2.0.
            # See issue #316.
            if not isinstance(key, (str, int, bool, type(None))):
```

signac/core/validators.py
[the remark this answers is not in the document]
bdice: @vyasr Let's drop non-str keys in a follow-up PR. I don't want to introduce any further design changes here right now. This has been through many iterations and I'm confident in the current behavior doing what it intends to do correctly (even though it's not what we want for signac 2.0).

signac/core/validators.py
```diff
@@ -0,0 +1,91 @@
...
+from ..errors import KeyTypeError
+
+try:
+    import numpy
+    NUMPY = True
+except ImportError:
+    NUMPY = False
+
+
+def no_dot_in_key(data):
```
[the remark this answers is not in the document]
vyasr: Actually I intended for the validators should be exposed as part of the API. That would allow users to freely apply validators in the event that we start prepackaging more of them.

signac/core/validators.py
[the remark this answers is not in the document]
vyasr: OK wow my computer just didn't refresh any of your comments. I made my last two without even seeing the ones immediately before them, but now I think I see the whole thread.

Yes, more precisely _signac_ will apply the JSON format validator to all back ends. I make this distinction to make the point that nowhere in any of the files defining SyncedCollection subclasses will we include `cls.add_validator(JSON_format_validator)` except in collection_json.py. Therefore, whatever back end the user chooses to use with signac, signac will internally need to call cls.add_validator prior to creating any instances of that class.

I'm fine with waiting to that until a future PR. The necessary conversion for backwards compatibility has been added to the JSONCollection class's sync method, so we would just remove that at some future point.

signac/core/validators.py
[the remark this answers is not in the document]
bdice: @vyasr Making them public sets a much higher bar for us to ensure their correctness, performance, and API stability. I suggest we make them internal for now, and reconsider making them public in the future. It's very hard to retract or alter public APIs, as you know.

signac/core/validators.py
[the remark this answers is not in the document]
csadorf: Agreed, the validator layer(s) should be independent of the backend and we should apply one consistent set of validators to all backends that we natively support.

signac/core/validators.py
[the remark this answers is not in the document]
vyasr: While I agree with you in principle, I think in this case we are actually better serving this interest by making these functions public. I think we should be designing the synced collections as a complete abstraction layer relative to the rest of signac, which means that signac should only consume the public API of these classes. Enforcing that requirement will force us to critically evaluate what should be public and private in the synced collections. Since `SyncedCollection.add_validator` is public, it makes sense for us to provide examples of what validators should look like; if we foresaw that changing, that would indicate that `add_validator` would also have to be internal, and that doesn't make any sense to me since (as I mentioned in [this comment](https://github.com/glotzerlab/signac/pull/378#discussion_r481559972)) signac will be applying the `JSON_format_validator` to all backends.  Exposing `add_validator` means the fundamental validator API must be clearly established before we can release this. Regarding correctness and performance, I would argue that issues with correctness constitute bugs that we'll have to fix anyway, and we don't make any performance guarantees with respect to validation, and optimizations are internal changes.

Since we've decided not to deal with removing non-str support for keys as of yet, we will be able to merge synced collections and make a 1.x release, so we'll have time to evaluate the API and if necessary make changes for 2.0. Alternatively, if you are particularly wary, we could leave them internal for now and make them public at the point of a 2.0 release. However, I would need a very strong reason to keep them internal after a 2.0 release; the need to do that would indicate in my view a significant flaw the design.

signac/core/validators.py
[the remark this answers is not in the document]
mikemhenry: I agree with @csadorf we need to have one set of validation that runs for every backend. It would be really annoying to be using a redis backend for performance but when it comes time to archive the data, find out that the keys used for redis are incompatible with json. I also agree that we should wait for a future PR to modify the current behavior.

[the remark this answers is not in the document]
mikemhenry: *ugh github ate my comment*
tl;dr I agree with @csadorf that we need to have one consistent set of validation criteria. It would be really annoying for a user to use a redis backend for performance but then when it comes time to archive the project, run issues when switching to json. I also agree we can wait for another PR to modify behavior.

bdice (approved):
@vishav1771 This looks good now! I approved it. I have one question about whether we should rename `JSON_format_validator` to `json_format_validator`, and I have asked for input on Slack.

signac/core/syncedattrdict.py
```diff
@@ -13,14 +13,13 @@
 from collections.abc import MutableMapping
```
vyasr, line 13: I noticed that the module-level docstrings for both `synced_list.py` and `synced_dict.py` say that these classes implement `to_base` and `from_base`. However, after our registry-based solution was implemented, only `to_base` needs to be implemented by subclasses because `from_base` uses the `SyncedCollection` implementation that calls the appropriate subclass constructors, so these docstrings need to be updated.

signac/core/syncedattrdict.py
```diff
@@ -193,18 +177,28 @@ def clear(self):
...
 
     def setdefault(self, key, default=None):
+        self._validate({key: default})
         self.load()
         with self._suspend_sync():
-            ret = self._data.setdefault(self._validate_key(key),
-                                        self.from_base(data=default, parent=self))
+            ret = self._data.setdefault(key, self.from_base(default, parent=self))
         self.sync()
         return ret
```
vyasr, line 203: While it is possibly premature, I think it's worth applying some minor easy optimizations to `setdefault`. If the key is already in the dict, there's no need to call `_validate`, `_data._setdefault`, or `from_base`: you can just set `ret = self._data[key]` after a load and skip the sync. I can't directly suggest because there are deleted lines in this function, but I'm thinking something like:
```
def setdefault(self, key, default=None):
    self.load()
    try:
        ret = self._data[key]
    except KeyError:
        self._validate({key: default})
        ret = self.from_base(default, parent=self))
        with self._suspend_sync():
            self._data[key] = ret
        self.sync()
    return ret
```
This method avoids a lot of extra work when the key is in the dict. I purposely chose not to use `self.__getitem__(key)` to avoid depending on that function calling `load` (in case we somehow change that implementation, just trying to avoid subtle bugs). I didn't test this implementation, please make sure I didn't make any errors, in particular with the order of when I call `from_base`.

signac/core/validators.py
```diff
@@ -0,0 +1,91 @@
+# Copyright (c) 2020 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+"""Validators for SyncedCollection API.
+
+Validators should be callable. They should act recursively for nested data and
+they should not return any values, only raise errors.
```
vyasr, line 7:
```suggestion
A validator is any callable that raises Exceptions when called with invalid data.
Validators should act recursively for nested data structures and should not
return any values, only raise errors.
```

signac/core/validators.py
```diff
@@ -0,0 +1,91 @@
...
+    TypeError
+        If data type is not supported.
+    """
+
+    if isinstance(data, (str, int, float, bool, type(None))):
+        return
+    if isinstance(data, Mapping):
+        for key, value in data.items():
+            # Support for non-str keys will be removed in version 2.0.
+            # See issue #316.
```
vyasr, line 76: I would also link the issue directly.

signac/core/validators.py
```diff
@@ -0,0 +1,91 @@
...
+    if isinstance(data, Sequence):
+        for value in data:
+            json_format_validator(value)
+        return
+    if NUMPY and isinstance(data, (numpy.ndarray, numpy.number)):
+        if numpy.iscomplex(data).any():
+            raise TypeError("NumPy object with complex value(s) is not JSON serializable")
+        else:
+            return
+    raise TypeError(f"Object of type {type(data).__name__} is not JSON serializable")
```
vyasr, line 91: All the extraneous `return` statements here are kind of like writing a reverse switch statement and shouldn't be necessary. Just convert all the `if` statements to `elif` so that nothing extra gets run, then you can remove the naked returns. Also, the final `raise TypeError` would then need to be in an `else` clause.

tests/test_synced_collection.py
```diff
@@ -154,6 +154,19 @@ def test_update(self, synced_dict, testdata):
         synced_dict.update(d)
         assert len(synced_dict) == 1
         assert synced_dict[key] == d[key]
+        synced_dict.update()
+        assert len(synced_dict) == 1
+        assert synced_dict[key] == d[key]
+        synced_dict.update(update2=testdata)
+        assert len(synced_dict) == 2
+        assert synced_dict['update2'] == testdata
+        synced_dict.update({key: 1}, update=2)
```
vyasr, line 163: Can you make a short comment here indicating that `key` is a string `update`? I did a double take reading this test and had to go back to realize why `synced_dict[key] == 2` and not 1 after this test (I didn't immediately realize `key` wasn't quoted and was actually a variable holding the string `'update'`).

tests/test_validators.py
```diff
@@ -0,0 +1,97 @@
...
+            with pytest.raises(KeyTypeError):
+                no_dot_in_key({key: testdata})
+
+
+class TestJSONFormatValidator:
+
+    def test_valid_data(self):
+        for data in ('foo', 1, 1.0, True, None, {}, []):
+            json_format_validator(data)  # should not raise any error
+        for data in ('foo', 1, 1.0, True, None, {}, []):
```
vyasr, line 50:
```suggestion
```
This line isn't necessary, you can just run both validations in one loop.

signac/core/synced_collection.py
```diff
@@ -41,47 +43,82 @@ def __init__(self, name=None, parent=None):
...
         """
         if not hasattr(cls, 'registry'):
             cls.registry = defaultdict(list)
-        for _cls in args:
-            cls.registry[_cls.backend].append(_cls)
+        for base_cls in args:
+            cls.registry[base_cls._backend].append(base_cls)
+
+    @property
+    def validators(self):
```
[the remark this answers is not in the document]
vyasr: @bdice @vishav1771 Was there any further discussion on this? My inclination is that this may not really be safe because users could always add new validators, but we could also forbid that explicitly.

If it was just decided that this request is out of scope for this PR, that is fine with me.

vyasr (changes requested):
One more round of changes. I think this should be the last one.

signac/core/synced_list.py
```diff
@@ -145,25 +149,29 @@ def __reversed__(self):
...
 
     def append(self, item):
+        self._validate(item)
         self.load()
         with self._suspend_sync():
             self._data.append(self.from_base(data=item, parent=self))
         self.sync()
 
     def extend(self, iterable):
+        self._validate(list(iterable))
```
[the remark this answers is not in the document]
bdice: Test added in 06c249ec6b282e2bfdc24d0c99f97a0fef74206a. Fixed in 1ca00fcbaf102fe5e2a598912fe1f3da3bc36ab1.

signac/core/synced_list.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, data=None, **kwargs):
         if data is None:
```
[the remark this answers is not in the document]
bdice: Fixed in ce11ad68faf58946aa4948cf2c96279b60e59bab.

signac/core/synced_list.py
```diff
@@ -145,25 +149,29 @@ def __reversed__(self):
         return reversed(self._data)
 
     def __iadd__(self, iterable):
+        self._validate(list(iterable))
```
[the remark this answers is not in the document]
bdice: Test added in 2516cbe533c85569b01f00a20a311ad935daa9cc. Fixed in 1ca00fcbaf102fe5e2a598912fe1f3da3bc36ab1.

signac/core/synced_collection.py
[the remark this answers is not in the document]
bdice: @vyasr Hmm. I guess that caching wouldn't be safe if new validators were applied to a parent class (it would be easy to empty the cache if a validator were added to _this_ class, but it would be hard to propagate that information from a parent class if the parent class's validators changed). Maybe we just leave this as it is for now. I was concerned that this might affect performance. We'll need to do thorough benchmarking/profiling before releasing the new backends. I recommend that we should defer this discussion until we look at the performance more carefully.

bdice:
@vishav1771 I made a few commits to fix some of the items @vyasr requested. I'll leave the rest to you.

signac/core/synced_collection.py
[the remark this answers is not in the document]
vyasr: Yes we'll need to profile everything once everything is feature complete. On the plus side, we've added a pretty good test suite so if we need to optimize we should be able to validate the changes relatively easy.

signac/core/syncedattrdict.py
bdice: @vishav1771 @vyasr Be aware that handling exceptions can be expensive. I recommend LBYL (`if key in self._data:` / `else:`) instead of EAFP (`try:` / `except KeyError:`) here.

Update code sample:
```python
def setdefault(self, key, default=None):
    self.load()
    if key in self._data:
        ret = self._data[key]
    else:
        self._validate({key: default})
        ret = self.from_base(default, parent=self))
        with self._suspend_sync():
            self._data[key] = ret
        self.sync()
    return ret
```

Evidence / demo of LBYL being faster:

```python
import timeit

hit = """data = {'foo': 0, 'bar': 0}"""
miss = """data = {'baz': 0, 'bar': 0}"""

lbyl = """if 'foo' in data:
  data['foo'] += 1
else:
  data['bar'] += 1
"""

eafp = """try:
  data['foo'] += 1
except KeyError:
  data['bar'] += 1
"""

hit_lbyl_time = timeit.timeit(setup=hit, stmt=lbyl, number=1000000)
hit_eafp_time = timeit.timeit(setup=hit, stmt=eafp, number=1000000)
miss_lbyl_time = timeit.timeit(setup=miss, stmt=lbyl, number=1000000)
miss_eafp_time = timeit.timeit(setup=miss, stmt=eafp, number=1000000)

print('Hit, LBYL:', hit_lbyl_time)
print('Hit, EAFP:', hit_eafp_time)
print('Miss, LBYL:', miss_lbyl_time)
print('Miss, EAFP:', miss_eafp_time)
```

Results:
```
Hit, LBYL: 0.0599388
Hit, EAFP: 0.0484887
Miss, LBYL: 0.0627916
Miss, EAFP: 0.2142205
```

vyasr: That's true, but the typical use case for `setdefault` is when keys are likely to be present. In this case EAFP expresses the expected usage pattern more appropriately. Having said that, I don't think there's any "representative" use case to know how the trade offs play out in practice. Based on your timings you would need >10:1 hit:miss ratio for EAFP to be faster. That's actually probably a pretty realistic in any case where I would actually use `setdefault`; in practice unless I expect mostly hits I write LBYL code myself rather than use `setdefault`. However, I don't really know what others usually do (tbh in my experience very few people use this method in general) so I don't have a strong opinion if your usage pattern indicates that switching would constitute a meaningful performance improvement.

bdice: I would find it hard to justify an assumed ratio of >10:1 or more. I lean towards LBYL, because we cannot know the true ratio to expect. For example, I have used `setdefault` in cases where I end up using the “default” more frequently than the stored value, which is the opposite of the “mostly read from existing key” (mostly hits) scenario posed above.

bdice: I looked through the CPython standard library and it’s a mixture of LBYL and EAFP approaches, depending on the class. There is no clear preference as far as I can tell.

vyasr: That's fine with me. Frankly I almost always end up using a `defaultdict` anywhere that I might otherwise use `setdefault`, the only real exception being if I'm modifying a dict produced by some third party code, so I'm probably not a representative user of `setdefault` insofar as such a user still exists in Python 3.8.

signac/core/validators.py
```diff
@@ -0,0 +1,89 @@
...
+
+    Parameters
+    ----------
+    data
+        Data to validate.
+
+    Raises
+    ------
+    KeyTypeError
+        If keys have unsupported data type.
```
vyasr, line 35:
```suggestion
        If key data type is not supported.
```

signac/core/validators.py
```diff
@@ -0,0 +1,89 @@
...
+    InvalidKeyError
+        If key is not supported.
```
vyasr, line 37:
```suggestion
        If the key contains invalid characters or is otherwise malformed.
```

signac/core/validators.py
```diff
@@ -0,0 +1,89 @@
...
+
+    Parameters
+    ----------
+    data
+        Data to validate.
+
+    Raises
+    ------
+    KeyTypeError
+        If keys of mapping have unsupported data type.
```
vyasr, line 67:
```suggestion
        If key data type is not supported.
```

signac/core/validators.py
```diff
@@ -0,0 +1,89 @@
...
+    TypeError
+        If data type is not supported.
```
vyasr, line 69:
```suggestion
        If the data type of ``data`` is not supported.
```

vyasr (approved):
Great work! I have some minor formatting changes, but they're all suggestions so I'm just going to apply them (and fix any style issues I cause...) and then go ahead and merge.

PR #379, opened 2020-08-26
Remove dependency on pytest-subtests.

bdice:
## Description
This PR changes a small number of tests that relied on the `pytest-subtests` plugin.

## Motivation and Context
The `pytest-subtests` plugin was causing some trouble because upgrades to `pytest` broke compatibility with `pytest-subtests` on two separate occasions:
1. 0.3.1 to 0.3.2 breakage is documented here: #366, https://github.com/glotzerlab/signac/pull/366#issuecomment-676579490.
2. 0.3.0 to 0.3.1 breakage is not documented but I remember running into it on my own, and found a comment to corroborate this on Slack.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -1480,85 +1480,68 @@ class TestProjectExportImport(TestProjectBase):
             assert len(tmp_project) == len(self.project)
 
 
-class TestProjectRepresentation(TestProjectBase):
-
+def add_jobs_homogeneous(project, num_jobs):
+    # Add jobs with many different state points
     valid_sp_values = [None, 0, 1, 0.0, 1.0, True, False, [0, 1, 2], [0, 1.0, False]]
+    for i in range(num_jobs):
+        project.open_job(
+            {'{}_{}'.format(i, j): v
+                for j, v in enumerate(valid_sp_values)}).init()
+
+
+def add_jobs_heterogeneous(project, num_jobs):
+    # Add jobs with many different state points
+    valid_sp_values = [None, 0, 1, 0.0, 1.0, True, False, [0, 1, 2], [0, 1.0, False]]
+    for i in range(num_jobs):
+        for v in valid_sp_values:
+            project.open_job(dict(a=v)).init()
+
+
+project_repr_generators = [
+    (add_jobs_homogeneous, 0),
+    (add_jobs_homogeneous, 10),
+    (add_jobs_homogeneous, 200),
+    (add_jobs_heterogeneous, 0),
+    (add_jobs_heterogeneous, 10),
+    (add_jobs_heterogeneous, 200)]
+
+
+class TestProjectRepresentation(TestProjectBase):
 
     num_few_jobs = 10
     num_many_jobs = 200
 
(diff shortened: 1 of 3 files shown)


tests/test_project.py
```diff
@@ -1480,85 +1480,68 @@ def test_import_own_project(self):
...
-    def test_repr_many_jobs_heterogeneous(self, subtests):
-        # Many jobs with many different state points
-        for i in range(self.num_many_jobs):
-            for v in self.valid_sp_values:
-                self.project.open_job(dict(a=v)).init()
-        self.call_repr_methods(subtests)
+    @pytest.mark.parametrize("project_generator,num_jobs", project_repr_generators)
+    def test_project_repr_methods(self, project_generator, num_jobs):
+        project_generator(self.project, num_jobs)
+        str(self.project)
```
csadorf, line 1517: Should we do a super simple sanity check here, something like `assert len(self.project) > 0` or `assert 'project' in str(self.project)` ?

tests/test_project.py
```diff
@@ -1480,85 +1480,68 @@ def test_import_own_project(self):
...
+                if use_pandas and not PANDAS:
+                    raise pytest.skip('requires use_pandas')
+                self.project.find_jobs(filter_)._repr_html_()
+
+    @pytest.mark.parametrize("project_generator,num_jobs", project_repr_generators)
+    def test_Schema_repr_methods(self, project_generator, num_jobs):
+        project_generator(self.project, num_jobs)
+        schema = self.project.detect_schema()
+        str(schema)
+        repr(schema)
```
csadorf, line 1543: Similar to my comment above.

csadorf (approved):

tests/test_project.py
bdice: Good suggestion! I added `assert len(str(self.project)) > 0` and some related tests. The other case is already handled twice by `TestProject.test_get_id` and `TestProject.test_property_id`.

bdice:
@klywang Can you please review?

tests/test_project.py
```diff
@@ -1480,85 +1480,71 @@ def test_import_own_project(self):
             assert len(tmp_project) == len(self.project)
 
 
-class TestProjectRepresentation(TestProjectBase):
+def add_jobs_homogeneous(project, num_jobs):
+    # Add jobs with many different state points
+    valid_sp_values = [None, 0, 1, 0.0, 1.0, True, False, [0, 1, 2], [0, 1.0, False]]
```
klywang, line 1485: If the goal is to always test the same values for `valid_sp_values` for `add_jobs_homogeneous` and `add_jobs_heterogeneous`, would it make sense to define `valid_sp_values` outside of both functions and take them in as inputs?

klywang (approved):
Everything looks good!

tests/test_project.py
bdice: Good idea! I'll define a module-level constant.

PR #399, opened 2020-10-28
Fix/find_job_ids() deprecation

atravitz:
## Description
I've gone through the code base and made appropriate replacements for all instances of `find_job_ids()`.

## Motivation and Context
`find_job_ids()` is deprecated. The current deprecation message instructs users to use `find_jobs().id` which is incorrect. Also, we still use `find_job_ids()` internally in tests. 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:


tests/test_sync.py
```diff
@@ -648,11 +648,11 @@ def test_selection(self):
...
         with pytest.raises(FileSyncConflict):
-            self.project_a.sync(self.project_b, selection=self.project_b.find_job_ids({'a': 0}))
+            self.project_a.sync(self.project_b, selection=self.project_b.find_jobs({'a': 0}))
         f = {'a': {'$ne': 0}}
         self.project_a.sync(self.project_b, selection=self.project_a.find_jobs(f))
         self.project_a.sync(self.project_b, selection=self.project_b.find_jobs(f))
-        self.project_a.sync(self.project_b, selection=self.project_a.find_job_ids(f))
-        self.project_a.sync(self.project_b, selection=self.project_b.find_job_ids(f))
+        self.project_a.sync(self.project_b, selection=self.project_a.find_jobs(f))
+        self.project_a.sync(self.project_b, selection=self.project_b.find_jobs(f))
```
vyasr, line 658: The goal of this is to test that passing either `Job` instances or job id strings works, so you'll want to do the conversion here. Otherwise the two modified lines are identical to the two lines above them.

signac/contrib/project.py
```diff
@@ -799,7 +799,7 @@ def detect_schema(self, exclude_const=False, subset=None, index=None):
         return ProjectSchema.detect(statepoint_index)
 
     @deprecated(deprecated_in="1.3", removed_in="2.0", current_version=__version__,
-                details="Use find_jobs().ids instead.")
+                details="Use [job.id for job in project.find_jobs()] to return job ids instead.")
```
vyasr, line 802: I'm not sure this is the right snippet to include here, or at least not the only snippet. Users may not always want to get all job ids directly like this. I think it would make sense to recommend just switching to `find_jobs`, then add another sentence indicating that users can access the ids of these jobs to replicate the original behavior, at which point you could include this list comprehension.

vyasr (changes requested):
I just realized that this hasn't been modified since our last discussion, so it still has a lot of `__eq__` usage etc. Once you're happy with this PR mark it as not a draft and I'll come back :)

csadorf:
Have we considered to add a function `JobsCursor.ids()` that *does* return an iterator over the ids? That would enable:
```python
ids = project.find_jobs().ids()
```

vyasr:
@csadorf I gave @atravitz quite a bit more advice regarding this PR via DM, including discussion of the question that you asked. To paraphrase my response, I think it might be useful but I have not ever personally needed it and I'm not sure how extensive the existing use cases are; in at least some of the internal usage we might be able to get away with directly working with the `Job` rather than the `id`. I proposed opening an issue to discuss that as a potential feature, but to postpone implementing it until we got more feedback, especially in light of all the ongoing work on aggregation (which will likely change how ids and maybe the `JobsCursor` work).

atravitz:
This is ready for review! However I'm stumped as to why my changes to `test_import_own_project()` in `test_project.py` are not passing.

signac/contrib/project.py
```diff
@@ -2271,7 +2271,7 @@ def next(self):
 
         """
         warnings.warn("Calling next() directly on a JobsCursor is deprecated!"
-                      "Use iter(..).next() instead.", DeprecationWarning)
+                      "Use next(iter(JobsCursor)) instead.", DeprecationWarning)
```
csadorf, line 2274: That's not quite right actually, since you would not want to call it on the `JobsCursor` class, but on the instance. So I suggest to keep using the ellipsis:
```suggestion
                      "Use next(iter(..)) instead.", DeprecationWarning)
```

csadorf:
Can you remind me: Is it established that `[job.id for job in project.find_jobs()]` is as efficient as `project.find_job_ids()` ? I am just wondering whether the performance implications are fully considered here and whether we might need to special-case some code paths here to retain performance. Because instead of completely removing `Project.find_job_ids()` we could also just make it private.

vyasr:
That's a great question. I don't think that we have benchmarked this before. However, if `find_job_ids` is truly faster I think I would prefer to take that as an opportunity to optimize `JobsCursor` rather than just leave this path in place. If the `JobSearchIndex` is much faster by virtue of the way it uses `Collection`s, then unifying the code paths further would allow us to reduce code and get better performance.

bdice:
tl;dr: `list(project._find_job_ids())` is really fast compared to `[job.id for job in project.find_jobs()]`. This is a performance regression.

Benchmarks:
```
All values in ms.
tool                           signac
N                               100    1000    10000
Determine N O(N)                 0.00   0.00    0.00
Select by ID O(1)                0.02   0.02    0.02
Iterate (multiple passes) O(N)   0.03   0.03    0.03
Iterate (single pass) O(N)       0.04   0.04    0.05
Search w/ lean filter O(N)       0.00   0.00    0.01
Search w/ rich filter O(N)       0.02   0.02    0.02
iterate_find_jobs                2.63  27.00  273.14
iterate_find_jobs_get_ids        2.49  25.49  259.99
iterate_find_job_ids             0.08   0.76    7.57
```

Here, I use the following benchmark definitions pasted into `benchmark.py` (which may or may not be most optimal):

```python
    run('iterate_find_jobs', Timer(
        stmt="list(project.find_jobs())",
        setup=setup))

    run('iterate_find_jobs_get_ids', Timer(
        stmt="[job.id for job in project.find_jobs()]",
        setup=setup))

    run('iterate_find_job_ids', Timer(
        stmt="list(project._find_job_ids())",
        setup=setup))
```

I see the same results for `iterate_find_job_ids` regardless of whether I use the deprecated public method `project.find_job_ids()` or the internal method `project._find_job_ids()`.

csadorf:
That's a highly significant performance regression. 😮 Those should certainly be caught by our CI performance test, right?

vyasr:
Nice thanks for doing that @bdice. I'm not too surprised, although the magnitude is larger than I expected. This seems to me like an opportunity to optimize the `JobsCursor`. One of the earlier comments on this thread discusses the possibility of enabling something like `JobsCursor(...).ids()`, and I suspect that if the cursor iterated over ids and opened jobs only on demand (and the `ids` method just returned without opening the job) you would recover most of the performance, so that might be worth trying.

We removed benchmarks from CI at some point due to inconsistent results. I'm not sure if this is an indication that we should reenable it; if we want to, I think we'd want to find somewhere other than CircleCI because of the large inconsistencies introduced by the fact that resources on there are not isolated to specific jobs.

atravitz:
Should we not be suggesting `[job.id for job in project.find_jobs()]` if it's this much slower? I think we should probably aim for having a comparably fast way of getting the same information (in a different PR), but I think in the meantime it's good to explicitly provide an alternative to users.

bdice:
I think this PR should be merged because the current suggestion is invalid. We should make an issue describing the performance regression. I think the best solution to this is to finish rewriting the backend. I think this is only one of the many places that performance can be improved and don't want to get hung up on it.

----

Tests are currently failing because:
```python
Job(project=Project.get_project('/tmp/signac_f2honvf1/pr'), statepoint={'a': 5})
```
is considered not equal to
```python
Job(project=Project.get_project('/tmp/signac_f2honvf1/wd/tmpakbkae57'), statepoint={'a': 5})
```

Let's switch back to verifying equality of the two sets of job ids. Apparently the jobs aren't considered equal because their project directory is defined differently (perhaps this is a symptom of an underlying issue and we should check realpaths of the found project directory instead?).

csadorf:
@bdice I initially thumbed down your comment, because I thought it was the wrong call. Since then I have thought about it a bit more and changed my mind, because the affected function was already marked as deprecated.

vyasr:
I think we can move forward with this PR, we simply can't remove `Project._find_job_ids` or change internal usage of that method until we have a comparably fast alternative.

bdice (approved):
I made some changes on this PR and I think it's ready to merge. I'd like to get a second reviewer from someone requested: @pepak13, @vyasr, @csadorf?

atravitz:
> I made some changes on this PR and I think it's ready to merge. I'd like to get a second reviewer from someone requested: @pepak13, @vyasr, @csadorf?

I'm concerned about replacing `project.find_job_ids()` with `[job.id for job in project.find_jobs()]` if it's going to hurt performance significantly. Is there a reason we can't keep `project.find_job_ids()` and not throw the deprecation warning until we come up with a more efficient solution?

bdice:
The inefficiency matters a lot if it’s in the internals of signac, but not as much if it’s in the user API. Since we have a private method that does the same thing, we don’t need to worry about the internal speed. I don’t think this is a commonly used function in the public user API and I expect the long term goal of adding lazy state point loading to correct any performance penalty a user faces with the suggested replacement.

atravitz:
okay then, I'm fine with merging this as soon as we get another review then!

tests/test_sync.py
```diff
@@ -657,14 +657,18 @@ def test_selection(self):
             )
         with pytest.raises(FileSyncConflict):
             self.project_a.sync(
-                self.project_b, selection=self.project_a.find_job_ids({"a": 0})
+                self.project_b, selection=[job.id for job in self.project_a.find_jobs()]
```
csadorf, line 660: Why not:
```suggestion
                self.project_b, selection=[job.id for job in self.project_a.find_jobs({"a": 0})]
```

tests/test_sync.py
```diff
@@ -657,14 +657,18 @@ def test_selection(self):
...
             )
         with pytest.raises(FileSyncConflict):
             self.project_a.sync(
-                self.project_b, selection=self.project_b.find_job_ids({"a": 0})
+                self.project_b, selection=[job.id for job in self.project_b.find_jobs()]
```
csadorf, line 664: same as above

csadorf (changes requested):
Great work! Just found two issues.

tests/test_sync.py
bdice:
```suggestion
                self.project_b, selection=[job.id for job in self.project_b.find_jobs({"a": 0})]
```

csadorf (approved):

Issue #281, opened 2020-01-30
H5Store pytest is failing on MacOS but passes on Linux


klywang:
### Description

Currently, one of the multiprocessing unit tests, `test_single_writer_multiple_reader_same_instance` from `test_h5store.py`, will fail on MacOS. The test passes on Linux and is skipped on Windows due to failure for an unknown reason. 

My system has `HDF5` version 10.4.1 and `h5py` version 2.10.0.

### To reproduce
On Mac:
```bash
$ cd /path/to/signac/
$ pytest tests/test_h5store.py -k "TestH5StoreMultiProcessing"
```
This will return an error. This test will pass on Linux (tested locally and tested in CI) and is skipped on Windows (not tested locally and tested on CI)

### Error output

```bash
OSError: Unable to open file (unable to lock file, errno = 35, error message = 'Resource temporarily unavailable')
```

### System configuration

Please complete the following information:

 - Operating System [e.g. macOS]: macOS 
 - Version of Python [e.g. 3.7]: 3.8.1
 - Version of signac [e.g. 1.0]: 1.3.0

Or copy & paste the output of: `python -c 'import platform; print(platform.platform()); import sys; print(sys.version); import signac; print(signac.__version__)'`

csadorf:
I've marked this as a bug, since this might be a potential issue on Mac OS systems. @klywang Would you be willing to add the Mac test to the CI pipeline?

klywang:
Would this branch be categorised as `fix` or `misc`?

bdice:
> Would this branch be categorised as `fix` or `misc`?

Branch names are very flexible - maybe something like `ci/mac`. Your call!

vyasr:
@klywang did you have any luck figuring out why this fails?

b-butler:
@klywang is this still an issue? If so, do you have time to add the mac test to our CI pipeline, and potentially find the bug?

vyasr:
I'm observing this issue again and will try to resolve it.

PR #411, opened 2020-11-02
Add support for h5py >=3.0.0.

bdice:
## Description
This PR adds support for h5py major version 3 and should be merged before #409 (to show that the code functions both before and after the upgrade).

The h5py package changed its default handling of strings in version 3. See this page for info: https://docs.h5py.org/en/latest/strings.html

This PR adds a shim to detect string data, so that it can be returned as `str` and not `bytes`. I tested this locally with h5py 2.10.0 and 3.0.0, and it should work for older versions of h5py as well (that's why I used the older function `check_dtype` instead of `check_string_dtype`, which was added in 2.10).

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -7,6 +7,14 @@ The **signac** package follows `semantic versioning <https://semver.org/>`_.
 Version 1
 =========
 
+next
+----
+
+Added
++++++
+
+ - Support for h5py version 3 (#411).
+
 [1.5.0] -- 2020-09-20
 ---------------------
 
--- a/signac/core/h5store.py
+++ b/signac/core/h5store.py
@@ -130,6 +130,7 @@ def _h5set(store, grp, key, value, path=None):
 
 def _h5get(store, grp, key, path=None):
     """Retrieve the underlying data for a key from its h5py container."""
+    import h5py
     path = path + '/' + key if path else key
     result = grp[key]
 
@@ -153,6 +154,11 @@ def _h5get(store, grp, key, path=None):
             return None
         elif shape:
             return result
+        elif h5py.version.version_tuple.major >= 3 and \
+                h5py.check_dtype(vlen=result.dtype) is str:
+            # h5py >=3.0.0 returns strings as bytes. This returns str for
+            # consistency with past behavior in signac.
+            return result.asstr()[()]
         else:
             return result[()]
     except AttributeError:


jennyfothergill:
Hi!
Pytest is throwing an error for me. Here is a file with the error output:
[pytest-error.txt](https://github.com/glotzerlab/signac/files/5477045/pytest-error.txt)
And just to make sure that I'm not doing something wrong, below are my install steps:
(in signac root dir)
```
gh pr checkout 411
conda env create -f environment.yml  # [environmentyml.txt](https://github.com/glotzerlab/signac/files/5477068/environmentyml.txt)
conda activate signac-dev
pip install -e .
```
I also tried changing the line in environment.yml to `h5py>=3` and then conda gave me package not found, so I tried installing like this:
```
conda create -n signac-dev pip
conda activate signac-dev
pip install 'h5py>=3' pytest
pip install -e .
```
and I get the same error from pytest.
I'm on osx 10.14.6 (18G103)

bdice:
@jennyfothergill 
1. Test failure: thanks for the details, that's helpful. I think this test failure is the same as #281. This seems to be testing a behavior that isn't very safe across different operating systems. We should probably consider removing this test since it only appears to work on Linux.

2. h5py 3.0 isn't available on conda-forge yet, so adding it to the `environment.yml` won't work yet. You're an early adopter! 😄 Seems like you got it installed from PyPI and saw the same error as above.

Sounds like the takeaway is that this works except for the known issue in #281. I appreciate the extensive testing!

jennyfothergill:
@bdice Ah, OK I wasn't sure if that error was already known, but I should've checked the issues first! Haha, I don't know about extensive testing. :) I'm not sure how to check that strings are working correctly. (I assuming that I should test if strings work as statepoint keys/values?) I tried:
```
signac init myproject
signac job '{"N": 1.0, "type": "heck"}' --create
```
and it seemed to work fine with both versions of h5py. Is there a better way to test?

bdice:
@jennyfothergill signac uses h5py for the H5Store class. This can be accessed with `job.data`, `project.data`, or as a standalone class `signac.H5Store`. This is useful for storing large arrays of values (like a big NumPy array) that are too large for a job document (stored as JSON, which is not an efficient way to store large amounts of binary data). See here for API examples: https://docs.signac.io/projects/core/en/latest/api.html#the-h5store

jennyfothergill (approved):
LGTM. I'm still not sure if I'm checking this correctly, but with h5py==2.10 and h5py>=3 both print `<class 'str'>`:
```python
from signac import H5Store
with H5Store('file.h5') as h5s:
    h5s['foo'] = 'bar'
    print(type(h5s.foo))
```

mikemhenry (commented):
LGTM! Thanks Jenny for that thorough review!

mikemhenry (approved):

PR #415, opened 2020-11-10
Use black and isort, managed with pre-commit.

bdice:
## Description
1. This PR introduces `black` and `isort` for automatically fixing code style.
2. The checks for `flake8`, `pydocstyle`, and `mypy` are the same as before, except they are no longer "local" hooks (they use the pre-commit virtual environment instead).

**I am seeking reviewer feedback on the new configuration before committing reformatted code.**

For an example of what the style looks like, see #416. I applied all the changes in that PR.

## Motivation and Context
1. The development team decided to adopt `black` as a code formatter at the previous developer meeting. I hope this will make the process of development faster as a whole. While there is a nonzero cost to having a pre-commit workflow with many hooks, it is worth it if this kind of change can reduce the number of iterations and "nitpick" style comments needed during PR review.

2. Consolidating pre-commit hooks (and not using local hooks) will make it easier to install and manage linters/fixers.

## Types of Changes
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,33 +1,38 @@
 repos:
--   repo: https://github.com/pre-commit/pre-commit-hooks
+  - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: ''
     hooks:
-    -   id: end-of-file-fixer
+      - id: end-of-file-fixer
         exclude: 'setup.cfg'
-    -   id: trailing-whitespace
+      - id: trailing-whitespace
         exclude: 'setup.cfg'
-    -   id: debug-statements
--   repo: https://github.com/asottile/pyupgrade
+      - id: debug-statements
+  - repo: https://github.com/asottile/pyupgrade
     rev: v2.7.3
     hooks:
-    -   id: pyupgrade
+      - id: pyupgrade
         exclude: '(?:configobj/.*)'
         args:
-        - --py36-plus
--   repo: local
+          - --py36-plus
+  - repo: https://github.com/psf/black
+    rev: 20.8b1
     hooks:
-    -   id: flake8
+      - id: black
+        language_version: python3
+  - repo: local
+    hooks:
+      - id: flake8
         name: flake8
         entry: flake8
         language: python
         types: [python]
-    -   id: pydocstyle
+      - id: pydocstyle
         name: pydocstyle
         entry: pydocstyle
         language: python
         types: [python]
         files: ^((?!doc|tests|configobj|db).)*$
-    -   id: mypy
+      - id: mypy
         name: mypy
         entry: mypy
         language: python
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: ''
+    rev: 'v3.3.0'
     hooks:
       - id: end-of-file-fixer
         exclude: 'setup.cfg'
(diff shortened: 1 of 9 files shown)


vyasr:
Is there a good reason to skip string normalization? I don't see a compelling reason not to do so, and part of the benefit of black is to completely remove these questions as points of discussion. I'd prefer not to configure anything formatting related other than line length (and honestly even that I could be convinced to change, I feel like we've had debates before over whether 100 is really "better").

There are also options for `isort`, those can be specified either in pyproject.toml or setup.cfg (or a .isort.cfg file). At minimum, you will want to set the line length in isort to match black. Other changes will be smaller and probably less controversial than black configuration, so we could change from the defaults if we like. My preference would probably be something like
```
line_length = 100
use_parentheses = True
include_trailing_comma = True
multi_line_output = 4
force_grid_wrap = 0
```
The first three are self-explanatory. The fourth basically sets it up so that multi-line imports form a grid as opposed to looking ragged or only one per line. The fifth option says to ignore the number of items on a given import line and always wrap as a grid. I think the first three options are no-brainers (line length is required). I'd suggest setting _something_ for `multi_line_output`, but if someone else feels strongly for some reason then we could choose a different layout. You can see the options [here](https://pycqa.github.io/isort/docs/configuration/options/#multi-line-output).

bdice:
> Is there a good reason to skip string normalization?

> I'd prefer not to configure anything formatting related other than line length (and honestly even that I could be convinced to change, I feel like we've had debates before over whether 100 is really "better").

I agree both of these are good ideas. Initially, I wanted to review some changes manually before letting `black` change this, because changing string normalization and line length touches a LOT of lines of code. I feel comfortable with the refactors that are currently in #416 and I can enable string normalization and default line lengths now. I won't review the changes by hand as I did for the other formatting.

> There are also options for isort, those can be specified either in pyproject.toml or setup.cfg (or a .isort.cfg file). At minimum, you will want to set the line length in isort to match black.

I forgot to do this, good catch. The easiest way to do this appears to be `isort --profile black`. https://pycqa.github.io/isort/docs/configuration/compatibility_black/

vyasr:
Yes that isort setting should work fine. The require black compatibility value for the multi_line_output is 3 I believe, but I think what you've done should be fine.

On my first pass through I assumed that you had done this intentionally just to wait, but perhaps it's just an oversight: you're not currently actually checking black anywhere are you? Why aren't you running `black --check` or installing `flake8-black` prior to running `flake8`?

bdice:
@vyasr It runs, that’s why the style checks are failing. The run config and most settings are inherited from the git repo where the hook is defined! 😁 See the CI output of #416.

vyasr:
> @vyasr It runs, that’s why the style checks are failing. The run config and most settings are inherited from the git repo where the hook is defined! 😁 See the CI output of #416.

All the failures are `isort` failures, not `black` failures...

b-butler (commented):
I can't seem to find black in the `.pre-comit-config.yaml` file.

vyasr:
> I can't seem to find black in the `.pre-comit-config.yaml` file.

Right, I think that's just missing.

bdice:
@vyasr @b-butler Haha, I think I made a silly mistake during a rebase and deleted the `black` configuration. Thanks for catching that, my error defeated the purpose of this PR. 😆 The changes (with `black`) are correctly committed on #416.

vyasr:
That's better lol.

vyasr (approved):

b-butler (approved):
LGTM

bdice:
I will merge #415 and #416 together. It will take some time to resolve merge conflicts on other open PRs.

.pre-commit-config.yaml
```diff
@@ -1,34 +1,39 @@
...
+  - repo: https://github.com/asottile/pyupgrade
+    rev: 'v2.7.3'
     hooks:
-    -   id: pyupgrade
+      - id: pyupgrade
         exclude: '(?:configobj/.*)'
         args:
-        - --py36-plus
--   repo: local
+          - --py36-plus
```
mikemhenry, line 16: Does `pyupgrade` support using  `pyproject.toml` for config options? My preference is to use  `pyproject.toml` as much as we can so when/if people run these tools manually, it uses the same config.

mikemhenry: @bdice sorry I was a bit late with this comment! I'll make an issue about moving args/config options out of the `.pre-commit-config.yaml` where possible

bdice: @mikemhenry It is not possible to use `pyproject.toml`. More detail in my comment on #418. Good question!

PR #422, opened 2020-11-24
Attempt to exit early from _init if a job manifest exists and is valid.

bdice:
## Description
Many methods/properties of `Job` invoke a call to `job.init()`, including the first access to `job.document` and all accesses to `job.stores` and `job.data`.

This PR optimizes `job.init()` for jobs that are already initialized, by exiting early.

## Motivation and Context
This is a performance enhancement for many common data access operations in signac.

Performance results for the script below:

| Test           | `master` | This branch (b4917d8) | Speedup         |
|----------------|----------|-----------------------|-----------------|
| First init     | 3.426s   | 3.546s                | 0.966x (slower) |
| Already init'd | 2.623s   | 1.774s                | 1.479x (faster) |

```python
import signac

for i in range(20000):
    project.open_job({'a': i, 'b': i*2, 'c': i*3}).init()
```

## Types of Changes
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -19,6 +19,11 @@ Added
  - Added macOS to CircleCI testing pipeline (#281, #414).
  - Official support for Python 3.9 (#417).
 
+Changed
++++++++
+
+ - ``Job.init()`` exits early for jobs that are initialized (#422).
+
 [1.5.0] -- 2020-09-20
 ---------------------
 
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -492,8 +492,6 @@ class Job:
         try:
             with open(fn_manifest, "rb") as file:
                 assert calc_id(json.loads(file.read().decode())) == self._id
-        except OSError:
-            raise
         except (AssertionError, ValueError):
             raise JobsCorruptedError([self._id])
 
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -449,6 +449,16 @@ class Job:
         """
         fn_manifest = os.path.join(self._wd, self.FN_MANIFEST)
 
+        # Attempt early exit if the manifest exists and is valid
+        try:
+            self._check_manifest()
+        except Exception:
+            # Any exception means this method cannot exit early
+            pass
+        else:
+            # The manifest is exists and is valid, exit early
+            return
+
         # Create the workspace directory if it did not exist yet.
         try:
             _mkdir_p(self._wd)
@@ -481,14 +491,13 @@ class Job:
             self._check_manifest()
 
     def _check_manifest(self):
-        """Check whether the manifest file is correct (if it exists)."""
+        """Check whether the manifest file exists and is correct."""
         fn_manifest = os.path.join(self._wd, self.FN_MANIFEST)
         try:
             with open(fn_manifest, "rb") as file:
                 assert calc_id(json.loads(file.read().decode())) == self._id
-        except OSError as error:
-            if error.errno != errno.ENOENT:
-                raise error
+        except OSError:
+            raise
         except (AssertionError, ValueError):
             raise JobsCorruptedError([self._id])
 
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -453,42 +453,38 @@ class Job:
         try:
             self._check_manifest()
         except Exception:
-            # Any exception means this method cannot exit early
-            pass
-        else:
-            # The manifest is exists and is valid, exit early
-            return
+            # Any exception means this method cannot exit early.
 
-        # Create the workspace directory if it did not exist yet.
-        try:
-            _mkdir_p(self._wd)
-        except OSError:
-            logger.error(
-                "Error occured while trying to create "
-                "workspace directory for job '{}'.".format(self)
-            )
-            raise
-
-        try:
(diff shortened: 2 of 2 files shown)


signac/contrib/job.py
```diff
@@ -481,14 +491,13 @@ def _init(self, force=False):
...
+        """Check whether the manifest file exists and is correct."""
         fn_manifest = os.path.join(self._wd, self.FN_MANIFEST)
         try:
             with open(fn_manifest, "rb") as file:
                 assert calc_id(json.loads(file.read().decode())) == self._id
-        except OSError as error:
-            if error.errno != errno.ENOENT:
-                raise error
+        except OSError:
+            raise
```
bdice, line 500: Explanation of my reasoning for this change:
- The `_check_manifest` method is _only_ called in `_init`. Thus it is safe to change its behavior as long as it is consistent with the conditions allowed within the calling context in `_init`.
- At the end of `_init` where the `try` statements succeed and `_check_manifest` is called, the manifest _must_ exist (this makes no sense otherwise, even in cases of filesystem error that are handled in `_init`, because `_check_manifest` is guarded by a `try: ... else:`).
- Thus, it must be safe to remove the exception handling code for cases where the manifest does not exist.

This change of behavior is necessary because I want `_check_manifest` to raise a `FileNotFoundError` (a subclass of `OSError`) in the "early exit" attempt. This "EAFP" approach of raising an error in `_check_manifest`, instead of the "LBYL" approach of checking that the file exists before validating, minimizes the total number of system calls needed for the early exit for optimal performance. ([EAFP vs LBYL](https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/))

changelog.txt
```diff
@@ -19,6 +19,11 @@ Added
  - Added macOS to CircleCI testing pipeline (#281, #414).
  - Official support for Python 3.9 (#417).
 
+Changed
++++++++
+
+ - ``Job.init()`` exits early for jobs that are initialized (#422).
```
csadorf, line 25:
```suggestion
 - Optimized performance of job initialization (#422).
```
I don't think that implementation details belong into the changelog.

signac/contrib/job.py
csadorf: I think you reasoning is sound, but then I would suggest to completely remove lines 499..500 as well, because you are just catching an exception and raising it again so you can just remove it.

signac/contrib/job.py
```diff
@@ -449,6 +449,16 @@ def _init(self, force=False):
         """
         fn_manifest = os.path.join(self._wd, self.FN_MANIFEST)
 
+        # Attempt early exit if the manifest exists and is valid
+        try:
+            self._check_manifest()
+        except Exception:
+            # Any exception means this method cannot exit early
+            pass
```
csadorf, line 457: Instead of a `pass`, might I suggest to remove the complete rest of the function block here? Then you can also remove lines 458..461. It will create a larger diff, but the logic is clearer.

csadorf (changes requested):

signac/contrib/job.py
bdice: Good catch. I removed those lines.

signac/contrib/job.py
bdice: I think I did what you wanted. Please check in 459632ec3e859d63ce8eabf4d324cceb0767ceb8.

signac/contrib/job.py
```diff
@@ -449,46 +449,49 @@ def _init(self, force=False):
...
-                "workspace directory for job '{}'.".format(self)
-            )
-            raise
-
-        try:
-            # Ensure to create the binary to write before file creation
-            blob = json.dumps(self._statepoint, indent=2)
+            self._check_manifest()
+        except Exception:
+            # Any exception means this method cannot exit early.
```
csadorf, line 456: I am actually wondering whether we should be specific here. Expected exceptions are `OSError`, what else?

bdice: `_check_manifest` can raise `JobsCorruptedError`, which is a subclass of `RuntimeError`. There may be other exception classes that are raised in that function (it uses `json`, `hashlib`, file reading/decoding). I am hesitant to specify the exception classes because I don't know how to be sure that we're providing sufficient exception coverage and the penalty is high. _Not_ catching an exception would lead to broken behavior, because the job wouldn't actually be initialized at the end of the initialization method.

csadorf: But any of those very unexpected errors are something we should actually raise. We could very easily test which error classes to catch by simply running this against "no file", "corrupted file", "empty file", "binary file", and "permissions wrong", and that should be about it IMO. Catching "all" errors here means to effectively change the behavior of this function, does it not?

bdice: I would be fine with testing against those cases explicitly. However, I don't think the behavior of the function is changed by catching `Exception` here. From the previous behavior, I would say that the "finishing condition" of the initialization is that `_check_manifest` can complete without raising. I'm just testing that condition at the beginning as a way to exit early. Any unusual errors would still be raised in the original behavior, when `_check_manifest` was called at the end.

csadorf (approved):

lyrivera (approved):
LGTM!

PR #421, opened 2020-11-24
Simplify internal function for mkdir -p.

bdice:
## Description
Since Python 2 support has been removed, we can replace the internal function `_mkdir_p` with `os.makedirs(path, exist_ok=True)`.

## Motivation and Context
Simplifies internal code. Standard library features should be used where possible.

Cases where _files_ exist with the name for the destination directory raise an error. This was previously caught and tested separately but no longer needs special error handling code.

## Types of Changes
- [x] New feature

## Checklist:
- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
- [x] The changes introduced by this pull request are covered by existing or newly introduced tests.

PR #428, opened 2020-12-13
Optimize buffer load checks.

bdice:
## Description
This PR changes the internal behavior of the buffer. When a new item is added to the buffer, the code checks the current buffer load (size of all items in the buffer) and ensures that the current load plus the size of the new item is less than the maximum buffer size. The function that calculates this buffer size appears to consume a lot of time in my profiles of signac-flow projects with 1000 jobs. I made changes that will increment the buffer load after each store, so that buffer size access can be done in constant time.

The buffer load is reset when the buffer is flushed. I did some manual testing to ensure that the size calculations agree during a normal filling and flushing of the buffer.

## Motivation and Context
The profiling I did on signac-flow involved some status checks with label functions that required job document access for 1000 jobs. This means that ~1000 items were probably stored in the buffer.

### Before
Without this optimization, the `_store_in_buffer` method consumed 18.2 seconds (see the box highlighted in pink, below).
image

### After
With this optimization, the `_store_in_buffer` method consumed only 0.719 seconds (see box higlighted in pink). That's a 25x speedup for this case, but the speedup will depend heavily on the input data. I believe this optimization should be strictly faster in all cases.
image

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -24,6 +24,7 @@ Changed
 
  - Optimized internal function ``_mkdir_p`` (#421).
  - Optimized performance of job initialization (#422).
+ - Optimized performance of buffer storage (#428).
 
 [1.5.0] -- 2020-09-20
 ---------------------
--- a/signac/core/jsondict.py
+++ b/signac/core/jsondict.py
@@ -61,7 +61,7 @@ def _hash(blob):
         return m.hexdigest()
 
 
-def _get_filemetadata(filename):
+def _get_file_metadata(filename):
     try:
         return os.path.getsize(filename), os.path.getmtime(filename)
     except OSError as error:
@@ -83,7 +83,7 @@ def _store_in_buffer(filename, blob, store_hash=False):
     _JSONDICT_BUFFER[filename] = blob
     if store_hash:
         if not _BUFFERED_MODE_FORCE_WRITE:
-            _JSONDICT_META[filename] = _get_filemetadata(filename)
+            _JSONDICT_META[filename] = _get_file_metadata(filename)
         _JSONDICT_HASHES[filename] = _hash(blob)
     return True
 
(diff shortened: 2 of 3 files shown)


mikemhenry (approved):
LGTM! Nice optimization!

bdice:
Thanks @mikemhenry! I'll merge this after tests pass. In case you're curious, I added some integration tests to the PR and also added profiling results to the PR description above. It should help a lot with performance in signac-flow (which always runs in buffered mode) for large numbers of jobs.

bdice:
We have some test failures that are not reproducible on my machine. I pushed some very minor changes to `master` prior to this pull request that should not have affected testing, so it's possible that the errors were introduced prior to the first commit that I see with failures, 48fc4506b6f183c9e128bada09bc029b4d61a514. These seem to be intermittent failures and I can't reproduce them locally. Both CPython and PyPy are affected. This was the first instance of test failures I could find:
 https://app.circleci.com/pipelines/github/glotzerlab/signac/1543/workflows/60993bbc-d398-4284-92e3-95beee6b918d/jobs/9416

I believe all the errors are "FileExistsError", error number 17.
```
FileExistsError: [Errno 17] File exists: '/tmp/signac_dr2gdfti/wd/9bfd29df07674bc4aa960cf661b5acd2' -> '/tmp/signac_dr2gdfti/wd/c3e2fb11d9fba414c7a7da3ee3a9115f'
```

bdice:
Error number 17 is `EEXIST`. I altered some related error handling in #421, but I believed my changes would not cause problems. I am going to go ahead and merge this PR and file a separate issue / PR for these failures.

csadorf:
That's awesome!

PR #429, opened 2020-12-13
Optimize _dfs_convert by exiting early for common types.

bdice:
## Description
The internal function `_dfs_convert` recursively converts containers into their synced equivalents. That is, when reading a state point or document, the `_dfs_convert` function will change a `dict` into a synced dictionary.

**Observation:** Most signac jobs have state points and documents that are relatively flat. That is, most values in a state point or document are common non-container types (string, float, int, bool) and only a few are (dict, list).

I noticed that the code path taken for non-container types must exhaustively test whether the value is a `Mapping`, tuple, list, or NumPy type. Only at the very end it returns the raw value.

Matching the "fast path" to the expected distribution of data is a micro-optimization, but it pays off. Benchmarks are below.

## Motivation and Context
Speeds up all operations in signac that involved synced structures (e.g. reading data from a state point or document).

### Benchmark:
I measured the total time for checking status, running, and submitting a sample workflow of 1000 jobs in signac flow with 3 operations with pre/post conditions. The state points and documents had no nested dicts/lists.

**Before:** 86.0 seconds total, 7.749 seconds spent in `_dfs_convert`.
**After:** 78.3 seconds total, 1.378 seconds spent in `_dfs_convert`.

This suggests a speedup of ~5x in that function, but it will depend heavily on the data sizes and types used.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -25,6 +25,7 @@ Changed
  - Optimized internal function ``_mkdir_p`` (#421).
  - Optimized performance of job initialization (#422).
  - Optimized performance of buffer storage (#428).
+ - Optimized performance of creating/loading synced data structures (#429).
 
 [1.5.0] -- 2020-09-20
 ---------------------
--- a/signac/core/synceddict.py
+++ b/signac/core/synceddict.py
@@ -125,15 +125,13 @@ class _SyncedDict(MutableMapping):
                 for k in root:
                     ret[k] = root[k]
             return ret
-        elif type(root) is tuple:
-            return _SyncedList(root, self)
-        elif type(root) is list:
-            return _SyncedList(root, self)
+        elif type(root) in (list, tuple):
+            return _SyncedList(root, parent=self)
         elif NUMPY:
             if isinstance(root, numpy.number):
                 return root.item()
             elif isinstance(root, numpy.ndarray):
-                return _SyncedList(root.tolist(), self)
+                return _SyncedList(root.tolist(), parent=self)
         return root
 
     @classmethod
--- a/signac/core/synceddict.py
+++ b/signac/core/synceddict.py
@@ -112,7 +112,10 @@ class _SyncedDict(MutableMapping):
             )
 
     def _dfs_convert(self, root):
-        if type(root) is type(self):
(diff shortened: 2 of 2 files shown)


bdice:
Note to self: this PR needs a few updates, along with tests/changelog. Tests are failing because of #431, which should be solved in #433.
- NumPy float64s are registered as subclasses of `float`. Need to check if that is a problem or not.
- Should the early-exit check use `isinstance(x, float)` or `type(x) == float`? Even though `isinstance` is typically preferred, it might not be the right choice here, depending on the NumPy situation previously mentioned and/or performance considerations.

vyasr:
> Note to self: this PR needs a few updates, along with tests/changelog. Tests are failing because of #431, which should be solved in #433.
> 
>     * NumPy float64s are registered as subclasses of `float`. Need to check if that is a problem or not.
> 
>     * Should the early-exit check use `isinstance(x, float)` or `type(x) == float`? Even though `isinstance` is typically preferred, it might not be the right choice here, depending on the NumPy situation previously mentioned and/or performance considerations.

The issue of numpy float64s also being floats is definitely something to look into. If that's a problem, then you should use `type(x) is float` (rather than `==`). If it's not, then you'll probably want to stick with `isinstance`, you probably won't get performance benefits from directly checking types and if anything it might slow the code down.

bdice:
@vyasr I agree with your comments in general. A review of the specific implementation in this PR would be helpful. The short answer on NumPy float64 was this:
- Reading data from the disk will always be type `float` (because that's how the JSON is decoded)
- Using a method like "reset state point" will use whatever type is passed. I used `type(x) in (bool, float, int, type(None), str)` because I didn't want to capture the `numpy.float64` class (I wanted to pass it through the "normal" path that will call `x.item()` to extract the Python type)

vyasr:
@bdice makes sense, thanks. I didn't look at the code at all, I just responded to your previous comment without realizing that you had already addressed it. I'll aim to review in the next couple of days (before the weekend).

csadorf (approved):

vyasr (approved):
LGTM

PR #434, opened 2020-12-15
Document complexity of min_len_unique_id.

bdice:
## Description
I noticed in signac-flow that the function `Project.min_len_unique_id` was being called for every job operation, which made the code much slower because the function is O(N), resulting in an O(N^2) code path.

I decided to add a note to the function here, so that the runtime complexity is more obvious. I am also opening a PR to signac-flow to address the slowdown there.

I also updated `signac find` so it only gets the minimum length of unique id if it will be used. This is the only place where the function was called in the code base (outside of tests).

This shouldn't need a changelog line since it's a minor change to the docs.

## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -410,6 +410,9 @@ class Project:
     def min_len_unique_id(self):
         """Determine the minimum length required for a job id to be unique.
 
+        This operation's runtime scales with the number of jobs in the
+        workspace.
+
         Returns
         -------
         int
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -410,7 +410,7 @@ class Project:
     def min_len_unique_id(self):
         """Determine the minimum length required for a job id to be unique.
 
-        This operation's runtime scales with the number of jobs in the
+        This method's runtime scales with the number of jobs in the
         workspace.
 
         Returns
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -373,6 +373,11 @@ def main_find(args):
     """Handle find subcommand."""
     project = get_project()
 
+    len_id = 6
+    if args.one_line:
+        # Only get the project's minimum length of unique id if it is needed.
+        len_id = max(len_id, project.min_len_unique_id())
+
     # --show = --sp --doc --pretty 3
     # if --sp or --doc are also specified, those subsets of keys will be used
     if args.show:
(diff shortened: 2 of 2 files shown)


cbkerr:
Is the output of `Project.min_len_unique_id()` something we can calculate only after we run anything that changes the number of jobs and store in a file at the project level? (Now that I'm aware that we have various project-level cache files)

This would be for a new issue.

signac/contrib/project.py
```diff
@@ -410,6 +410,9 @@ def _check_schema_compatibility(self):
     def min_len_unique_id(self):
         """Determine the minimum length required for a job id to be unique.
 
+        This operation's runtime scales with the number of jobs in the
```
cbkerr, line 413:
```suggestion
        This function's runtime scales with the number of jobs in the
```
I suggest reserving "operation" for functions related to signac-flow.

cbkerr (approved):
The changes all make sense to me ("only needed" = "need to condense output for `--oneline` command"), although I didn't confirm that the revised print statement prints as expected.

Looks good with a small word change commented on.

bdice:
> 
> 
> Is the output of `Project.min_len_unique_id()` something we can calculate only after we run anything that changes the number of jobs and store in a file at the project level? (Now that I'm aware that we have various project-level cache files)
> 
> This would be for a new issue.

@cbkerr Good idea, and I considered this when creating the PR. Unfortunately there's no safe way to cache / store this value. The project's `min_len_unique_id` can change at any time if new jobs are added or removed, so this has to be freshly computed from the current list of all job ids. That in turn requires a call to `_find_job_ids()` :arrow_right: `_job_dirs()` ➡️ `os.listdir`. If we had a way to know that no jobs had been added/removed, then this could be cached, but there's no way to know that without listing the directories in the workspace (for instance, a user could be accessing this project in a separate process or from another computer via the network).

signac/contrib/project.py
bdice: Definitely, that's an oversight on my part. I'm going to use the word "method" because that's the usual name for a function attached to an object.

bdice:
> although I didn't confirm that the revised print statement prints as expected.

I checked it manually. 👍

tcmoore3 (approved):
LGTM!

PR #443, opened 2020-12-22
Optimize H5Store init.

bdice:
## Description
Optimizes the `H5Store` initializer to use `os.path.abspath` instead of `os.path.realpath`. As mentioned in #442, the cost of `realpath` is fairly high and it isn't necessary to use a "real" path in this case.

(Note that this optimization would *not* apply to `job.document` or anything that must interface with signac's buffering logic, in my understanding. Buffering relies on "real" paths to ensure that data is correctly loaded in the buffer.)

## Motivation and Context
Speeds up H5Store initialization.

I noticed this optimization while using a signac-flow project with a condition that checks for a key in `job.data`. For my simple test case with 157 jobs, the time spent in the condition function went from 2.72 seconds to 2.15 seconds (21% faster).

Note: Another 20-30% of the time could be trimmed by adding lazy state point loading and avoiding the state point validation when accessing the `job.data` (which never needs to know the state point).

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -14,6 +14,7 @@ Changed
 +++++++
 
  - Optimized job hash and equality checks (#442).
+ - Optimized ``H5Store`` initialization (#443).
 
 
 [1.5.1] -- 2020-12-19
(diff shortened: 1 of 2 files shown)


kidrahahjo (approved):
I don't have expertise in the buffering logic currently but, in the interest of this PR, this is a straight forward change, so I am approving this PR.

PR #451, opened 2020-12-29
Optimization: job/project initialization.

bdice:
## Description
This is a follow-up PR to #239, but it may be possible to adopt this set of changes separately from #239 if that PR stalls. I recommend that reviewers focus on #239 first and then handle this PR second.

This PR refactors the `__init__` methods of the `Job` and `Project` classes so that less work is done during initialization, instead deferring (and often caching) that work in a property. This requires no API breaking changes, because the existing methods and properties were enough to handle this change.

One tricky part of this is that I added a "mutation hook" to the Project class. In 99% of users' workflows, the Project config (which contains information about the workspace directory name, project name, etc.) does not need to be modified once the project is instantiated. While in-memory config mutation is allowed, mutating the configuration of a `Project` instance has been deprecated since version 1.3. However, there is a significant performance penalty to allowing this mutation: every job that is opened must determine its workspace directory by first checking the project's workspace directory and joining its own id. That requires an access of the project configuration (and more importantly, various path manipulations) for every job, and it shows up quite obviously when profiling iteration over a large number of jobs. To solve this, I added a `_mutate_hook` method that will reset the values of any cached properties that are derived from the project configuration if the configuration is edited in memory.

I have verified that the existing tests should cover these changes, particularly the changes to the `Project` cached properties and config mutation hook.

## Motivation and Context
This makes the "lazy state point loading" in #239 even lazier, and thus extremely fast for many common operations (e.g. in signac-flow).

| Test (best of 3, seconds)            | master (426d2ed) | feature/lazy-statepoint-loading (9a12405) | feature/lazy-job-init (32fef57) |
|--------------------------------------|------------------|-------------------------------------------|---------------------------------|
| `[job for job in pr]`; N=30,000      | 1.014            | 0.471 (2.15x)                             | 0.141 (7.19x)                    |
| `[job.sp() for job in pr]`; N=30,000 | 1.442            | 2.208 (0.653x)                            | 1.381 (1.04x)                   |

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


signac/contrib/job.py
```diff
@@ -91,32 +91,30 @@ def __init__(self, project, statepoint=None, _id=None):
             raise ValueError("Either statepoint or _id must be provided.")
 
         if statepoint is not None:
-            # Set state point if provided
+            # Set state point if provided. The id is computed from the state point.
             self._statepoint = SyncedAttrDict(statepoint, parent=_sp_save_hook(self))
+            # Validate the state point and recursive convert to supported types.
```
bdice, line 96:
```suggestion
            # Validate the state point and recursively convert to supported types.
```

signac/contrib/job.py
```diff
@@ -219,10 +219,9 @@ def reset_statepoint(self, new_statepoint):
         # Update this instance
         self.statepoint._data = dst.statepoint._data
         self._id = dst._id
-        self._wd = dst._wd
-        self._fn_doc = dst._fn_doc
+        self._wd = None
         self._document = None
-        self._data = None
+        self._stores = None
```
bdice, line 224: This is a bug left in from long ago. The `_data` attribute doesn't exist, it's called `_stores`.

signac/contrib/project.py
```diff
@@ -1196,16 +1214,18 @@ def write_statepoints(self, statepoints=None, fn=None, indent=2):
         with open(fn, "w") as file:
             file.write(json.dumps(tmp, indent=indent))
 
-    def _register(self, job):
+    def _register(self, _id, statepoint):
```
bdice, line 1217: I changed this method to reduce overhead. The id and state point are already known in all the cases where `project._register` is called, but the previous implementation had to explicitly perform a conversion back from the synced state point to a plain dict. That accounted for some of the slowdown between `master` and `feature/lazy-statepoint-loading` in #239 when opening jobs by state point.

signac/contrib/job.py
```diff
@@ -644,7 +647,7 @@ def remove(self):
                     if not error.errno == errno.ENOENT:
                         raise error
                 self._document = None
-            self._data = None
+            self._stores = None
```
bdice, line 650: This is a bug left in from long ago. The `_data` attribute doesn't exist, it's called `_stores`.

bdice:
I'm tagging @klywang and @cbkerr as reviewers because they expressed specific interest in performance-related improvements.

signac/contrib/project.py
```diff
@@ -196,7 +218,12 @@ class Project:
     def __init__(self, config=None, _ignore_schema_version=False):
         if config is None:
             config = load_config()
-        self._config = _ProjectConfig(config)
+        self._config = _ProjectConfig(config, _mutate_hook=partial(_mutate_hook, self))
```
cbkerr, line 221: I don&rsquo;t understand this line `self._config = _ProjectConfig(config, _mutate_hook=partial(_mutate_hook, self))`. I know that partial will cause the `_mutate_hook` argument of the `_ProjectConfig` to be `_mutate_hook` applied to itself, but where is `_mutate_hook` defined? It seems not defined in this scope. Does it take the value of `_mutate_hook` of the `self` within `_ProjectConfig`?

signac/contrib/project.py
```diff
@@ -376,10 +382,12 @@ def id(self):
...
 
         """
-        try:
-            return str(self.config["project"])
-        except KeyError:
-            return None
+        if self._id is None:
+            try:
+                self._id = str(self.config["project"])
+            except KeyError:
```
cbkerr, line 388: Do we need to raise an error here?

cbkerr (changes requested):
I've studied the diffs (less the surrounding code) and my main concern is the naming issue. See my inline comments.

signac/contrib/project.py
```diff
@@ -156,9 +169,18 @@ def __setitem__(self, key, value):
             )
 
             assert version.parse(__version__) < version.parse("2.0")
+        if self._mutate_hook is not None:
+            self._mutate_hook()
         return super().__setitem__(key, value)
 
 
+def _mutate_hook(project):
```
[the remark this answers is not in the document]
bdice: Good idea. I changed the specific method name to `_invalidate_config_cache` and left the `_ProjectConfig` attribute as `_mutate_hook`, to indicate that it's a cache invalidation triggered by config mutation.

signac/contrib/project.py
```diff
@@ -139,9 +140,21 @@ def find_job_ids(self, filter=None, doc_filter=None):
...
 class _ProjectConfig(Config):
-    """Extends the project config to make it immutable."""
+    r"""Extends the project config to make it immutable.
 
-    def __init__(self, *args, **kwargs):
+    Parameters
+    ----------
+    \*args :
+        Forwarded to :class:`~signac.common.config.Config` constructor.
+    _mutate_hook : callable
```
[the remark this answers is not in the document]
bdice: A callable can be a function, class with a `__call__` method, etc. The `_mutate_hook` is a [callback](https://en.wikipedia.org/wiki/Callback_(computer_programming)). Sometimes callbacks are also called hooks. Would it make more sense to name this `_mutate_callback`?

signac/contrib/project.py
bdice: I'll break it down: `_ProjectConfig(config, ...)` calls the constructor of the `_ProjectConfig` class. We provide a keyword argument named `_mutate_hook` with the value `partial(_mutate_hook, self)`. That partial takes the function `_mutate_hook` (now renamed to `_invalidate_config_cache`) and binds the first parameter of that function to `self`. The `partial` returns a callable with no free parameters.

This could equivalently be written as:
```python
func = lambda: _invalidate_config_cache(self)

self._config = _ProjectConfig(config, _mutate_hook=func)
```
or
```python
def func():
    _invalidate_config_cache(self)

self._config = _ProjectConfig(config, _mutate_hook=func)
```
where `self` is defined in the scope of `Project.__init__` and that scope is shared when the function is passed along.

signac/contrib/project.py
bdice: No. The previous behavior was to return `None` if the `"project"` key is missing in the config. This change retains the previous behavior, because it `pass`es and returns `self._id`, which is already `None`. It may be more clear to make that explicit, so I'll replace `pass` with `return None`.

signac/contrib/project.py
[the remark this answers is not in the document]
cbkerr: Perfect!

signac/contrib/project.py
cbkerr: ah that does help

signac/contrib/project.py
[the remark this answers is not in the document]
cbkerr: The "hook" terminology is okay; I think this was residual confusion from `_mutate_hook` referring to a method and a property.

signac/contrib/project.py
```diff
@@ -156,9 +169,18 @@ def __setitem__(self, key, value):
             )
 
             assert version.parse(__version__) < version.parse("2.0")
+        if self._mutate_hook is not None:
+            self._mutate_hook()
```
cbkerr, line 173: Just checking my understanding: this line `self._mutate_hook()` calls the callable that was passed in.

bdice: Correct. `self._mutate_hook` is a callable, and the `()` calls it with no arguments.

cbkerr:
Will `_mutate_hook` be deprecated in version 2.0 when mutating the configuration of a `Project` will be deprecated? If so, should we add a flag to deprecate it then too?

signac/contrib/project.py
cbkerr: thank you!

bdice:
> 
> 
> Will `_mutate_hook` be deprecated in version 2.0 when mutating the configuration of a `Project` will be deprecated? If so, should we add a flag to deprecate it then too?

Not necessary - `_ProjectConfig` is a private class and `_invalidate_config_cache` is a private function. We don't have to worry about deprecating/changing private APIs because users should not rely on them. The existing deprecation warning shown when `_ProjectConfig` is mutated is sufficient to indicate that the behavior will change.

cbkerr (approved):

bdice:
@csadorf @tommy-waltmann @klywang I would appreciate a second review if you are able. @cbkerr Thank you for the review.

csadorf (approved):
Looks good! 👍 Sorry for the delay.

PR #453, opened 2020-12-30
Feature/synced collection/optimization

vyasr:
## Description

This PR optimizes all JSON-based SyncedCollections, including the buffered mode. I benchmarked a number of operations comparing the new JSONDict to the old one on two machines, a 2020 Macbook Pro and our Linux workstations using a slow network file system (as opposed to faster scratch space). I'm happy to share some plots for those interested. Brief summary: depending on the operation being benchmarked, whether buffering is enabled, and the operating system, I see performance improvements in the range of 1.3x-2.7x. In general the biggest performance improvements are observed for operations that change the dict since they touch all of the code paths that I optimized. Some more detailed information:

- Overall I see bigger improvements on the Mac than on the Linux machines. I have not benchmarked the exact bottlenecks on the Linux machines, but I assume that these differences are mostly related to file I/O; the Mac has an SSD, so the I/O costs will be much lower and we can more directly observe the differences in the code complexity.
- Operations that mutate dicts (__setitem__ or update) are 2.2-2.7x faster on the Mac, whereas they are 1.2-2.2x faster on the Linux workstation. 
- Operations that only access data without changing it (__getitem__, keys, values, __call__) are 1.4-2.2x faster on the Mac. These operations gain a little less performance than the mutating operations because they require only a save, not a load call. On the Linux workstations this drops to more like a 1.2-1.8x speed bump.
- On the Mac, buffering only increases performance by about 20% for most operations. Moreover, for some operations it is actually slower. The new JSONDict is also faster than the old JSONDict, even when the new JSONDict does not have buffering activated. Again, this is largely a function of the fast SSD. However, it reveals serious bottlenecks with our buffering strategy that we could exploit; I'll make another issue to discuss that and link it when I do.
- On the Linux workstations, buffering increases performance by up to an order of magnitude, so that's more as expected. The old buffered JSONDict is 4x faster than the new unbuffered JSONDict, but once buffering is enabled the new JSONDict wins by about 2.5x again.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/core/synced_collections/synced_attr_dict.py
+++ b/signac/core/synced_collections/synced_attr_dict.py
@@ -126,31 +126,37 @@ class SyncedAttrDict(SyncedCollection, MutableMapping):
             self._data.clear()
         elif _mapping_resolver.get_type(data) == "MAPPING":
             with self._suspend_sync():
-                # This loop avoids rebuilding existing synced collections for performance.
-                for key in data:
-                    if key in self._data:
-                        if data[key] == self._data[key]:
+                for key, new_value in data.items():
+                    try:
+                        # The most common usage of SyncedCollections is with a
+                        # single object referencing an underlying resource at a
+                        # time, so we should almost always find that elements
+                        # of data are already contained in self._data, so EAFP
+                        # is the best choice for performance.
+                        existing = self._data[key]
+                    except KeyError:
+                        # If the item wasn't present at all, we can simply
+                        # assign it.
+                        self._validate({key: new_value})
+                        self._data[key] = self._from_base(new_value, parent=self)
+                    else:
+                        if new_value == existing:
                             continue
-                        if _sc_resolver.get_type(self._data[key]) == "SYNCEDCOLLECTION":
+                        if _sc_resolver.get_type(existing) == "SYNCEDCOLLECTION":
                             try:
-                                # The key must already be valid to be in this
-                                # object, so we only need to validate the data.
-                                self._validate(data[key])
-                                self._data[key]._update(data[key])
+                                existing._update(new_value)
                                 continue
                             except ValueError:
                                 pass
 
-                    # Fall through if:
-                    #    1) The key is not currently in self, OR
-                    #    2) The key is in self, AND
-                    #        1) The existing value is not a SyncedCollection
-                    #           (in which case we would have tried to update it, OR
-                    #        2) The existing value is a SyncedCollection, but
-                    #           the new value is not a compatible type for _update.
-                    # Validate each element as it is assigned.
-                    self._validate({key: data[key]})
-                    self._data[key] = self._from_base(data[key], parent=self)
+                        # Fall through if the new value is not identical to the
+                        # existing value and
+                        #     1) The existing value is not a SyncedCollection
+                        #        (in which case we would have tried to update it, OR
+                        #     2) The existing value is a SyncedCollection, but
+                        #       the new value is not a compatible type for _update.
+                        self._validate({key: new_value})
+                        self._data[key] = self._from_base(new_value, parent=self)
 
                 to_remove = [key for key in self._data if key not in data]
                 for key in to_remove:
--- a/signac/core/synced_collections/synced_attr_dict.py
+++ b/signac/core/synced_collections/synced_attr_dict.py
@@ -99,9 +99,7 @@ class SyncedAttrDict(SyncedCollection, MutableMapping):
         bool
 
         """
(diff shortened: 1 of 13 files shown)


vyasr:
@bdice suggested that @klywang @cbkerr might be interested in seeing these results.

signac/core/synced_collections/synced_attr_dict.py
```diff
@@ -114,27 +121,43 @@ def _update(self, data=None):
...
-                remove = set()
-                for key in self._data:
-                    if key not in data:
-                        remove.add(key)
-                for key in remove:
+
+                        # Fall through if the new value is not identical to the
+                        # existing value and
+                        #     1) The existing value is not a SyncedCollection
+                        #        (in which case we would have tried to update it, OR
```
bdice, line 153:
```suggestion
                        #        (in which case we would have tried to update it), OR
```

signac/core/synced_collections/utils.py
```diff
@@ -0,0 +1,138 @@
...
+
+class AbstractTypeResolver:
+    """Mapping between recognized types and their abstract parents.
+
+    Synced collections are heavily reliant on checking the types of objects to
+    determine the appropriate type of behavior in various scenarios. For maximum
+    generality, most of these checks use the ABCs defined in :py:mod:`collections.abc`.
+    The price of this flexibility is that `isinstance` checks with these classes
+    are very slow because the ``__instancecheck__`` hooks are implemented in pure
+    Python and reuqire checking many different cases.
```
bdice, line 25:
```suggestion
    Python and require checking many different cases.
```

signac/core/synced_collections/utils.py
```diff
@@ -0,0 +1,138 @@
...
+            argument to the constructor. If the object's type cannot be identified,
+            will return ``None``.
+
+        """
+        dtype = type(obj)
+        enum_type = None
+        try:
+            enum_type = self.type_map[dtype]
+        except KeyError:
+            for adt, id_func in self.abstract_type_identifiers.items():
```
bdice, line 79: I don't know what `adt` means. Please use a more descriptive name.

signac/core/synced_collections/utils.py
```diff
@@ -0,0 +1,138 @@
...
+
+        Returns
+        -------
+        str
+            The name of the type, where valid types are the keys of the dict
+            argument to the constructor. If the object's type cannot be identified,
+            will return ``None``.
+
+        """
+        dtype = type(obj)
```
bdice, line 74: To avoid confusion with NumPy dtypes, could we call this `obj_type` instead?

signac/core/synced_collections/utils.py
```diff
@@ -0,0 +1,138 @@
...
+    Warnings
+    --------
+    JSON encoding of numpy arrays is not invertible; once encoded, reloading
+    the data will result in converting arrays to lists and numpy numbers into
+    ints or floats.
+
+    """
+    if NUMPY:
+        if isinstance(o, numpy.number):
+            return o.item()
```
bdice, line 103: I see Codecov has marked a few places that are untested. I believe we have tests for the similar places where this is already done in signac, but we need to port those tests so they cover this synced collections code.

bdice (commented):
This is nice, @vyasr! The design of the abstract type resolver is unusual (because it's handling an unusual problem) but I think it will be fine to use in the long term. For now, I'm just going to leave a few small comments from a quick read-through.

cbkerr:
Thanks for tagging me @vyasr. I'm trying to understand `SyncedCollection` before proceeding on this review. I tried to clarify some of my confusion [here](https://github.com/glotzerlab/signac/issues/249#issuecomment-753122358).

signac/core/synced_collections/utils.py
vyasr: The relevant tests are actually already in place. The reason that this isn't tested here any more is because numpy arrays are now converted further upstream. The new SyncedList type is much more sophisticated than the original one, and numpy arrays are converted to lists as soon as they are assigned.

I suppose that leaves a question of whether we need to bother including this functionality in the custom JSON encoder at all. I'd be fine with removing it, I just included it for completeness. What do you think?

vyasr: For now I left it in place in case it comes in handy later. I've added explicit tests of the utils (both the type resolver and the encoder) so I think we're set here.

PR #455, opened 2020-12-30
Handle non-Job objects in Job equality check.

bdice:
## Description
I caught a couple minor errors in signac-flow tests (to be addressed separately) that were exposed by the changes in #442. This unrelated problem in signac-flow indicated an issue in the new `__eq__` check, which doesn't correctly handle comparisons to non-Job objects. This PR fixes that bug so the `Job` class correctly adheres to the Python data model.

Specifically, the old `__eq__` behavior did nothing to ensure that the objects being tested for equality were Job objects. If a job object and a tuple were compared, the previous `Job` equality test just checked if `hash(job) == hash(tuple_of_data)`, which hid the test failures in signac-flow until #442 was merged and errors were raised by the comparison. However, the comparison _should not_ raise errors or return `False`, and should instead return `NotImplemented` [according to the Python data model](https://docs.python.org/3/reference/datamodel.html#object.__eq__) and some StackOverflow posts I found.

## Motivation and Context
Fixes minor bug in `__eq__` implementation.

[This issue from another library](https://github.com/googleapis/google-cloud-python/issues/3455) and [this section of the Python documentation](https://docs.python.org/3/reference/datamodel.html#object.__eq__) are helpful for explaining the motivation of the change.

I added tests that cover a range of cases.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/tests/test_job.py
+++ b/tests/test_job.py
@@ -154,12 +154,15 @@ class TestJob(TestJobBase):
 
         non_job = NonJob(job)
         assert job != non_job
+        assert non_job != job
 
         sub_job = JobSubclass(job)
         assert job == sub_job
+        assert sub_job == job
 
         job2 = self.project.open_job({"a": 0})
         assert job == job2
+        assert job2 == job
 
     def test_isfile(self):
         job = self.project.open_job({"a": 0})
--- a/tests/test_job.py
+++ b/tests/test_job.py
@@ -131,6 +131,36 @@ class TestJob(TestJobBase):
         job = self.project.open_job({"a": 0})
         assert str(job) == job.id
 
+    def test_eq(self):
+        job = self.project.open_job({"a": 0})
+
+        # Make sure that Jobs can only be equal to other Job instances.
+        class NonJob:
+            """Minimal class that cannot be compared with Job objects."""
+
+            def __init__(self, job):
+                self.id = job.id
+                self._workspace = job.workspace()
+
+        class JobSubclass(Job):
+            """Minimal subclass that can be compared with Job objects."""
+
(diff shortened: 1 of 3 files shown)


tests/test_job.py
```diff
@@ -131,6 +131,39 @@ def test_str(self):
...
+            def __init__(self, job):
+                self._id = job.id
+                self._workspace = job.workspace()
+
+            def workspace(self):
+                return self._workspace
+
+        non_job = NonJob(job)
+        assert job != non_job
+        assert non_job != job
```
atravitz, line 157: Why does this ensure that we have `NotImplemeted` as opposed to `False`?

bdice: I'm testing both `a != b` and `b != a` because of the order in which Python resolves `__eq__` operators. This section of the Python docs explains why both tests are relevant here, and why we return `NotImplemented` here: https://docs.python.org/3/reference/datamodel.html#object.__eq__

Specifically, the order of resolution is controlled by this rule: "If the operands are of different types, and right operand’s type is a direct or indirect subclass of the left operand’s type, the reflected method of the right operand has priority, otherwise the left operand’s method has priority."

bdice: Also, I used this issue as a reference implementation because it appeared to correct the equality behavior that I was unsure about for subclasses (`JobSubclass`) and objects of different types with similar APIs (`NonJob`): https://github.com/googleapis/google-cloud-python/issues/3455

tcmoore3 (approved):
Good find, and nice fix!

Issue #397, opened 2020-10-08
Exceeding the buffer size in parallel context causes errors during flush


bdice:
### Description

I have a project with lots of jobs and large job documents. I tried to check the status of the project using signac-flow. One of the conditions involves checking for a key in the job document. The status check attempts to operate in parallel over all jobs. The status check _also_ operates in signac's "buffered" mode, to minimize I/O. This combination of parallel + buffering appears to be unsafe if the buffer size is exceeded. **I would guess the buffer is also susceptible to race conditions in parallel reading/writing, but the specific issue being reported here is read-only.**

The buffer metadata appears to be kept in global state: https://github.com/glotzerlab/signac/blob/d639e682ca7ebaff781d68621a2d86cc26de04b9/signac/core/jsondict.py#L23-L30

If signac-flow's (parallel) status checks overflow the 32 MB buffer size, a flush is forced, but the metadata doesn't seem to be in sync across threads/processes. Thus, the flush (perhaps the second flush? unsure) can't find the keys corresponding to filenames that no longer exist in the metadata / buffer. I have a reproduction script below.

I have verified that the following changes will prevent the problem:
1. Increasing the buffer size (so data fits in buffer)
2. Decreasing the data size (so data fits in buffer)

As for solutions, it's not obvious how this should be fixed. This seems like a design problem in our buffering scheme, and whatever we do in #363 should probably handle this. Some possibilities:
1. Make sure every process has its own buffer. This will fix the issue with flushing but would cause other issues with inconsistency between concurrent processes.
2. Introduce a locking mechanism for writing to the global buffer? Not sure if this is a solution, would need to think more.
3. Force flushes to synchronize across processes?

### To reproduce

```python
# init.py
import numpy as np
import signac
from tqdm import tqdm


# Create project files
project = signac.init_project("overflow")

print('Creating jobs...')
for i in tqdm(range(1000)):
    job = project.open_job({"i": i}).init()
    # Need enough data in the job documents to overflow the default buffer size of 32 MB
    job.doc['some_data'] = np.random.rand(100, 100).tolist()
    job.doc['parity'] = "odd" if i % 2 else "even"
```

```python
# project.py

import numpy as np
import signac
from tqdm import tqdm
from flow import FlowProject


class Project(FlowProject):
    pass


@Project.operation
@Project.pre(lambda job: job.doc['parity'] == "even")
@Project.post.isfile("out.txt")
def write_id_if_even(job):
[...]

PR #463, opened 2021-01-05
Feature/synced collection/thread safety

vyasr:
## Description

This PR makes operations on SyncedAttrDict and SyncedList objects thread safe. It provides opt-in mechanisms for thread safety for different backends to support thread-safe execution. It also allows the user to enable or disable thread safety. This feature may be inadvisable in the long term, but is convenient for benchmarking purposes and so will be relevant while we integrate these objects into the core of signac.

## Motivation and Context

Note that this does not address #397, which requires making buffer loads/saves thread safe; that will be addressed in a separate PR once #462 is merged.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/core/synced_collections/synced_attr_dict.py
+++ b/signac/core/synced_collections/synced_attr_dict.py
@@ -215,23 +215,19 @@ class SyncedAttrDict(SyncedCollection, MutableMapping):
             )
 
     def keys(self):  # noqa: D102
-        with self._thread_lock():
-            self._load()
+        self._load()
         return self._data.keys()
 
     def values(self):  # noqa: D102
-        with self._thread_lock():
-            self._load()
+        self._load()
         return self._to_base().values()
 
     def items(self):  # noqa: D102
-        with self._thread_lock():
-            self._load()
+        self._load()
         return self._to_base().items()
 
     def get(self, key, default=None):  # noqa: D102
-        with self._thread_lock():
-            self._load()
+        self._load()
         return self._data.get(key, default)
 
     def pop(self, key, default=None):  # noqa: D102
--- a/signac/core/synced_collections/synced_attr_dict.py
+++ b/signac/core/synced_collections/synced_attr_dict.py
@@ -65,7 +65,8 @@ class SyncedAttrDict(SyncedCollection, MutableMapping):
                     key: self._from_base(data=value, parent=self)
                     for key, value in data.items()
                 }
-            self._save()
+            with self._thread_lock():
+                self._save()
 
     def _to_base(self):
         """Convert the SyncedDict object to Dictionary.
@@ -174,11 +175,12 @@ class SyncedAttrDict(SyncedCollection, MutableMapping):
         # directly set using those rather than looping over data.
         data = {key: value}
         self._validate(data)
-        self._load()
-        with self._suspend_sync():
-            for key, value in data.items():
-                self._data[key] = self._from_base(value, parent=self)
-        self._save()
+        with self._thread_lock():
+            self._load()
+            with self._suspend_sync():
+                for key, value in data.items():
+                    self._data[key] = self._from_base(value, parent=self)
+            self._save()
 
     def reset(self, data=None):
         """Update the instance with new data.
@@ -203,7 +205,8 @@ class SyncedAttrDict(SyncedCollection, MutableMapping):
                     key: self._from_base(data=value, parent=self)
                     for key, value in data.items()
                 }
-            self._save()
+            with self._thread_lock():
+                self._save()
         else:
             raise ValueError(
                 "Unsupported type: {}. The data must be a mapping or None.".format(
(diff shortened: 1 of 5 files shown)


b-butler (approved):
I haven't looked too much at this code, but the threading logic seems correct. The use of context-managers and the implementation are fairly clean as well. We may want someone who has looked at this code more to approve this as well to be safe.

signac/core/synced_collections/synced_attr_dict.py
```diff
@@ -212,36 +215,43 @@ def reset(self, data=None):
...
+            self._load()
+            ret = self._data.popitem()
+            self._save()
         return ret
 
     def clear(self):  # noqa: D102
         self._data = {}
-        self._save()
+        with self._thread_lock():
+            self._save()
```
bdice, line 254: Do we want to overwrite the internal contents of `self._data` before the thread lock is acquired? I'm not sure.

signac/core/synced_collections/synced_collection.py
```diff
@@ -26,6 +28,27 @@
...
+    """A nullary context manager to simulate a lock in backends that don't support them."""
+
+    def __enter__(self):
+        pass
+
+    def __exit__(self, type, value, traceback):
+        pass
+
+
+@contextmanager
```
bdice, line 41: Based on [this StackOverflow post](https://stackoverflow.com/questions/26152934/why-the-staggering-overhead-50x-of-contextlib-and-the-with-statement-in-python), I think it would be better to have a class implementing `__enter__` and `__exit__` instead of using `contextmanager` here. I was able to reproduce these benchmarks on my own machine, and the fastest way would look like:
```python
class ManualCatchContext(object):
    def __enter__(self):
        # Acquire lock
        pass

    def __exit__(self, exc_type, exc_val, exc_tb):
        # Release lock
        return True

preinstantiated_manual_catch_context = ManualCatchContext() 
def manual_with_catch_cache(fn):
    with preinstantiated_manual_catch_context:
        fn()
```

signac/core/synced_collections/synced_list.py
```diff
@@ -183,58 +185,69 @@ def reset(self, data=None):
...
+            self._load()
+            with self._suspend_sync():
+                self._data.remove(self._from_base(data=value, parent=self))
+            self._save()
 
     def clear(self):  # noqa: D102
         self._data = []
-        self._save()
+        with self._thread_lock():
+            self._save()
```
bdice, line 253: Same question as above - can `self._data` be changed without a lock?

tests/test_synced_collections/synced_collection_test.py
```diff
@@ -398,6 +398,23 @@ class A:
             with pytest.raises(TypeError):
                 synced_collection[key] = testdata
 
+    def test_multithreaded(self, synced_collection):
```
bdice, line 401: Also needs tests that use the `enable_multithreading` / `disable_multithreading` methods.

bdice (commented):
I have a few comments that would be good to address before merging.

vyasr:
@bdice asked me offline why the read operations in this class need to acquire the mutex. The reason is that the default implementation of file writing in the JSONCollection class is not atomic. Unless we set `write_concern=True` in the constructor, which triggers a "write to tempfile->copy" sequence, the modification of the file when saving is not atomic and therefore can lead to reads of the invalid data. Bradley wrote some quick benchmarks that we tested on various machines to find that safer sequence is anywhere from 15-50% slower depending on the filesystem being tested.

The solution I'm planning to implement is to simply force the safer (atomic) write operation whenever threading is enabled. That seems like a much cleaner alternative than forcing normally thread-safe read operations to serialize execution.

signac/core/synced_collections/synced_collection.py
vyasr: Yes, I've seen these benchmarks. Did you try and see what effect this has on the overall performance of the code? I'd like to address this in a separate PR that handles context manager performance more generally since we use similar context managers for buffering and suspending synchronization. If contextlib is really a performance bottleneck on the rest of the code then I'd like to just define a simpler contextmanager factory of our own and use it consistently.

bdice: I haven’t tried this explicitly with the new SyncedCollections code. I _have_ seen this contextmanager overhead appear in the benchmarks I’ve been running for signac-flow. That uses the current implementation and not SyncedCollections, but I suspect the optimization will offer a measurable speedup.

signac/core/synced_collections/synced_attr_dict.py
vyasr: Doing so should be safe. In theory it's better for performance, since the Python interpreter could do that assignment concurrently with I/O on other threads. In practice it'll never be noticeable, I just aimed to lock the smallest portions of the code necessary for thread safety to minimize the impact everywhere and this is what happens here.

signac/core/synced_collections/synced_collection.py
vyasr: Got it. I'll make an issue of this then.

tests/test_synced_collections/synced_collection_test.py
vyasr: Done.

PR #467, opened 2021-01-07
Fix: don't delete invalid statepoints

bdice:
## Description
The (unreleased) lazy loading code recently introduced in #239 and/or #451 appears to delete invalid state point files if the state point on disk is invalid AND `job.init()` is called before `job.statepoint`. In my understanding, the only case where a state point file should be deleted is if the initial file creation/write operation fails during `job.init()`.

## Types of Changes
- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -537,6 +537,8 @@ class Job:
                 except OSError as error:
                     if error.errno not in (errno.EEXIST, errno.EACCES):
                         raise
+            except JobsCorruptedError:
+                raise
             except Exception as error:
                 # Attempt to delete the file on error, to prevent corruption.
                 try:
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -337,10 +337,9 @@ class Job:
 
         """
         if self._statepoint is None:
-            # Load state point manifest lazily
+            # Load state point manifest lazily and assign to
+            # self._statepoint
             statepoint = self._check_manifest()
-            self._statepoint = SyncedAttrDict(statepoint, parent=_sp_save_hook(self))
-
             # Update the project's state point cache when loaded lazily
             self._project._register(self.id, statepoint)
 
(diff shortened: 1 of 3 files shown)


tcmoore3 (commented):
I'm confused what behavior is actually affected by this PR, besides setting the state point in `_check_manifest`. How do the changes stop corrupt jobs from being deleted if the state point file is invalid?

bdice:
> 
> 
> I'm confused what behavior is actually affected by this PR, besides setting the state point in `_check_manifest`. How do the changes stop corrupt jobs from being deleted if the state point file is invalid?

@tcmoore3 The specific case that caused problems was introduced in previous PRs that added lazy loading. The crucial change is here:
https://github.com/glotzerlab/signac/blob/09a6677a963185634d1060b9495f6001bdadf4d7/signac/contrib/job.py#L529-L534

The previous code (before lazy loading) would already know the state point (it wasn't loaded lazily) and thus could not raise a `JobsCorruptedError` exception here if the data was invalid. However, any exception that _did_ get raised during a dump of the state point to disk was assumed to be an issue with file creation/saving (probably an `OSError`) and would trigger a rollback (deletion of the state point file). The change in this PR is to catch any `JobsCorruptedError` that is raised by calling `self.statepoint()` if the state point was loaded lazily _during `init()`_ but is corrupted on disk. That case should _not_ trigger rollback (file deletion), because we want to grant the user an opportunity to fix the file on disk if the state point is corrupt but not beyond manual repair (whatever that manual repair should look like is partially up to the user and is not within signac's scope -- the important thing is that signac shouldn't just delete the file when raising an error about its invalidity).

The test that I added also provides an example case of what behavior to expect when data is invalid. This fix is simply an effort to provide additional safety when handling invalid states and it is not a case that should arise in normal use.

signac/contrib/job.py
```diff
@@ -569,6 +574,8 @@ def _check_manifest(self):
         manifest = self._read_manifest()
         if calc_id(manifest) != self.id:
             raise JobsCorruptedError([self.id])
+        if self._statepoint is None:
+            self._statepoint = SyncedAttrDict(manifest, parent=_sp_save_hook(self))
```
bdice, line 578: This is an optimization: if we have to load and validate the manifest, we might as well set the job's lazy-loaded state point here if it hasn't been set yet.

signac/contrib/job.py
```diff
@@ -510,35 +509,37 @@ def _init(self, force=False):
...
-                        raise
+            try:
+                # Open the file for writing only if it does not exist yet.
+                with open(fn_manifest, "w" if force else "x") as file:
+                    file.write(blob)
+            except OSError as error:
+                if error.errno not in (errno.EEXIST, errno.EACCES):
+                    raise
             except Exception as error:
                 # Attempt to delete the file on error, to prevent corruption.
```
klywang, line 544: When you added:

https://github.com/glotzerlab/signac/blob/09a6677a963185634d1060b9495f6001bdadf4d7/signac/contrib/job.py#L529-L534

Did that prevent this portion of the code from erasing the manifest? - Just want to make sure I understand these changes properly.

bdice: Yes. It will raise the `JobsCorruptedError` and will not continue to the later part of the code that attempts to roll back (delete) the manifest file creation.

klywang: Thank you for the clarification!

klywang (approved):
Looks good!

signac/contrib/job.py
tcmoore3: 👍

tcmoore3:
@bdice I see. This looks good.

tcmoore3 (approved):

PR #470, opened 2021-01-11
Feature/synced collection/contexts

### LINKED ISSUE #465: Use custom context managers rather than contextlib
Noted by @bdice in [this comment](https://github.com/glotzerlab/signac/pull/463/files#r552232801) #463, contextlib's contextmanagers have a lot of overhead. We should benchmark these in SyncedCollections since we use many of them and if appropriate, write our own simpler context managers to improve performance.

vyasr:
## Description

This PR converts all context managers in synced collections from `contextlib.contextmanager` to custom classes.

## Motivation and Context

Resolves #465. Using `contextlib.contextmanager` can have considerable overhead, so it's worthwhile to avoid. I should note that most of my benchmarks were not sensitive enough to capture such performance losses, since in general the contents of context managers outweigh the overhead of the context itself. Moreover, a number of different context managers throughout the code base follow a simple counting based paradigm that is now encoded in a pair of standardized classes in the utils module.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/core/synced_collections/buffered_collection.py
+++ b/signac/core/synced_collections/buffered_collection.py
@@ -134,13 +134,13 @@ class BufferedCollection(SyncedCollection):
         or instead written to a temporary buffer based on the buffering mode.
         """
         if not self._suspend_sync:
-            if self._parent is None:
+            if self._root is None:
                 if self._is_buffered:
                     self._save_to_buffer()
                 else:
                     self._save_to_resource()
             else:
-                self._parent._save()
+                self._root._save()
 
     def _load(self):
         """Load data from the backend but buffer if needed.
@@ -150,7 +150,7 @@ class BufferedCollection(SyncedCollection):
         or instead read from a temporary buffer based on the buffering mode.
         """
         if not self._suspend_sync:
-            if self._parent is None:
+            if self._root is None:
                 if self._is_buffered:
                     data = self._load_from_buffer()
                 else:
@@ -158,7 +158,7 @@ class BufferedCollection(SyncedCollection):
                 with self._suspend_sync:
                     self._update(data)
             else:
-                self._parent._load()
+                self._root._load()
 
     def _save_to_buffer(self):
         """Store data in buffer.
--- a/signac/core/synced_collections/buffered_collection.py
+++ b/signac/core/synced_collections/buffered_collection.py
@@ -187,7 +187,7 @@ class BufferedCollection(SyncedCollection):
     @property
     def _is_buffered(self):
         """Check if we should write to the buffer or not."""
-        return self.buffered or type(self).buffer_all
+        return self.buffered or _BUFFER_ALL_CONTEXT
 
     def _flush(self):
         """Flush data associated with this instance from the buffer."""
@@ -199,27 +199,16 @@ class BufferedCollection(SyncedCollection):
         pass
 
 
-# Have to monkey-patch this into the class after the fact because the
-# flush_all_backends method is not fully defined and bound as a static method
-# until after the class definition is complete.
-BufferedCollection.buffer_all = _CounterFuncContext(  # type: ignore
-    BufferedCollection._flush_all_backends
-)
-"""Enter a globally buffer context for all BufferedCollection instances.
-
-All future operations use the buffer whenever possible. Write operations
-are deferred until the context is exited, at which point all buffered
-backends will flush their buffers. Individual backends may flush their
-buffers within this context if the implementation requires it; this context
-manager represents a promise to buffer whenever possible, but does not
-guarantee that no writes will occur under all circumstances.
-"""
+# This module-scope variable is a context that can be accessed via the
+# buffer_all method for the purpose of buffering all subsequence read and write
+# operations.
+_BUFFER_ALL_CONTEXT = _CounterFuncContext(BufferedCollection._flush_all_backends)
 
 
 # This function provides a more familiar module-scope, function-based interface
 # for enabling buffering rather than calling the class's static method.
 def buffer_all():
-    """Enter a globally buffer context for all BufferedCollection instances.
+    """Return a global buffer context for all BufferedCollection instances.
 
     All future operations use the buffer whenever possible. Write operations
     are deferred until the context is exited, at which point all buffered
@@ -227,7 +216,5 @@ def buffer_all():
     buffers within this context if the implementation requires it; this context
     manager represents a promise to buffer whenever possible, but does not
     guarantee that no writes will occur under all circumstances.
-
-    This method is a trivial alias for :meth:`~.BufferedCollection.buffer_all`
     """
-    return BufferedCollection.buffer_all
+    return _BUFFER_ALL_CONTEXT
--- a/signac/core/synced_collections/buffered_collection.py
+++ b/signac/core/synced_collections/buffered_collection.py
@@ -178,7 +178,7 @@ class BufferedCollection(SyncedCollection):
                     data = self._load_from_buffer()
                 else:
                     data = self._load_from_resource()
-                with self._suspend_sync():
+                with self._suspend_sync:
                     self._update(data)
             else:
                 self._parent._load()
--- a/signac/core/synced_collections/buffered_collection.py
+++ b/signac/core/synced_collections/buffered_collection.py
@@ -36,7 +36,6 @@ buffer flushes will occur when all such managers have been exited.
 """
 
 import logging
-from contextlib import contextmanager
 from inspect import isabstract
 from typing import Any, List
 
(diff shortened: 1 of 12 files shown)


signac/core/synced_collections/buffered_collection.py
```diff
@@ -233,4 +199,35 @@ def _flush_buffer(self):
...
+    All future operations use the buffer whenever possible. Write operations
+    are deferred until the context is exited, at which point all buffered
+    backends will flush their buffers. Individual backends may flush their
+    buffers within this context if the implementation requires it; this context
+    manager represents a promise to buffer whenever possible, but does not
+    guarantee that no writes will occur under all circumstances.
+
+    This method is a trivial alias for :meth:`~.BufferedCollection.buffer_all`
+    """
+    return BufferedCollection.buffer_all
```
bdice, line 233: Do we need the monkey patch? Would this function be enough? (It might need some kind of singleton-like pattern?)

signac/core/synced_collections/synced_collection.py
```diff
@@ -128,10 +128,19 @@ class SyncedCollection(Collection):
     _validators: List[Callable] = []
     # Backends that support threading should modify this flag.
     _supports_threading: bool = False
+    _LoadSaveType = _LoadAndSave
 
     def __init__(self, parent=None, *args, **kwargs):
-        self._parent = parent
-        self._suspend_sync_ = 0
+        if parent is not None:
+            self._parent = parent._parent if parent._parent is not None else parent
```
bdice, line 135: This might be silly, but it took me a little time to understand. This line makes it _seem_ as if you're accessing the "grandparent" whereas you're really accessing the top level ancestor (by construction, since the parent's `_parent` actually refers to the top level ancestor by this same logic). Would it make sense to change the name of this attribute to something like `_ancestor` or `_root` or `_root_synced_collection`, or at least add a comment to explain this fact? I realize that might require changes in several places, but just wanted to raise that for consideration.

signac/core/synced_collections/synced_collection.py
```diff
@@ -128,10 +128,19 @@ class SyncedCollection(Collection):
...
+        else:
+            self._parent = parent
```
bdice, line 137: In addition to the suggestion above, I believe this logic could be shorter (but possibly not clearer -- you can make the call of whether to change this).
```suggestion
        self._parent = getattr(parent, "_parent", None) or parent
```

signac/core/synced_collections/synced_collection.py
bdice: Thinking some more about this, the name `_parent` seems semantically inaccurate since it doesn't refer to "one level up." The name `_root` makes more sense to me.

signac/core/synced_collections/synced_collection.py
```diff
@@ -128,10 +128,19 @@ class SyncedCollection(Collection):
...
+        self._suspend_sync = (
+            _CounterContext() if parent is None else parent._suspend_sync
+        )
+        self._load_and_save = (
+            type(self)._LoadSaveType(self) if parent is None else parent._load_and_save
+        )
```
bdice, line 143: It might make more sense to get these attributes from `self._root` instead of the passed `parent` (depending on the suggested name change).

signac/core/synced_collections/buffered_collection.py
vyasr: I'm not sure I follow your suggestion. If you're just asking whether we can just skip putting `buffer_all` into the namespace of the `BufferedCollection`, we certainly can. From a conceptual standpoint, to me it actually makes more sense in the `BufferedCollection` than as a module-scope function, and I only included the function as a nod to backwards compatibility. If you think that the monkey-patch is confusing, though, then that's fine; we'd just need to preserve the context as a global variable at the module scope and return it via the function.

signac/core/synced_collections/synced_collection.py
vyasr: Yes, this is a recursive construction to flatten the parent hierarchy that we currently store. There's really no reason to know the parent, you only ever need to know the top level, so you might as well store the root. I agree that changing the name makes sense.

signac/core/synced_collections/synced_list.py
```diff
@@ -185,8 +185,8 @@ def reset(self, data=None):
 
     def __setitem__(self, key, value):
         self._validate(value)
-        with self._load_and_save():
-            with self._suspend_sync():
+        with self._load_and_save:
+            with self._suspend_sync:
```
bdice, line 189: I believe these can be written more concisely as:
```suggestion
        with self._load_and_save, self._suspend_sync:
```

signac/core/synced_collections/utils.py
```diff
@@ -141,3 +141,64 @@ def default(self, o: Any) -> Dict[str, Any]:  # noqa: D102
...
             return super().default(o)
+
+
+class _NullContext:
+    """A nullary context manager.
+
+    There are various cases where we sometimes want to perform a task within a
+    particular context, but at other times we wish to ignore that context. The
+    most obvious example is a lock for threading: since
+    :class:`SyncedCollection`s allow multithreading support to be enabled or
```
bdice, line 152: To pluralize a class reference in rST, you either need a raw string with an escaped space like:
```python
r"""This is how to pluralize :class:`SyncedCollection`\ s."""
```
or (my preference) you can just treat the class as a collective/singular noun.
```suggestion
    :class:`SyncedCollection` allows multithreading support to be enabled or
```

signac/core/synced_collections/utils.py
```diff
@@ -141,3 +141,64 @@ def default(self, o: Any) -> Dict[str, Any]:  # noqa: D102
...
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        pass
+
+
+class _CounterContext:
+    """A context manager that maintains a total entry count.
+
+    This class simply contains an internal counter that is incremented on every
+    entrance and decremented on every exit. It is also truthy and only evaluates
+    to False if the count is 0.
```
bdice, line 174: This is clearer to me:
```suggestion
    to True if the count is greater than 0.
```

signac/core/synced_collections/utils.py
```diff
@@ -141,3 +141,64 @@ def default(self, o: Any) -> Dict[str, Any]:  # noqa: D102
...
+    callable to be executed anytime the context exits and the counter hits zero.
+    """
+
+    def __init__(self, func):
+        super().__init__()
+        self._func = func
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        super().__exit__(exc_type, exc_val, exc_tb)
+        if self._count == 0:
```
bdice, line 203: Perhaps rely on the truthiness?
```suggestion
        if not self:
```

signac/core/synced_collections/buffered_collection.py
bdice: Yes, I think I would prefer only one or the other. Having both the class method and module-level function seems unnecessary. Given that, I would prefer to skip the monkey-patch to put this in the `BufferedCollection` namespace and leave only the module-level function. I don't have strong opinions on this but wanted to at least pose the suggestion because monkey-patches are pretty hacky.

signac/core/synced_collections/synced_list.py
bdice: Supported by Python > 3.1: https://docs.python.org/3/reference/compound_stmts.html#the-with-statement
> With more than one item, the context managers are processed as if multiple with statements were nested:
>
> ```python
> with A() as a, B() as b:
>     SUITE
> ```
> is semantically equivalent to:
>
> ```python
> with A() as a:
>     with B() as b:
>         SUITE
> ```

bdice:
@vyasr Also can you profile / benchmark this PR compared to `contextmanager`?

vyasr:
@bdice I ran some longer benchmarks than my originals to try to capture the difference. It works out to a 2-6% performance increase for common operations (setitem, keys, getitem, values, call).

signac/core/synced_collections/buffered_collection.py
vyasr: That's fine, I'll stick with a global context. I agree that monkey-patches can be hacky and would make things harder to maintain.

signac/core/synced_collections/synced_collection.py
vyasr: While shorter, I believe this is harder to follow and am skipping this change.

signac/core/synced_collections/synced_list.py
vyasr: Good catch, I'm aware of this construct but have never actually seen it in production use anywhere so I always forget that it's an option.

PR #471, opened 2021-01-18
Fix/issue 281

### LINKED ISSUE #281: H5Store pytest is failing on MacOS but passes on Linux

vyasr:
## Description

The test modified by this PR was reading from an hdf5 file in one process while writing to it in a separate process using nested contexts. Even a single concurrent read is not safe while writing in other processes, but I suspect that parallel-safe reads for a single reader might be an implementation detail in some versions of hdf5 but not others (leading to the intermittently successful behavior on OS X and the consistently correct behavior on Linux). Given that a number of the other tests explicitly test SWMR, however, I'm not sure why this test doesn't have it enabled. Perhaps @bdice or @klywang can shed some light on that, I believe they wrote/inspected/modified various versions of these tests.

## Motivation and Context

Resolves #281. In addition, the test will now pass on Windows.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/tests/test_h5store.py
+++ b/tests/test_h5store.py
@@ -671,6 +671,11 @@ class TestH5StoreMultiThreading(TestH5StoreBase):
         assert self.get_h5store()["x"] in set(range(100))
 
 
+def _read_from_h5store(filename, **kwargs):
+    with H5Store(filename, **kwargs) as h5s:
+        list(h5s)
+
+
 class TestH5StoreMultiProcessing(TestH5StoreBase):
     def test_single_writer_multiple_reader_same_process(self):
         with self.open_h5store() as writer:
@@ -688,12 +693,12 @@ class TestH5StoreMultiProcessing(TestH5StoreBase):
     def test_single_writer_multiple_reader_same_instance(self):
         from multiprocessing import Process
 
-        def _read_from_h5store(filename, **kwargs):
-            with H5Store(self._fn_store, mode="r", swmr=True) as h5s:
-                list(h5s)
-
         def read():
-            p = Process()
+            p = Process(
+                target=_read_from_h5store,
+                args=(self._fn_store,),
+                kwargs=(dict(mode="r", swmr=True, libver="latest")),
+            )
             p.start()
             p.join()
             # Ensure the process succeeded
--- a/tests/test_h5store.py
+++ b/tests/test_h5store.py
@@ -671,13 +671,6 @@ class TestH5StoreMultiThreading(TestH5StoreBase):
         assert self.get_h5store()["x"] in set(range(100))
 
 
-def _read_from_h5store(filename, **kwargs):
-    from signac.core.h5store import H5Store
-
-    with H5Store(filename, **kwargs) as h5s:
-        list(h5s)
-
-
 class TestH5StoreMultiProcessing(TestH5StoreBase):
     def test_single_writer_multiple_reader_same_process(self):
         with self.open_h5store() as writer:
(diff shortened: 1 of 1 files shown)


vyasr:
@klywang since you weren't able to reproduce the problem anymore I'm not sure if you can validate these changes, but if you have any way to do so that would be super helpful!

tests/test_h5store.py
```diff
@@ -690,23 +683,24 @@ def test_single_writer_multiple_reader_same_process(self):
...
+            with H5Store(self._fn_store, mode="r", swmr=True) as h5s:
+                list(h5s)
+
         def read():
-            p = Process(
-                target=_read_from_h5store,
-                args=(self._fn_store,),
-                kwargs=(dict(mode="r")),
-            )
+            p = Process()
```
bdice, line 696: I'm confused. Does this process _do_ anything? It seems like it just gets created and joined.

bdice:
@vyasr (tagging @glotzerlab/signac-committers for visibility and to solicit any contradicting opinions)

I believe I wrote these tests of SWMR behavior, but I've never felt confident about them and we've had a great deal of problems with them over time. Given how much effort we've expended on this sub-feature of HDF5 SWMR support (which to the best of my knowledge is not utilized at all by signac users), I'm not sure that it is in still in scope for us as signac maintainers to test extensive guarantees about SWMR behavior in signac. It seems so highly dependent on parts of the software stack that we don't control (specifics of the HDF5 build, h5py, Python implementation, and OS level support?!). Thus:

1. I am of the opinion that power users with extremely specific needs for SWMR are likely experienced enough to learn and correctly use h5py directly, rather than the simple wrapper that signac provides.
2. I am in favor of _literally any solution that will eliminate the problems SWMR has caused us in testing_, whether that means fixing tests so that they reliably pass, marking them as expected failures, or removing the tests entirely and advocating that users use h5py directly if they have such advanced needs.

tests/test_h5store.py
vyasr: Well that's strange. I'm not sure how it got pushed in this weird intermediate state, my fault for context switching between multiple branches. I've pushed the actual fix now.

vyasr:
> @vyasr (tagging @glotzerlab/signac-committers for visibility and to solicit any contradicting opinions)
> 
> I believe I wrote these tests of SWMR behavior, but I've never felt confident about them and we've had a great deal of problems with them over time. Given how much effort we've expended on this sub-feature of HDF5 SWMR support (which to the best of my knowledge is not utilized at all by signac users), I'm not sure that it is in still in scope for us as signac maintainers to test extensive guarantees about SWMR behavior in signac. It seems so highly dependent on parts of the software stack that we don't control (specifics of the HDF5 build, h5py, Python implementation, and OS level support?!). Thus:
> 
>     1. I am of the opinion that power users with extremely specific needs for SWMR are likely experienced enough to learn and correctly use h5py directly, rather than the simple wrapper that signac provides.
> 
>     2. I am in favor of _literally any solution that will eliminate the problems SWMR has caused us in testing_, whether that means fixing tests so that they reliably pass, marking them as expected failures, or removing the tests entirely and advocating that users use h5py directly if they have such advanced needs.

I agree that SWMR is highly dependent on a lots of software that we don't control, and I'd be fine not worrying about it. We're not actually doing anything to support SWMR; any user who wants SWMR support is expected to pass in the appropriate kwargs to transparently be passed through to h5py. Having said that, while I wasn't very involved with earlier passes through this code so I don't know what caused the problems with tests, at least in this case the issue is an improper usage of h5py in the test itself. I'd say that we move forward with this fix for now since there haven't been any recent hdf5 issues that I'm aware of aside from this one (even this issue is almost a year old at this point). If we run into more problems in the future I'd be fine revisiting the question of just removing these tests since ultimately they're largely testing external code.

bdice (approved):
Looks good to me.

PR #444, opened 2020-12-23
doc/bettter document update and reset statepoint

cbkerr:
## Description

The docstrings for `job.update_statepoint()` and `job.reset_statepoint()` are confusing because they are too short and use the same words in the definition.

Here, I propose an update for them to be able to stand on their own.

I also linked them to the relevant part of signac docs (**no intersphinx because the other link in the file didn't use it.**)

Note: preview the changes live here: https://signac--444.org.readthedocs.build/projects/core/en/444/api.html#the-job-class

## Motivation and Context

See a bigger write-up about the confusion here: glotzerlab/signac-docs#108

In an offline discussion with @vyasr last week, we also realized that the main documentation states that these functions change the job id while the docstrings do not. Here, I also propose we add that notice to the docstrings.

## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change<sup>1</sup>

<sup>1</sup>The change breaks (or has the potential to break) existing functionality.

## Checklist:

- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
- [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [ ] I have updated the API documentation as part of the package doc-strings.
- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
- [ ] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added all related issue and pull request numbers for future reference (if applicable). See example below.

PR #472, opened 2021-01-19
Feature/synced collection/replace jsondict

vyasr:
## Description

This PR integrates the new synced collections into the core of signac. The job and project documents as well as the job statepoint now use the new classes. As part of the change, I also cleaned up parts of the Job class and removed basically all of the internal methods, many of which were quite small or only called once. The new classes are seamless fits for the job and project documents, I was able to perform basically a drop-in replacement. The statepoint is a little bit trickier, because it violates some of the core assumptions of how you'd expect SyncedCollections to typically work. A list (not necessarily exhaustive) of relevant points:

1. Statepoints almost never load from disk because they rely on data provided at initialization whenever possible (possibly from the project cache)
2. There are multiple ways in which a statepoint can be saved to disk, and depending on the context the in-memory representation may or may not be valid (because we intercept the normal saving process and instead call `Job.reset_statepoint`). Moreover, in some of these cases the Job can actually move the underlying file itself, which leaves the statepoint in pseudo-invalid states (not really invalid, since the file of a JSONDict need not exist, but it's something of an antipattern to be intentionally moving the file prior to having in-memory changes reflected).
3. Shallow copies of a job should reflect modifications, in particular the `id` of all copies should be synchronously updated when the statepoint is modified, so the typical approaches to make modifications thread safe go out the window.

These issues were already partly reflected by the necessarily slightly awkward design of the `_sp_save_hook`, in particular the need to store many jobs in a single hook to trigger updates of shallow copies. I'm not 100% happy with the results, in large part because there are 5 or 6 places where the Job is calling internal methods or modifying attributes of the statepoint. I think it's still a significant improvement over the old code, though, and I think integrating it will help us identify possible improvements. I'm very open to suggestions on ways that this can be improved.

One other note: as discussed with multiple other committers, the new JSON collections do not implement a force write mode comparable to the old JSONDict. I've added enough of a backwards compatibility layer that it won't affect the API, but of course the behavior will be no longer be the same. Aside from this very minor change, however, the new classes don't introduce any API breakage, so (looking forward a bit) I would feel comfortable making a minor release or two with the new classes as internal APIs only while we iron out any kinks. This would also give us a little bit of time to finish up the remaining items on the signac 2.0 checklist.

**Edit**

I've made some edits and am now much happier with the design. Much of the logic from the old `reset_statepoint` method has now been moved to the `_save` method of the `_StatepointDict` class, which helps resolve the majority of the concerns. The old `reset_statepoint` method is now a one-liner requesting that the statepoint reset itself. Since the migration logic is now contained within the `_StatepointDict`, there are no longer issues with the `_StatepointDict` being in an invalid state in the middle of `Job` functions. Moreover, thread locks can now be transparently migrated by the dict itself, allowing thread-safe modification of these objects. These changes largely remove the need for most complicated logic in the Job class.

## Motivation and Context

Next step in #249.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/__init__.py
+++ b/signac/__init__.py
@@ -26,10 +26,10 @@ from .contrib import (
 )
 from .contrib import filesystems as fs
 from .contrib import get_job, get_project, index, index_files, init_project
+from .contrib.job import get_buffer_load, get_buffer_size
 from .core.h5store import H5Store, H5StoreManager
 from .core.jsondict import JSONDict
 from .core.jsondict import flush_all as flush
-from .core.jsondict import get_buffer_load, get_buffer_size
 from .core.synced_collections.buffered_collection import buffer_all as buffered
 from .core.synced_collections.buffered_collection import is_buffered
 from .db import get_database
--- a/signac/__init__.py
+++ b/signac/__init__.py
@@ -26,7 +26,7 @@ from .contrib import (
 )
 from .contrib import filesystems as fs
 from .contrib import get_job, get_project, index, index_files, init_project
-from .contrib.job import get_buffer_load, get_buffer_size
+from .contrib.job import get_buffer_load, get_buffer_size, set_buffer_size
 from .core.h5store import H5Store, H5StoreManager
 from .core.jsondict import JSONDict
 from .core.jsondict import flush_all as flush
@@ -69,6 +69,7 @@ __all__ = [
     "flush",
     "get_buffer_size",
     "get_buffer_load",
+    "set_buffer_size",
     "JSONDict",
     "H5Store",
     "H5StoreManager",
--- a/signac/__init__.py
+++ b/signac/__init__.py
@@ -28,10 +28,10 @@ from .contrib import filesystems as fs
 from .contrib import get_job, get_project, index, index_files, init_project
 from .contrib.job import get_buffer_load, get_buffer_size, set_buffer_size
 from .core.h5store import H5Store, H5StoreManager
-from .core.jsondict import JSONDict
 from .core.jsondict import flush_all as flush
 from .core.synced_collections.buffered_collection import buffer_all as buffered
 from .core.synced_collections.buffered_collection import is_buffered
+from .core.synced_collections.collection_json import JSONDict
 from .db import get_database
 from .diff import diff_jobs
 from .version import __version__
--- a/signac/__init__.py
+++ b/signac/__init__.py
@@ -28,10 +28,10 @@ from .contrib import filesystems as fs
 from .contrib import get_job, get_project, index, index_files, init_project
 from .core.h5store import H5Store, H5StoreManager
 from .core.jsondict import JSONDict
-from .core.jsondict import buffer_reads_writes as buffered
 from .core.jsondict import flush_all as flush
 from .core.jsondict import get_buffer_load, get_buffer_size
-from .core.jsondict import in_buffered_mode as is_buffered
+from .core.synced_collections.buffered_collection import buffer_all as buffered
+from .core.synced_collections.buffered_collection import is_buffered
 from .db import get_database
 from .diff import diff_jobs
 from .version import __version__
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -36,7 +36,7 @@ class _StatepointDict(JSONDict):
            Job directory migrations.
     """
 
-    _PROTECTED_KEYS = ("_jobs",)
+    _PROTECTED_KEYS = ("_jobs", "_requires_init")
     # State points are rarely modified and are not designed for efficient
     # modification, so they do not support multithreaded execution.
     # Implementing thread safe modifications would also be quite difficult
@@ -105,7 +105,7 @@ class _StatepointDict(JSONDict):
         if force or not os.path.isfile(self._filename):
             super()._save()
 
-    def load(self):
+    def load(self, job_id):
         """Force a load from disk.
 
         Unlike normal JSONDict objects, this class requires the ability to load
@@ -113,7 +113,22 @@ class _StatepointDict(JSONDict):
         validated against the data on disk; at all other times, the in-memory
         data is assumed to be accurate to avoid unnecessary I/O.
         """
-        super()._load()
+        if not self._suspend_sync:
+            if self._root is None:
+                try:
+                    data = self._load_from_resource()
+                except JSONDecodeError:
+                    raise JobsCorruptedError([job_id])
+
+                if calc_id(data) != job_id:
+                    raise JobsCorruptedError([job_id])
+
+                with self._suspend_sync:
+                    self._update(data)
+
+                return data
+            else:
+                self._root._load()
 
 
 class Job:
@@ -367,9 +382,7 @@ class Job:
         if self._statepoint._requires_init:
             # Load the state point lazily.
             try:
-                self._statepoint.load()
-                statepoint = self._statepoint()
-                assert calc_id(statepoint) == self.id
+                statepoint = self._statepoint.load(self.id)
             except (JSONDecodeError, AssertionError):
                 raise JobsCorruptedError([self.id])
 
@@ -565,9 +578,7 @@ class Job:
         try:
             # Attempt early exit if the state point file exists and is valid.
             try:
-                statepoint = self._statepoint._load_from_resource()
-                if calc_id(statepoint) != self.id:
-                    raise JobsCorruptedError([self.id])
+                statepoint = self._statepoint.load(self.id)
             except Exception:
                 # Any exception means this method cannot exit early.
 
@@ -597,8 +608,7 @@ class Job:
                     raise error
                 else:
                     try:
-                        statepoint = self._statepoint._load_from_resource()
-                        assert calc_id(statepoint) == self.id
+                        statepoint = self._statepoint.load(self.id)
                     except (JSONDecodeError, AssertionError):
                         raise JobsCorruptedError([self.id])
 
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -41,8 +41,7 @@ class _StatepointDict(JSONDict):
     # Implementing thread safe modifications would also be quite difficult
     # because state point modification triggers a migration that moves the
     # file. Moreover, since shallow copies of jobs share state points to
-    # trigger id updates, and since Job._reset_statepoint is called within
-    # _StatepointDict._save, the filename will actually change withiin the
+    # trigger id updates, so the filename will actually change within the
     # context. Since this linkage between the Job and the _StatepointDict
     # allows the _StatepointDict to be in invalid intermediate states during
     # the process, making the threading work would require somewhat complex and
@@ -84,8 +83,47 @@ class _StatepointDict(JSONDict):
     def _save(self):
         # State point modification triggers job migration for all jobs sharing
         # this state point (shallow copies of a single job).
+        new_id = calc_id(self)
+
+        # All elements of the job list are shallow copies of each other, so any
+        # one of them is representative.
+        job = self._jobs[0]
+        if job._id == new_id:
+            return
+
+        tmp_statepoint_file = self.filename + "~"
+        should_init = False
+        try:
+            os.replace(self.filename, tmp_statepoint_file)
+            new_workspace = os.path.join(job._project.workspace(), new_id)
+            try:
+                os.replace(job.workspace(), new_workspace)
+            except OSError as error:
+                os.replace(tmp_statepoint_file, self.filename)  # rollback
+                if error.errno in (errno.EEXIST, errno.ENOTEMPTY, errno.EACCES):
+                    raise DestinationExistsError(new_id)
+                else:
+                    raise
+            else:
+                should_init = True
+        except OSError as error:
+            if error.errno != errno.ENOENT:  # OK if file is not initialized.
+                raise
+
+        # Update each job instance.
         for job in self._jobs:
-            job._reset_statepoint(self)
+            old_id = job._id
+            job._id = new_id
+            job._wd = None
+            job._document = None
+            job._stores = None
+            job._cwd = []
+            self._filename = job._statepoint_filename
+        if should_init:
+            # All the jobs are equivalent, so just init the last one from the
+            # for loop.
+            job.init()
+        logger.info(f"Moved '{old_id}' -> '{new_id}'.")
 
     def save(self, force=False):
         """Trigger a save to disk.
@@ -160,23 +198,6 @@ class _StatepointDict(JSONDict):
 
         return data
 
-    @property
-    def filename(self):
-        """str: Get or set the name of the file this collection is synchronized with.
-
-        This property is overridden in order to support setting. In the parent
-        class the semantics for setting a filename are unclear; since a JSONDict
-        is always synced to a file, setting a filename might be expected to move
-        the file if it exists. In this subclass, we can safely define exactly the
-        behavior required, which is that the filename changes but the file is not
-        moved.
-        """
-        return self._filename
-
-    @filename.setter
-    def filename(self, new_filename):
-        self._filename = new_filename
-
 
 class Job:
     """The job instance is a handle to the data of a unique state point.
@@ -224,7 +245,7 @@ class Job:
             self._statepoint_requires_init = False
             self._statepoint = _StatepointDict(jobs=[self], data=statepoint)
             self._id = calc_id(self._statepoint()) if _id is None else _id
-            self._statepoint.filename = self._statepoint_filename
+            self._statepoint._filename = self._statepoint_filename
 
             # Update the project's state point cache immediately if opened by state point
             self._project._register(self.id, statepoint)
@@ -335,45 +356,6 @@ class Job:
         """
         self._statepoint.reset(new_statepoint)
 
-    def _reset_statepoint(self, new_statepoint):
-        if isinstance(new_statepoint, JSONDict):
-            new_statepoint = new_statepoint()
-
-        new_id = calc_id(new_statepoint)
-        if self._id == new_id:
-            return
-
-        tmp_statepoint_file = self.statepoint.filename + "~"
-        should_init = False
-        try:
-            os.replace(self.statepoint.filename, tmp_statepoint_file)
-            new_workspace = os.path.join(self._project.workspace(), new_id)
-            try:
-                os.replace(self.workspace(), new_workspace)
-            except OSError as error:
-                os.replace(tmp_statepoint_file, self.statepoint.filename)  # rollback
-                if error.errno in (errno.EEXIST, errno.ENOTEMPTY, errno.EACCES):
-                    raise DestinationExistsError(new_id)
-                else:
-                    raise
-            else:
-                should_init = True
-        except OSError as error:
-            if error.errno != errno.ENOENT:  # OK if file is not initialized.
-                raise
-
-        # Update this instance
-        old_id = self._id
-        self._id = new_id
-        self._wd = None
-        self._document = None
-        self._stores = None
-        self._cwd = []
-        self._statepoint.filename = self._statepoint_filename
-        if should_init:
-            self.init()
-        logger.info(f"Moved '{old_id}' -> '{new_id}'.")
-
     def update_statepoint(self, update, overwrite=False):
         """Update the state point of this job.
 
@@ -409,7 +391,7 @@ class Job:
                 if statepoint.get(key, value) != value:
                     raise KeyError(key)
         statepoint.update(update)
-        self.reset_statepoint(statepoint)
+        self._statepoint.reset(statepoint)
 
     @property
     def statepoint(self):
@@ -452,7 +434,7 @@ class Job:
             The new state point to be assigned.
 
         """
-        self.reset_statepoint(new_statepoint)
+        self._statepoint.reset(new_statepoint)
 
     @property
     def sp(self):
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -620,7 +620,8 @@ class Job:
                     raise
 
                 self._statepoint.save(force=force)
-                # Re-load from disk as a validation.
+                # Re-load as a validation (required to detect invalid data on
+                # disk).
                 statepoint = self._statepoint.load(self.id)
 
             # Update the project's state point cache if the saved file is valid.
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -336,10 +336,6 @@ class Job:
         if isinstance(new_statepoint, JSONDict):
             new_statepoint = new_statepoint()
 
-        # This technically isn't 100% equivalent to the old logic, because this
-        # doesn't check workspace equality. However, since the old logic opened
-        # the new job using self._project it wouldn't actually be possible to
-        # have two different projects, so checking the id is sufficient.
         new_id = calc_id(new_statepoint)
         if self._id == new_id:
             return
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -243,9 +243,10 @@ class Job:
             raise ValueError("Either statepoint or _id must be provided.")
         elif statepoint is not None:
             self._statepoint_requires_init = False
-            self._statepoint = _StatepointDict(jobs=[self], data=statepoint)
-            self._id = calc_id(self._statepoint()) if _id is None else _id
-            self._statepoint._filename = self._statepoint_filename
+            self._id = calc_id(statepoint) if _id is None else _id
+            self._statepoint = _StatepointDict(
+                jobs=[self], filename=self._statepoint_filename, data=statepoint
+            )
 
             # Update the project's state point cache immediately if opened by state point
             self._project._register(self.id, statepoint)
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -41,7 +41,7 @@ class _StatepointDict(JSONDict):
     # Implementing thread safe modifications would also be quite difficult
     # because state point modification triggers a migration that moves the
     # file. Moreover, since shallow copies of jobs share state points to
-    # trigger id updates, and since Job.reset_statepoint is called within
+    # trigger id updates, and since Job._reset_statepoint is called within
     # _StatepointDict._save, the filename will actually change withiin the
     # context. Since this linkage between the Job and the _StatepointDict
     # allows the _StatepointDict to be in invalid intermediate states during
@@ -85,7 +85,7 @@ class _StatepointDict(JSONDict):
         # State point modification triggers job migration for all jobs sharing
         # this state point (shallow copies of a single job).
         for job in self._jobs:
-            job.reset_statepoint(self)
+            job._reset_statepoint(self)
 
     def save(self, force=False):
         """Trigger a save to disk.
@@ -176,22 +176,6 @@ class _StatepointDict(JSONDict):
     @filename.setter
     def filename(self, new_filename):
         self._filename = new_filename
-
-    def reset_data(self, data):
-        """Reset the in-memory data without saving.
-
-        When a statepoint is modified prior to the job being initialized, the
-        job's reset_statepoint method needs a way to modify the in-memory data
-        without calling save. This method provides that interface via a transparent
-        call-through to the internal _update method.
-
-        Parameters
-        ----------
-        data : dict
-            The new data.
-
-        """
-        self._update(data)
 
 
 class Job:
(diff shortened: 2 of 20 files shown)


signac/contrib/job.py
```diff
@@ -818,7 +803,9 @@ def __exit__(self, err_type, err_value, tb):
 
     def __setstate__(self, state):
         self.__dict__.update(state)
-        self.statepoint._parent.jobs.append(self)
+        # Note that we append to a list of jobs rather than replacing to
+        # support transparent id updates between shallow copies of a job.
+        self.statepoint._jobs.append(self)
```
bdice, line 808: Perhaps rename to `_job_references` or something that indicates that all of the elements are actually the _same_ job.

signac/contrib/project.py
```diff
@@ -26,7 +26,7 @@
 from ..common.config import Config, get_config, load_config
 from ..core import json
```
bdice, line 27: We'll want to carefully review any remaining references to `core/json.py`.

tests/test_buffered_mode.py
```diff
@@ -165,6 +172,7 @@ def test_buffered_mode_change_buffer_size(self):
                 with signac.buffered(buffer_size=14):
                     pass
 
+    @pytest.mark.xfail(reason="This test uses various deprecated APIs.")
```
bdice, line 175: Most of this test should still be viable. I'd rather see us rip out or alter the deprecated bits and keep the bulk of the code here, instead of `xfail`-ing it.

bdice (commented):
Nice work, @vyasr! I'm excited to try out these changes. I have some preliminary comments here that may be helpful.

signac/contrib/job.py
```diff
@@ -24,31 +24,81 @@
...
-    When a job's state point is changed, in addition
-    to the contents of the file being modified this hook
-    calls :meth:`~Job._reset_sp` to rehash the state
-    point, compute a new job id, and move the folder.
+    It's also not obvious how to achieve thread-safety for statepoint
+    modifications within the current framework because when multiple copies of
+    a job (shallow copies owning the same statepoint) exist and one of them is
+    modified, the calls to reset_statepoint will invalidate the per-file locks
+    because the folders are moved. Since statepoint accesses do not need to be
+    thread safe, this context manager simply removes that functionality.
```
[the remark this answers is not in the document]
vyasr: I make all these commits to fix drafted docstrings and yet I still miss this... *sigh. Yes, I'll rewrite this.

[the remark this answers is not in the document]
vyasr: Actually I totally forgot that my design for SyncedCollections already accommodates this requirement in a much cleaner way because I made it possible for classes to indicate that they don't support multithreading using a class-leveling boolean. So this class is superfluous (although the comment is still useful and can be moved). Yay for thinking ahead.

signac/contrib/job.py
```diff
@@ -376,7 +395,8 @@ def document(self):
 
             If you need a deep copy that will not modify the underlying
             persistent JSON file, use :attr:`~Job.document` instead of :attr:`~Job.doc`.
-            For more information, see :attr:`~Job.statepoint` or :class:`~signac.JSONDict`.
+            For more information, see
+            :class:`~signac.core.synced_collections.collection_json.BufferedJSONDict`.
```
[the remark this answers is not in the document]
vyasr: This docstring is just wrong, it should say to use the call operator. Both `doc` and `document` have this problem. I was debating whether to fix it in this PR in the interest of not confusing reviewers. It's small enough that I guess I might as well, though.

signac/contrib/job.py
```diff
@@ -398,7 +419,7 @@ def document(self, new_doc):
 
         Parameters
         ----------
-        new_doc : :class:`~signac.JSONDict`
+        new_doc : :class:`~signac.core.synced_collections.collection_json.BufferedJSONDict`
```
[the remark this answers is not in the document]
vyasr: For the moment, yes: [`api.rst`](https://github.com/glotzerlab/signac/blob/feature/synced_collections/doc/api.rst#synced_collections-package) was updated in an earlier PR for synced collections. As to whether that will be true in the final version, I'd say that's a decision we can make when I make the final `feature/synced_collections`-> `master` PR. The only files that should be affected are `job.py` and `project.py`.

signac/contrib/job.py
```diff
@@ -594,7 +539,47 @@ def init(self, force=False):
...
+                    raise error
+                else:
+                    try:
+                        statepoint = self.statepoint._load_from_resource()
+                        assert calc_id(statepoint) == self.id
+                    except (JSONDecodeError, AssertionError):
+                        raise JobsCorruptedError([self.id])
+
+            # Update the project's state point cache if the manifest is valid
+            self._project._register(self.id, self.statepoint())
```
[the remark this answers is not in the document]
vyasr: Good catch, I made sure to not call the statepoint in various other places but missed this one.

signac/contrib/job.py
```diff
@@ -594,7 +539,47 @@ def init(self, force=False):
...
+                        pass
+                    raise error
+                else:
+                    try:
+                        statepoint = self.statepoint._load_from_resource()
+                        assert calc_id(statepoint) == self.id
+                    except (JSONDecodeError, AssertionError):
+                        raise JobsCorruptedError([self.id])
+
+            # Update the project's state point cache if the manifest is valid
```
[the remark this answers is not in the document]
vyasr: Yes I debated doing it at as part of this PR. I didn't originally for the same reason I didn't fix the `Job.document` docstring, not wanting to pollute the PR review, but I'll go ahead and fix this now.

signac/contrib/job.py
vyasr: They aren't the same job object, they are _shallow copies_ of the same job.

Actually, it's worth noting that your question is relevant to a bug I identified while looking at this code (that is also already present on master). Consider this script:
```
import signac

with signac.TemporaryProject() as tmp_project:
    tmp_project.open_job({'a': 1}).init()
    job1 = next(iter(tmp_project))
    job2 = next(iter(tmp_project))
    print(f"The original ids are {job1.id} and {job2.id}")
    job1.sp.a = 0
    print(f"The new ids are {job1.id} and {job2.id}")
```
Running this script *on master* outputs
```
The original ids are 42b7b4f2921788ea14dac5566e6f06d0 and 42b7b4f2921788ea14dac5566e6f06d0
The new ids are 9bfd29df07674bc4aa960cf661b5acd2 and 42b7b4f2921788ea14dac5566e6f06d0
```
Fixing this link would require some much more sophisticated linkages between all jobs, and even that would only be process-local (I can't see any way to fix this between processes without I/O). I also doubt it's a significant problem in practical uses. Just noting that it is one limitation of assuming that in-memory representations are ever valid in our data model.

signac/contrib/project.py
vyasr: Yes. Half of them are in deprecated modules, the other half (as far as I can tell) don't actually need the one bit of extra functionality provided by `signac/core/json.py` (making it possible to dump `_SyncedDict` objects). We should be able to deprecate `json.py`. I would like to do that in a follow-up PR that also deprecates the old `_SyncedDict` family of classes.

tests/test_buffered_mode.py
vyasr: While you're right that the test would still be viable if we modified it correctly, it would basically all be redundant. If you strip out the logic associated with the various deprecated buffer-related functions, then all of the behaviors are tested elsewhere. The only exception I see is the parts testing the amount of data stored in the buffer, and even those are only "new" in the sense that they'd be testing signac APIs that we would need to patch for the new classes, which already test the correct underlying behavior (i.e. there are synced collection tests verifying that buffers are flushed etc, so this test would just make sure that `signac.get_buffer_load() == JSONDict.get_buffer_size()` or so).

signac/contrib/job.py
```diff
@@ -88,30 +138,26 @@ class Job:
...
-                self._id = calc_id(statepoint)
-                # Update the project's state point cache immediately if opened by state point
-                self._project._register(self.id, statepoint)
-            else:
-                self._id = _id
+            self._statepoint = _StatepointDict(jobs=[self], data=statepoint)
+            try:
+                self._id = calc_id(self._statepoint._to_base()) if _id is None else _id
+            except TypeError:
+                raise KeyTypeError
```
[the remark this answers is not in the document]
vyasr: I'm going to resolve all of these requests in favor of #473. I don't want to try to add those tests as part of this pull request, we should add them on master (and fix any associated issues) then merge that down to this feature branch.

signac/contrib/job.py
```diff
@@ -24,31 +23,173 @@
...
+class _StatepointDict(JSONDict):
+    """A JSON-backed dictionary for storing job statepoints.
 
+    There are three principal reasons for extending the base JSONDict:
+        1. Saving needs to trigger a job directory migration, and
+        2. Statepoints are assumed to not support external modification, so
+           they never need to load from disk _except_ the very first time a job
+           is opened by id and they're not present in the cache.
+        3. It must be possible to load and/or save on demand during tasks like
+           Job directory migrations.
```
bdice, line 35: I actually kind of like capitalizing Job to indicate the class, but it's inconsistent with the rest of our docs.
```suggestion
           job directory migrations.
```

signac/contrib/job.py
```diff
@@ -24,31 +23,173 @@
...
+                raise
+
+        # Update each job instance.
+        for job in self._jobs:
+            old_id = job._id
+            job._id = new_id
+            job._wd = None
+            job._document = None
+            job._stores = None
+            job._cwd = []
```
bdice, line 113: I think we should make this "clearing out" an internal method of the `Job` and also call it in `Job.__init__`. Something like `_prepare`. (Otherwise we run the risk of being unsafe if new properties / etc. are added.)

signac/contrib/job.py
```diff
@@ -24,31 +23,173 @@
...
+        except OSError as error:
+            # The most likely reason we got here is because the state point
+            # file move failed due to the job not being initialized so the file
+            # doesn't exist, which is OK.
+            if error.errno != errno.ENOENT:
+                raise
+
+        # Update each job instance.
+        for job in self._jobs:
+            old_id = job._id
```
bdice, line 108: `old_id` should just be fetched once, above. (in the same place that it exits early `if job._id == new_id`)

signac/contrib/job.py
```diff
@@ -24,31 +23,173 @@
...
+            Job id used to validate contents on disk.
+
+        Returns
+        -------
+        data : dict
+            Dictionary of state point data.
+
+        Raises
+        ------
+        JobsCorruptedError
```
bdice, line 176: I fixed this error reference in #444 for all existing cases.
```suggestion
        :class:`~signac.errors.JobsCorruptedError`
```

signac/contrib/job.py
```diff
@@ -73,9 +214,9 @@ class Job:
     """
 
     FN_MANIFEST = "signac_statepoint.json"
```
bdice, line 216: Can we replace "manifest" with "state point" universally? (Depends on whether we consider this constant to be internal API, which it probably should be.) Maybe a topic for another PR / issue...?
```suggestion
    _FN_STATE_POINT = "signac_statepoint.json"
```

signac/contrib/job.py
```diff
@@ -88,29 +229,27 @@ class Job:
     def __init__(self, project, statepoint=None, _id=None):
         self._project = project
 
+        # Prepare wd in advance so that the attribute exists in checks below.
+        self._wd = None
```
bdice, line 233: See above comment about moving all this logic for initializing `_wd`, `_document`, etc. into a `_prepare` method.

signac/contrib/job.py
```diff
@@ -88,29 +229,27 @@ class Job:
...
-                # Compute the id from the state point if not provided.
-                self._id = calc_id(statepoint)
-                # Update the project's state point cache immediately if opened by state point
-                self._project._register(self.id, statepoint)
-            else:
-                self._id = _id
+            self._statepoint_requires_init = False
+            self._id = calc_id(statepoint) if _id is None else _id
+            self._statepoint = _StatepointDict(
+                jobs=[self], filename=self._statepoint_filename, data=statepoint
```
bdice, line 241: How does shallow-copying a Job trigger an append to the _jobs property of the _StatepointDict? (I started discussing with @vyasr on Slack but we're deferring this conversation for now until one of us can investigate.)

signac/contrib/job.py
```diff
@@ -336,12 +409,12 @@ def statepoint(self):
             Returns the job's state point.
 
         """
-        if self._statepoint is None:
-            # Load state point manifest lazily and assign to
-            # self._statepoint
-            statepoint = self._check_manifest()
+        if self._statepoint_requires_init:
+            statepoint = self._statepoint.load(self.id)
```
bdice, line 413: I think this comment is important for developers / readers. It was removed in the diff.
```suggestion
            # Load state point data lazily (on access).
            statepoint = self._statepoint.load(self.id)
```

signac/contrib/job.py
```diff
@@ -409,9 +486,18 @@ def doc(self):
 
         .. warning::
 
+            Even deep copies of :attr:`~Job.doc` will modify the same file, so
+            changes will still effectively be persisted between deep copies.
             If you need a deep copy that will not modify the underlying
-            persistent JSON file, use :attr:`~Job.document` instead of :attr:`~Job.doc`.
-            For more information, see :attr:`~Job.statepoint` or :class:`~signac.JSONDict`.
+            persistent JSON file, use the call operator to get an (otherwise
+            equivalent) raw dictionary: ``job.doc()``.
```
bdice, line 493:
```suggestion
            persistent JSON file, use the call operator to get an equivalent
            plain dictionary: ``job.doc()``.
```

signac/contrib/job.py
```diff
@@ -373,9 +446,13 @@ def document(self):
 
         .. warning::
 
+            Even deep copies of :attr:`~Job.document` will modify the same file,
+            so changes will still effectively be persisted between deep copies.
             If you need a deep copy that will not modify the underlying
-            persistent JSON file, use :attr:`~Job.document` instead of :attr:`~Job.doc`.
-            For more information, see :attr:`~Job.statepoint` or :class:`~signac.JSONDict`.
+            persistent JSON file, use the call operator to get an (otherwise
+            equivalent) raw dictionary: ``job.document()``.
```
bdice, line 453:
```suggestion
            persistent JSON file, use the call operator to get an equivalent
            plain dictionary: ``job.document()``.
```

signac/contrib/job.py
```diff
@@ -493,96 +579,11 @@ def data(self, new_data):
...
-            If the manifest hash is not equal to the job id.
-
-        """
-        manifest = self._read_manifest()
-        if calc_id(manifest) != self.id:
-            raise JobsCorruptedError([self.id])
-        if self._statepoint is None:
-            self._statepoint = SyncedAttrDict(manifest, parent=_sp_save_hook(self))
-        return manifest
-
```
bdice, line 580: _nice work_ 🌞

signac/contrib/job.py
```diff
@@ -24,31 +23,173 @@
...
+                errno.EEXIST,
+                errno.EACCES,
+            ):
+                # Attempt to delete the file on error, to prevent corruption.
+                # OSErrors that are EEXIST or EACCES don't need to delete the file.
+                try:
+                    os.remove(self._filename)
+                except Exception:  # ignore all errors here
+                    pass
+                raise error
```
bdice, line 154: I can't remember how the tracebacks look for `except Exception as error: raise error` versus a bare `raise` that re-raises. I _think_ we want to prefer the bare `raise` (so that the original source of the error is what shows up first in the traceback).

signac/contrib/project.py
```diff
@@ -26,7 +26,7 @@
...
 from ..core.h5store import H5StoreManager
-from ..core.jsondict import JSONDict
+from ..core.synced_collections.collection_json import BufferedJSONDict
```
bdice, line 29: I think I'd like to lay the groundwork for re-structuring signac 2.0 and/or separating out the `synced_collections` package by moving `synced_collections` to the top level outside of `core`.

signac/core/synced_collections/collection_json.py
```diff
@@ -111,6 +112,8 @@ def _load_from_resource(self):
         except OSError as error:
             if error.errno == errno.ENOENT:
                 return None
+            else:
+                raise error
```
bdice, line 116: See above discussion of re-raising. Not 100% sure if we want to make this change or not.
```suggestion
                raise
```

signac/core/synced_collections/synced_list.py
```diff
@@ -150,30 +150,24 @@ def _update(self, data=None):
...
 
-    def reset(self, data=None):
+    def reset(self, data):
         """Update the instance with new data.
 
         Parameters
         ----------
-        data: non-string Sequence, optional
-            Data to update the instance (Default value = None).
+        data: non-string Sequence
```
bdice, line 158:
```suggestion
        data : non-string Sequence
```

tests/test_buffered_mode.py
bdice: But the purpose of an integration test is that you're testing all of it together. It doesn't matter that it's redundant, that's the point. 😆

bdice (changes requested):
Nice, this PR is getting better with every iteration. I have a handful of comments. I'd like another round of review before approving.

signac/contrib/job.py
vyasr: Here's an example. Using `raise error` adds an extra level to the traceback, otherwise mostly identical. In both cases the final line in the traceback is the original source. I'm fine with whichever, @bdice let me know what you think:

```
>>> def f():
...     raise ValueError("In f")
...
>>> def g():
...     try:
...             f()
...     except ValueError as error:
...             raise
...
>>> def h():
...     try:
...             f()
...     except ValueError as error:
...             raise error
...
>>> g()
  File "<stdin>", line 3, in g
  File "<stdin>", line 2, in f
ValueError: In f
>>> h()
  File "<stdin>", line 5, in h
  File "<stdin>", line 3, in h
  File "<stdin>", line 2, in f
ValueError: In f
```

signac/contrib/job.py
vyasr: Yes, I'd consider all the class-level constants to currently be part of the API (and they even have docstrings). Let's discuss the possibility of changing this in a separate issue.

signac/contrib/job.py
vyasr: After a copy is made `__setstate__` will get called, which is where the list is appended to.

signac/contrib/job.py
vyasr: 👍

tests/test_buffered_mode.py
vyasr: The thing you'd be integrating here is the test of buffering with different force modes. If `force_write` is no longer supported, there's nothing in this integration test that isn't tested (even in an integrated context) in other tests. Unless you think there's value in just checking this long sequence of changes?

signac/contrib/job.py
bdice: I don't think that's a bug as long as `job1` and `job2` maintain consistency with their respective id / statepoint.

signac/contrib/project.py
bdice: That's fine, we can resolve this if you don't think there are any action items yet.

signac/contrib/job.py
vyasr: I think it is, at best, unexpected behavior, but I also don't think there's any way to fix this with our current data model without unacceptable amounts of additional file I/O (you'd have to reload from disk every time when checking the id). I'm not proposing any actions here, just pointing out a limitation related to this discussion.

tests/test_buffered_mode.py
bdice: > Unless you think there's value in just checking this long sequence of changes?

That's all I'm going for. This test was added specifically to ensure that long sequences of changes work fine. I would just remove the `force_write` bits (probably just replacing them with standard non-forced calls, even if redundant).

signac/contrib/job.py
bdice: Yes, please use a bare `raise`.
```suggestion
                raise
```

signac/core/synced_collections/collection_json.py
bdice: Yes, let's just `raise`.

signac/contrib/job.py
```diff
@@ -825,7 +843,9 @@ def __exit__(self, err_type, err_value, tb):
 
     def __setstate__(self, state):
         self.__dict__.update(state)
-        self.statepoint._parent.jobs.append(self)
+        # We append to a list of jobs rather than replacing to support
+        # transparent id updates between shallow copies of a job.
+        self.statepoint._jobs.append(self)
```
bdice, line 848: If using a WeakSet as described above:
```suggestion
        self.statepoint._jobs.add(self)
```

signac/core/synced_collections/synced_collection.py
```diff
@@ -340,11 +340,15 @@ def is_base_type(cls, data):
     def _load_from_resource(self):
         """Load data from underlying backend.
 
-        This method must be implemented for each backend.
+        This method must be implemented for each backend. Backends may choose
+        to return ``None``, signaling that no modification should be performed
+        on the data in memory. This mode is useful for backends where the underlying
+        resource (e.g. a file) may not initially exist, but can be transparently
+        created on save.
```
bdice, line 347: Beautiful! 👍

bdice (changes requested):
Additional comments attached. This is probably my last round of feedback before approving.

signac/contrib/job.py
```diff
@@ -24,31 +23,170 @@
...
-    When a job's state point is changed, in addition
-    to the contents of the file being modified this hook
-    calls :meth:`~Job._reset_sp` to rehash the state
-    point, compute a new job id, and move the folder.
-
-    Parameters
-    ----------
-    jobs : iterable of `Jobs`
-        List of jobs(instance of `Job`).
+class _StatepointDict(JSONDict):
```
bdice, line 26: Sooooooo. Statepoint vs. state point. Probably needs to be:
```suggestion
class _StatePointDict(JSONDict):
```

signac/contrib/job.py
```diff
@@ -24,31 +23,170 @@
...
+        data=None,
+        parent=None,
+        *args,
+        **kwargs,
+    ):
+        # A job-statepoint mapping need not be unique because multiple Python
+        # Job objects can point to the same data on disk. We need to store
+        # these jobs in a shared list here so that shallow copies can point to
+        # the same place and trigger each other to update.
+        self._jobs = list(jobs)
```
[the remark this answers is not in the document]
vyasr: Nice idea, I hadn't considered that possibility.

signac/contrib/job.py
```diff
@@ -24,31 +23,170 @@
...
+        # Update each job instance.
+        for job in self._jobs:
+            job._id = new_id
+            job._initialize_lazy_properties()
+
+        # Since all the jobs are equivalent, just grab the filename from the
+        # last one and init it. Also migrate the lock for multithreaded support.
+        old_lock_id = self._lock_id
+        self._filename = job._statepoint_filename
+        type(self)._locks[self._lock_id] = type(self)._locks.pop(old_lock_id)
```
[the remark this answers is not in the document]
vyasr: Yes, exactly. Per-collection locks need to be implemented by each backend since each backend has a distinct way of identifying a unique resource. For file-based backends like JSON, the unique identifier is the filename. I decided not to document this here since anyone trying to understand (or modify) this code would have to read the docs for the `_lock_id` property anyway.

signac/contrib/job.py
```diff
@@ -24,31 +23,170 @@
...
+        job = self._jobs[0]
+        old_id = job._id
+        if old_id == new_id:
+            return
+
+        tmp_statepoint_file = self.filename + "~"
+        should_init = False
+        try:
+            # Move the statepoint to an intermediate location as a backup.
+            os.replace(self.filename, tmp_statepoint_file)
```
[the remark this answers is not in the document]
vyasr: For posterity, this was an intentional decision by @csadorf:
> The fact that the backup file is never removed could be considered a bug. I remember deliberating on this one and ultimately decided to keep the backup in place in case that users discover that their move was actually a bad idea and want to recover from that situation.

I had left it in place, but I've reconsidered after you asked again. This behavior is undocumented and it's highly unlikely that any user could take advantage of this file existing since they wouldn't know what the file contained. I'd rather remove this for now, and reintroduce additional (documented) data integrity measures if we find the need.

signac/contrib/job.py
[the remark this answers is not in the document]
vyasr: Although this is a good idea in principle, it's unfortunately not so easy in practice because 1) weakrefs are not pickleable by default, and 2) the interaction between weakrefs and the copy module has some gotchas, a simple replacement as suggested here does not lead to the `_jobs` being correctly modified by setstate (but it works fine with a list). We could probably make it work by implementing multiple additional methods, but it would require substantial additional code for no practical benefit, just some nice semantic clarity, so I'm going to leave this as is.

signac/__init__.py
```diff
@@ -26,10 +26,10 @@
 )
 from .contrib import filesystems as fs
 from .contrib import get_job, get_project, index, index_files, init_project
+from .contrib.job import get_buffer_load, get_buffer_size
 from .core.h5store import H5Store, H5StoreManager
 from .core.jsondict import JSONDict
```
bdice, line 31: Should this be importing `JSONDict` from synced collections? Would there be API breakage?

signac/contrib/job.py
```diff
@@ -47,10 +48,13 @@ def __init__(
         *args,
         **kwargs,
     ):
-        # A job-statepoint mapping need not be unique because multiple Python
-        # Job objects can point to the same data on disk. We need to store
-        # these jobs in a shared list here so that shallow copies can point to
-        # the same place and trigger each other to update.
+        # Multiple Python Job objects can share a single `_StatepointDict`
```
bdice, line 51:
```suggestion
        # Multiple Python Job objects can share a single `_StatePointDict`
```

signac/contrib/job.py
```diff
@@ -24,31 +24,189 @@
...
+            if error.errno != errno.ENOENT:
+                raise
+
+        # Update each job instance.
+        for job in self._jobs:
+            job._id = new_id
+            job._initialize_lazy_properties()
+
+        # Remove the temporary statepoint file if it was created. Have to do it
+        # here because we need to get the updated job statepoint filename.
```
bdice, line 117:
```suggestion
        # here because we need to get the updated job state point filename.
```

signac/contrib/job.py
```diff
@@ -24,31 +24,189 @@
...
+        new_id = calc_id(self)
+
+        # All elements of the job list are shallow copies of each other, so any
+        # one of them is representative.
+        job = next(iter(self._jobs))
+        old_id = job._id
+        if old_id == new_id:
+            return
+
+        tmp_statepoint_file = self.filename + "~"
```
bdice, line 88: 🤷‍♂️ 🙊 I really don't know what to do with this. We use `statepoint` in the code a lot. Probably leave this as-is...
```suggestion
        tmp_state_point_file = self.filename + "~"
```

tests/test_buffered_mode.py
```diff
@@ -72,6 +73,9 @@ def test_basic_and_nested(self):
             assert job.doc.a == 2
         assert job.doc.a == 2
 
+    @pytest.mark.xfail(
+        reason="The new SyncedCollection does not implement force_write."
```
bdice, line 77: Can you leave a code comment indicating what to do here in the future? Like `# Remove this test in signac 2.0`? Similar for other xfail tests below.

bdice (approved):
Very minor comments. This is ready to go from my side! Eager to hear feedback from another reviewer if anyone is able, @glotzerlab/signac-committers.

signac/__init__.py
vyasr: Yep good catch.

signac/contrib/job.py
vyasr: Yeah I intentionally chose to ignore variables with "statepoint" in them. We can do that in a separate PR on master at some point if we really want to, but that would also be API breaking (for example `project.open_job(state_point=...)`) so it would require additional discussion. As part of the signac 2.0 prototype Simon and I discussed moving to more generic terminology anyway (something like `attrs` rather than `statepoint`, which is very thermo-based), so this seems like a future discussion.

PR #475, opened 2021-01-22
adding documentation for JobsCursor api

atravitz:
Adding JobsCursor documentation

## Description
Including JobsCursor in the `api.rst` file

## Motivation and Context
The JobsCursor class was not included in our api documentation, so methods like `to_dataframe` were not documented at all.

Note: I'm not sure how to handle redundant docstrings between `Project` and `JobsCursor`. Should we document everything in `JobsCursor` and just have `Project` link to it? If so, how much information should we still include so that things like previews in text editors are still complete?

Also, I plan to make a PR to `signac-docs` to have an example for how to use pandas efficiently with signac, similar to the Sacred example.

## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -61,6 +61,28 @@ The Project
     :show-inheritance:
     :exclude-members: Job
 
+
+The JobsCursor class
+====================
+.. _python-api-jobscursor:
+
+.. currentmodule:: signac.contrib.project
+
+.. rubric:: Attributes
+
+.. autosummary::
+    JobsCursor.export_to
+    JobsCursor.groupby
+    JobsCursor.groupbydoc
+    JobsCursor.to_dataframe
+
+
+.. autoclass:: JobsCursor
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+
 The Job class
 =============
 
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -1042,8 +1042,7 @@ class Project:
             ):
                 print(key, list(group))
 
-        If `key` is None, jobs are grouped by identity (by id), placing one job
-        into each group.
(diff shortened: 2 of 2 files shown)


bdice (approved):
Thanks for the fix, @atravitz!

> I'm not sure how to handle redundant docstrings between Project and JobsCursor.

I just updated the `groupby` and `groupbydoc` docstrings on this PR so that they're a bit more consistent. It might be nice to use "See Also" and a reference. That would be like this example from NumPy, where the method `conj` is defined for the `ndarray` object and is also in the `numpy` root namespace. https://numpy.org/doc/stable/reference/generated/numpy.ndarray.conj.html#numpy.ndarray.conj

If you agree, feel free to make that change (replacing the `Project.groupby` and `Project.groupbydoc` docstrings with short references to the same methods of the `JobsCursor`) and then we can merge this. We may want to check that we use the same approach for `export_to` and `to_dataframe`, too.

If this gets too difficult or is delayed, we can just merge with duplicate docstrings and file an issue for the rest. Either way, I'd like to settle on a solution (even if temporary) and merge in the next 2 days (I'm planning to make the 1.6.0 release late this weekend).

cbkerr (approved):
Regarding how to handle duplication between JobsCursor and Project, I think that since this PR is about exposing existing documentation, I think it's okay to leave duplicated for the release and properly de-duplicate in the future.

I did notice that the docstrings are not an *exact duplicate* for the `to_dataframe` methods, and possibly others.

bdice:
Sounds good, thanks for the review @cbkerr. I'll go ahead and merge this so I can prepare the 1.6.0 release.

Issue #380, opened 2020-08-26
Support NumPy "array-like" objects in synced collections.


bdice:
### Feature description

I would like to be able to convert and store CuPy, Dask, or other array-like objects in the same way we transparently cast NumPy arrays to lists for storage in synced collections.

### Proposed solution

Currently we do a check like

```python
if isinstance(data, np.ndarray):
    data = data.tolist()
```

I _think_ the only code change would be to perform a different check for array-like objects and add a second conversion:

```python
if hasattr(data, '__array__'):  # data is array-like
    data = np.asarray(data).tolist()
```

The `np.asarray` call will convert any array-like object to a base `ndarray` instance, which we can then convert to a list and save as usual.

Testing will be slightly harder, but not too tough. The NumPy Developer Docs have a [nice example of a custom array-like object](https://numpy.org/devdocs/user/basics.dispatch.html) that could be used.

### Additional context

NEP 16: https://numpy.org/neps/nep-0016-abstract-array.html
NEP 18: https://numpy.org/neps/nep-0018-array-function-protocol.html
NEP 22: https://numpy.org/neps/nep-0022-ndarray-duck-typing-overview.html

vyasr:
I like this feature, but before we implement, I'll play devil's advocate. Should we instead move the other way and stop automatically serializing numpy arrays? The obvious knee-jerk reaction is "no, of course we should support numpy, so much of our data is numpy arrays." However, consider that in 2.0 we are dropping support for non-str keys in part because we can't represent them faithfully (see #316): under the hood we convert them to str. Similarly, is it expected behavior if you can do

```
x = np.random.rand(5)
job.doc.arr = x
assert type(job.doc.arr) == type(x)  # this will fail
```

One obvious counterargument is that numpy ducktypes lists to numpy arrays, so in many cases things will work as expected. However, there are also plenty of cases where it won't work (`job.doc.arr + 1` isn't valid, even though `x + 1` is), and there are subtle edge cases (possible `FutureWarning`s that are actually errors if you try to do comparisons with numpy arrays of strings). If we're going to force our users to convert `int` keys to `str`, maybe we should err on the side of safety here and also ask users to do `job.doc.arr = x.tolist()`. @glotzerlab/signac-committers would be curious to hear opinions.

bdice:
@vyasr I think that's a solid argument. I have some additional thoughts.

Removing the feature could be a significant UX regression:
[...]

PR #477, opened 2021-01-22
Add support for zero-dimensional numpy arrays.

### LINKED ISSUE #476: 0-dimensional numpy arrays not supported in statepoint values
### Description

Using a 0-dimensional numpy array
```
import numpy as np
project = signac.init_project("debug")  
project.open_job({"mass": np.array(1)})
```
results in 
`TypeError: 'float' object is not iterable`

### To reproduce
(see above)
0D arrays are needed for unyt quantities:
```
import unyt as u
project = signac.init_project("debug")  
project.open_job({"mass": 1.0 * u.gram})
```

### Error output
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-25-e12695a17e5e> in <module>
      1 import unyt as u
      2 project = signac.init_project("debug")
----> 3 project.open_job({"mass": 1.0 * u.gram})
~/miniconda3/envs/planckton-dev/lib/python3.7/site-packages/signac/contrib/project.py in open_job(self, statepoint, id)
    613         if id is None:
    614             # second best case
--> 615             job = self.Job(project=self, statepoint=statepoint)
    616             if job._id not in self._sp_cache:
    617                 self._sp_cache[job._id] = job.statepoint._as_dict()
~/miniconda3/envs/planckton-dev/lib/python3.7/site-packages/signac/contrib/job.py in __init__(self, project, statepoint, _id)
     87 
     88         # Set statepoint and id
---> 89         self._statepoint = SyncedAttrDict(statepoint, parent=_sp_save_hook(self))
     90         self._id = calc_id(self._statepoint()) if _id is None else _id
     91 
~/miniconda3/envs/planckton-dev/lib/python3.7/site-packages/signac/core/synceddict.py in __init__(self, initialdata, parent)
     82             self._data = {
     83                 self._validate_key(k): self._dfs_convert(v)
---> 84                 for k, v in initialdata.items()
     85             }
     86         self._suspend_sync_ = 0
~/miniconda3/envs/planckton-dev/lib/python3.7/site-packages/signac/core/synceddict.py in <dictcomp>(.0)
     82             self._data = {
     83                 self._validate_key(k): self._dfs_convert(v)
---> 84                 for k, v in initialdata.items()
     85             }
     86         self._suspend_sync_ = 0
~/miniconda3/envs/planckton-dev/lib/python3.7/site-packages/signac/core/synceddict.py in _dfs_convert(self, root)
    121                 return root.item()
    122             elif isinstance(root, numpy.ndarray):
--> 123                 return _SyncedList(root.tolist(), self)
    124         return root
    125 
~/miniconda3/envs/planckton-dev/lib/python3.7/site-packages/signac/core/synceddict.py in __init__(self, iterable, parent)
     24     def __init__(self, iterable, parent):
     25         self._parent = parent
---> 26         super(_SyncedList, self).__init__(iterable)
     27 
     28     def __deepcopy__(self, memo):
TypeError: 'float' object is not iterable
```

### System configuration

Darwin-18.7.0-x86_64-i386-64bit
3.7.9 | packaged by conda-forge | (default, Dec  9 2020, 20:58:55) 
[Clang 11.0.0 ]
1.5.0

bdice:
## Description
Adds support for zero-dimensional NumPy arrays like `numpy.array(1.0)`.

I note that this is _not_ a problem in our JSON encoder, only in `_dfs_convert` because it assumes that all NumPy arrays can be used to construct a SyncedList (which is not true for 0-d arrays, since SyncedLists must be 1-d or higher).

## Motivation and Context
Resolves #476.

## Types of Changes
- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/core/synceddict.py
+++ b/signac/core/synceddict.py
@@ -128,10 +128,13 @@ class _SyncedDict(MutableMapping):
         elif type(root) in (list, tuple):
             return _SyncedList(root, parent=self)
         elif NUMPY:
-            if isinstance(root, numpy.number):
+            if isinstance(root, numpy.ndarray):
+                if root.shape == ():
+                    return root.item()
+                else:
+                    return _SyncedList(root.tolist(), parent=self)
+            elif isinstance(root, numpy.number):
                 return root.item()
-            elif isinstance(root, numpy.ndarray):
-                return _SyncedList(root.tolist(), parent=self)
         return root
 
     @classmethod
--- a/tests/test_numpy_integration.py
+++ b/tests/test_numpy_integration.py
@@ -33,6 +33,13 @@ class TestNumpyIntegration(TestProjectBase):
             assert [i] == job.sp.a
             assert numpy.array([i]) == job.sp.a
 
+    def test_store_zero_dim_array_in_sp(self):
+        # Zero-dimensional array
+        value = 1.0
+        job = self.project.open_job(dict(a=numpy.array(value))).init()
+        assert value == job.sp.a
+        assert numpy.array(value) == job.sp.a
+
     def test_store_array_in_doc(self):
         for i in range(10):
             job = self.project.open_job(dict(a=i))
@@ -42,3 +49,14 @@ class TestNumpyIntegration(TestProjectBase):
             assert i == job.sp.a
             assert (numpy.array([i, i, i]) == job.doc.array).all()
             assert [i] * 3 == job.doc.array
+
+    def test_store_zero_dim_array(self):
+        # Zero-dimensional array
+        value = 1.0
+        job = self.project.open_job(dict(a=numpy.array(value))).init()
+        assert value == job.sp.a
+        assert numpy.array(value) == job.sp.a
+        job.doc.array = numpy.array(value)
+        numpy.testing.assert_equal(job.doc.array, numpy.array(value))
+        assert value == job.doc.array
+        assert numpy.array(value) == job.doc.array
--- a/tests/test_numpy_integration.py
+++ b/tests/test_numpy_integration.py
@@ -33,13 +33,6 @@ class TestNumpyIntegration(TestProjectBase):
             assert [i] == job.sp.a
             assert numpy.array([i]) == job.sp.a
 
-    def test_store_zero_dim_array_in_sp(self):
-        # Zero-dimensional array
-        value = 1.0
-        job = self.project.open_job(dict(a=numpy.array(value))).init()
-        assert value == job.sp.a
-        assert numpy.array(value) == job.sp.a
-
     def test_store_array_in_doc(self):
         for i in range(10):
             job = self.project.open_job(dict(a=i))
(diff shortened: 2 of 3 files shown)


tests/test_numpy_integration.py
```diff
@@ -42,3 +49,14 @@ def test_store_array_in_doc(self):
             assert i == job.sp.a
             assert (numpy.array([i, i, i]) == job.doc.array).all()
             assert [i] * 3 == job.doc.array
+
+    def test_store_zero_dim_array(self):
+        # Zero-dimensional array
+        value = 1.0
+        job = self.project.open_job(dict(a=numpy.array(value))).init()
```
vyasr, line 56: Store a normal Python int (or float if you like 1.0 instead of 1) in the doc instead of a numpy array so that this test only tests the doc and the other tests the statepoint. Also rename this test appropriately (suffix with `_doc`). That way each of these tests is independent and it'll make it just a little easier to pinpoint issues in the future.

vyasr (approved):
One small request, otherwise this looks fine to me. I'll wait for @jennyfothergill to test this before taking any further action.

signac/core/synceddict.py
```diff
@@ -128,10 +128,13 @@ def _dfs_convert(self, root):
         elif type(root) in (list, tuple):
             return _SyncedList(root, parent=self)
         elif NUMPY:
-            if isinstance(root, numpy.number):
+            if isinstance(root, numpy.ndarray):
+                if root.shape == ():
+                    return root.item()
```
jennyfothergill, line 133: For a unyt.quantity, this strips the unit value.
```
import unyt as u
project = signac.init_project("debug")  
job = project.open_job({"mass": 1.0 * u.gram})
job.sp["mass"]
```
is 1.0 of type float
Would there be an issue with 
```
                if root.shape == ():
                    return root
```
?

jennyfothergill (commented):
This change allows for a unyt quantity to be used in a statepoint without error, but the unit is stripped and could lead to unexpected behavior.

signac/core/synceddict.py
vyasr: Yes, this is an unfortunate limitation of the JSON backend and is basically the same reason why the lines below convert `numpy.ndarray` into a JSONList:
```
>>> import numpy as np
>>> job = project.open_job({'a': np.array([1])})
>>> job
Job(project=Project.get_project('/Users/vyasramasubramani/tmp2'), statepoint={'a': [1]})
>>> job.sp.a
[1]
>>> type(job.sp.a)
<class 'signac.core.synced_collections.collection_json.JSONList'>
```
There's no way to store a numpy array in JSON (just try doing `json.dumps(numpy.array([1]))` and you'll see that it fails), so we have to convert it into something like a list. I previously raised errors like this as a possible reason not to automatically convert and force users to do it up front (see [this thread](https://github.com/glotzerlab/signac/issues/380#issuecomment-681178761)), but the upshot is that there's no easy way to support a faithful representation of objects like numpy arrays in JSON. unyt ends up falling in the same bucket.

bdice: @jennyfothergill: @vyasr's comment is accurate, signac requires state points to be JSON-encodable, which limits us to basic types like `dict`, `list`, `float`, `int`, `str`. If you need to store unyt quantities to disk, you may want to adopt one of the methods suggested in the unyt docs: https://unyt.readthedocs.io/en/stable/usage.html#writing-data-with-units-to-disk If you need to use a unyt quantity in a job state point or document, you might be better off writing your own wrapper that converts quantities to something like `{"value": 1.0, "unit": "gram"}` (that's a JSON-encodable dictionary).

jennyfothergill: OK this solution `{"value": 1.0, "unit": "gram"}` will work. Should there be some sort of warning for unyt.quantities? I don't know how commonly they're used.

jennyfothergill (approved):

signac/core/synceddict.py
vyasr: Since numpy arrays are the only object that we currently convert, what I would suggest is that we throw a warning (using `warnings.warn`) whenever the `_SyncedDict` or `JSONDict` detect something that is a numpy array. @bdice thoughts? I've advocated that before as well, I think it's helpful to the user to indicate when we're converting something that might be unexpected.

bdice: This discussion is closely related to #380. I don't really know what to do, we've been back and forth on "transparent" NumPy support for a long time. It's fine with me if someone wants to make that PR to add a warning.

vyasr: @bdice have we discussed this anywhere else? I don't recall any discussion about the viability of transparent numpy support outside of #380, which is where I suggested adding the warning. Is there any other context that I'm missing with respect to this decision? If not, I'd suggest that @jennyfothergill just add the warning here and we can move on.

PR #481, opened 2021-01-26
Feature/synced collection/reorg

vyasr:
## Description

Moves synced_collections out of core into the top level of signac and reorganizes the internals.

## Motivation and Context

We plan to flatten the structure of signac and eventually extract synced_collections into a separate package, both of which will be facilitated by this PR.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:


signac/contrib/project.py
```diff
@@ -26,8 +26,10 @@
 from ..common.config import Config, get_config, load_config
 from ..core import json
 from ..core.h5store import H5StoreManager
-from ..core.synced_collections.collection_json import BufferedJSONDict
 from ..sync import sync_projects
+
+# from ..synced_collections.backends.collection_json import
```
bdice, line 31: Delete this?

signac/errors.py
```diff
@@ -80,6 +73,7 @@ class KeyTypeError(TypeError):
     "IncompatibleSchemaVersion",
     "InvalidKeyError",
     "JobsCorruptedError",
+    "KeyTypeError",
```
bdice, line 76: You may want to merge `master` up into your PR -- I also did this same thing in another PR.

signac/synced_collections/backends/__init__.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2020 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+"""This subpackage defines supported backends.
+
+No backends are imported by default. Users should import desired backends as
+needed.
+"""
+__all__ = [""]
```
bdice, line 9: Should it be an empty list, or a list containing an empty string? Is this line needed? If unsure, I vote to delete.

signac/synced_collections/buffers/__init__.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2020 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+"""This subpackage defines the buffering protocol and supported implementations.
+
+No buffers are imported by default. Users should import desired buffers as
+needed.
+"""
+__all__ = [""]
```
bdice, line 9: Should it be an empty list, or a list containing an empty string? Is this line needed? If unsure, I vote to delete.

signac/synced_collections/errors.py
```diff
@@ -4,10 +4,7 @@
 """Errors raised by synced collections."""
 
 
-from ..errors import Error
-
-
-class BufferException(Error):
+class BufferException(Exception):
```
bdice, line 7: Should this be a subclass of `RuntimeError`?

signac/synced_collections/errors.py
```diff
@@ -4,10 +4,7 @@
...
     """An exception occured in buffered mode."""
```
bdice, line 8: We need to clarify that this is not just an exception "in" buffered mode, since that sounds like _any_ exception raised during a buffered context will be re-raised by this class. I'm not sure if my suggestion solves it... but I tried to match the style of the other errors' docstrings.
```suggestion
    """Raised when an exception related to buffering occurs."""
```

bdice (changes requested):
Questions:
1. There are a few signac-specific things in synced collections, such as the "dots in keys" validator and automatic handling of NumPy values. How should we plan to separate these things if/when synced collections becomes its own package?
2. Can we consider a name scheme like `json_collection` instead of `collection_json`? It is weird that the backend module names have an inverted structure when all other components are named plainly.

signac/errors.py
vyasr: Hmm I just did that earlier today and this didn't get merged. I just double-checked the diff from that merge and this didn't change, I'm not sure why. We can double check when I open the synced_collections->master PR.

signac/synced_collections/backends/__init__.py
vyasr: Empty list vs empty string are functionally equivalent, but an empty string is necessary to avoid mypy complaining (either that, or explicitly including a type annotation). We could remove this line entirely, but I left it here in anticipation of a discussion of whether we agree that no backends should be imported by default. While I don't want backend classes to be imported at the top level of synced collections (`synced_collections.JSONDict` shouldn't work), I would prefer to be able to do `from synced_collections.backends import JSONDict` rather than `from synced_collections.backends.collection_JSON import JSONDict`. In that case we would populate the `__all__` list appropriately.

signac/synced_collections/buffers/__init__.py
vyasr: Same discussion as above.

vyasr:
> Questions:
> 
>     1. There are a few signac-specific things in synced collections, such as the "dots in keys" validator and automatic handling of NumPy values. How should we plan to separate these things if/when synced collections becomes its own package?
The dots in keys is not signac specific. We can't allow dots in keys for any `SyncedAttrDict` because it introduces ambiguities in the `getattr` syntax: is `synced_dict['a.b.c'] = 1` equivalent to `synced_dict['a'] = {'b': {'c': 1}}`? It shouldn't be, but we can't distinguish. As for numpy arrays, I don't think of that as signac-specific since it does affect any JSON collection. We might need to make it more backend specific than the code currently is (it's not only applied to JSON collections, it applies to any `SyncedList` class), but I do think the logic belongs somewhere in synced_collections, not in signac.

> 
>     2. Can we consider a name scheme like `json_collection` instead of `collection_json`? It is weird that the backend module names have an inverted structure when all other components are named plainly.

This naming was specifically requested by @csadorf IIRC.

signac/synced_collections/backends/__init__.py
bdice: I don't think we want to trigger loading any backends by default. I would rather the modules be very lightweight and import only what is specifically needed by a user, at the cost of having verbose names.

bdice: PEP8 says we need `__all__` to be an empty list, not `[""]`. https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces

It's fine if you have to add a type annotation. I'm approving conditional on that change.

bdice (approved):
PEP8 says we need `__all__` to be an empty list, not `[""]`. https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces

It's fine if you have to add a type annotation. I'm approving conditional on that change.

signac/synced_collections/backends/__init__.py
vyasr: Sounds good, no backends it is.

PR #484, opened 2021-01-27
Feature/synced collections

### LINKED ISSUE #196: _SyncedList dictionary children should be _SyncedDicts

### LINKED ISSUE #234: Calling `_SyncedDict.values()` gives `dict` objects

### LINKED ISSUE #249: Proposal: Unify dict classes and improve buffering and synchronization

### LINKED ISSUE #316: Unexpected errors when using ints as keys for JSONDict

### LINKED ISSUE #383: Handle NumPy extended precision values.
### Description

Related to #378 and #380. Currently it appears that signac is unable to transparently cast extended-precision NumPy values like `np.float128(9827389123.18237128)`.

I suggest the following:
1. Validators introduced in #378 should raise an error (and not accept extended-precision values as valid).
2. Tests of NumPy numeric types need to consider extended-precision values.

### To reproduce

```python
import numpy as np
import signac

doc = signac.JSONDict('test.json')
value = np.float128(1892739817.8127812)
doc.value = value
```

### Error output

A `RecursionError` is raised when attempting to call `value.item()` repeatedly (calling `value.item()` on an extended-precision value just returns itself, rather than a Python built-in type with an equivalent value): https://github.com/glotzerlab/signac/blob/7086d8981c926703a023654d1c59bbedcfae6298/signac/core/json.py#L28-L29

```pytb
  File "test.py", line 6, in <module>
    doc.value = value
  File "/home/bdice/code/signac/signac/core/attrdict.py", line 42, in __setattr__
    self.__setitem__(key, value)
  File "/home/bdice/code/signac/signac/core/synceddict.py", line 200, in __setitem__
    self._synced_save()
  File "/home/bdice/code/signac/signac/core/synceddict.py", line 187, in _synced_save
    self.save()
  File "/home/bdice/code/signac/signac/core/synceddict.py", line 192, in save
    self._save()
  File "/home/bdice/code/signac/signac/core/jsondict.py", line 282, in _save
    blob = json.dumps(data).encode()
  File "/home/bdice/code/signac/signac/core/json.py", line 42, in dumps
    return CustomJSONEncoder(sort_keys=sort_keys, indent=indent).encode(o)
    return _iterencode(o, 0)
  File "/home/bdice/code/signac/signac/core/json.py", line 29, in default
    return o.item()
RecursionError: maximum recursion depth exceeded while calling a Python object
```

### System configuration

 - Operating System: Ubuntu, AMD64
 - Version of Python: 3.8
 - Version of signac: 1.4.0

### LINKED ISSUE #397: Exceeding the buffer size in parallel context causes errors during flush

### LINKED ISSUE #465: Use custom context managers rather than contextlib

vyasr:
## Description

**This pull request should not be squashed upon merge.**
This pull request adds the `synced_collections` subpackage, which introduces a new `SyncedCollection` abstract base class defining a standardized interface for a `collections.abc.Collection`-like object that must be transparently synchronized with some underlying resource. These collections are designed to replace the `_SyncedDict`, `SyncedAttrDict`, and `JSONDict` classes with a new framework of classes that make it possible to easily interchange different data types (aside from dictionaries) and different resources (aside from JSON files). The new classes generalize the validation process so that different data types and/or resource types can specify the exact validation required. The new classes also generalize the buffering protocol for synced collections so that new buffering methods can be easily defined and interchanged. In addition, the new classes are more performant than the old versions, with the added bonus of thread-safety in parallel environments.

 All of this is implemented via extensive inheritance. The new code is unfortunately substantially more complicated than the old code, but the separation of concerns is much clearer with the new classes and the new code is much more easily extensible. There is much less bleeding between the responsibilities of the Job class and the state point as well, while the new `BufferedJSONDict` is a drop-in replacement for the old `JSONDict` class in the Job and Project documents.

**Breaking change**: There is one very minor breaking change introduced by this PR. The `force_write` option in buffered mode is not supported by the new code. I have added a compatibility layer so that the argument will be processed, so it's not an API break, but it is a minor change in behavior because it will not actually force writing. This feature is extremely minor, and it's not used by anyone I'm aware of, so I'm comfortable with making this change within a minor release. As a result, I don't think that this PR requires a new major release of signac.

**Documentation**: Should I write up something for signac-docs? I started writing this up, but a lot of the associated documentation is internal-facing or at least for the purpose of others who wish to extend these classes. From signac's perspective, the existing documentation of the job/project document and state point already cover the necessary points. I added synced collections to the API documentation, and the docstrings within these classes are extensive.

## Motivation and Context

Resolves #234 
Resolves #196 
Resolves #397 
Resolves #316 
Resolves #465 
Resolves #249
Resolves #383 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


signac/synced_collections/backends/collection_mongodb.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def __init__(self, collection=None, uid=None, parent=None, **kwargs):
+        if not MONGO:
+            raise RuntimeError(
+                "The PyMongo package must be installed to use the MongoDBCollection."
+            )
+
+        self._collection = collection
+        self._uid = uid
+        super().__init__(parent=parent, **kwargs)
```
bdice, line 56: I don't understand how this MongoDB data model is supposed to work with nested synced collections. I don't think I'm following how the `collection` / `uid` are used in conjunction with the `parent` to write the correct data to the correct Mongo document.

signac/synced_collections/backends/collection_mongodb.py
```diff
@@ -0,0 +1,187 @@
...
+
+    **Thread safety**
+
+    The MongoDBCollection is not thread-safe.
+
+    Parameters
+    ----------
+    collection : :py:class:`pymongo.collection.Collection`
+        The MongoDB client in which to store data.
+    uid : dict
```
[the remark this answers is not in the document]
bdice: We have a special key "data" that cannot exist in the `uid` mapping. This should probably be documented.

signac/synced_collections/backends/collection_mongodb.py
```diff
@@ -0,0 +1,187 @@
...
+        The unique key-value mapping added to the data and stored in the document
+        so that it is uniquely identifiable in the MongoDB collection.
+
+    Warnings
+    --------
+    The user is responsible for providing a unique id such that there are no
+    possible collisions between different :class:`SyncedCollection` instances
+    stored in the same MongoDB collection. Failure to do so may result data
```
bdice, line 40:
```suggestion
    stored in the same MongoDB collection. Failure to do so may result in data
```

signac/synced_collections/backends/collection_mongodb.py
```diff
@@ -0,0 +1,187 @@
...
+    corruption if multiple documents are found to be apparently associated with
+    a given :class:`SyncedCollection`.
```
bdice, line 42:
```suggestion
    a given ``uid``.
```

signac/synced_collections/backends/collection_mongodb.py
```diff
@@ -0,0 +1,187 @@
...
+
+        >>> doc['foo'] = dict(bar=True)
+        >>> doc
+        {'foo': {'bar': True}}
+        >>> doc.foo.bar = False
+        {'foo': {'bar': False}}
+
+    Warnings
+    --------
+
```
bdice, line 117:
```suggestion
```

signac/synced_collections/backends/collection_mongodb.py
```diff
@@ -0,0 +1,187 @@
...
+        assert 'foo' in doc
+        del doc['foo']
+
+    .. code-block:: python
+
+        >>> doc['foo'] = dict(bar=True)
+        >>> doc
+        {'foo': {'bar': True}}
+        >>> doc.foo.bar = False
+        {'foo': {'bar': False}}
```
bdice, line 113: Fix these code blocks (match whether to use doctest style / Examples heading) or remove them. Also applies to all subsequent example code blocks in all collection classes.

signac/synced_collections/backends/collection_mongodb.py
```diff
@@ -0,0 +1,187 @@
...
+    Parameters
+    ----------
+    collection : pymongo.collection.Collection, optional
+        A :class:`pymongo.collection.Collection` instance (Default value = None).
+    uid: dict, optional
+        The unique key-value mapping identifying the collection (Default value = None).
+    data: non-str :py:class:`collections.abc.Sequence`, optional
+        The intial data pass to MongoDBList. Defaults to `list()`.
+    parent: MongoDBCollection, optional
+        A parent instance of MongoDBCollection (Default value = None).
```
bdice, line 167:
```suggestion
    uid : dict, optional
        The unique key-value mapping identifying the collection (Default value = None).
    data : non-str :py:class:`collections.abc.Sequence`, optional
        The initial data (Default value = []).
    parent : MongoDBCollection, optional
        A parent instance of MongoDBCollection (Default value = None).
```

signac/synced_collections/backends/collection_mongodb.py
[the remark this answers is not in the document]
bdice: The default value of `uid` is None in this class and the `MongoDBDict` and `MongoDBList` classes. That seems invalid since this is expecting a `dict`-like object.

signac/synced_collections/backends/collection_redis.py
```diff
@@ -0,0 +1,153 @@
...
+        """str: The key associated with this collection stored in Redis."""
+        return self._key
+
+    def __deepcopy__(self, memo):
+        # The underlying Redis client cannot be deepcopied.
+        raise TypeError("RedisCollection does not support deepcopying.")
+
+
+class RedisDict(RedisCollection, SyncedAttrDict):
+    """A dict-like mapping interface to a persistent Redis-database.
```
bdice, line 69:
```suggestion
    """A dict-like mapping interface to a persistent Redis database.
```

signac/synced_collections/backends/collection_redis.py
```diff
@@ -0,0 +1,153 @@
...
+    Parameters
+    ----------
+    client: redis.Redis, optional
+        A redis client (Default value = None).
+    key: str, optional
+        The key of the  collection (Default value = None).
+    data: :py:class:`collections.abc.Mapping`, optional
+        The intial data pass to RedisDict. Defaults to `dict()`
+    parent: RedisCollection, optional
+        A parent instance of RedisCollection (Default value = None).
```
bdice, line 96:
```suggestion
    client : redis.Redis, optional
        A Redis client (Default value = None).
    key : str, optional
        The key of the collection (Default value = None).
    data : :py:class:`collections.abc.Mapping`, optional
        The initial data (Default value = {}).
    parent : RedisCollection, optional
        A parent instance of RedisCollection (Default value = None).
```

signac/synced_collections/backends/collection_redis.py
```diff
@@ -0,0 +1,153 @@
...
+
+    Warnings
+    --------
+
```
bdice, line 100:
```suggestion
```

signac/synced_collections/backends/collection_redis.py
```diff
@@ -0,0 +1,153 @@
+# Copyright (c) 2020 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+"""Implements a Redis SyncedCollection backend."""
+import json
+
+from .. import SyncedAttrDict, SyncedCollection, SyncedList
+
+
+class RedisCollection(SyncedCollection):
```
bdice, line 10: Many suggestions/questions/comments from the MongoDB backend also apply to this file.

signac/synced_collections/backends/collection_redis.py
```diff
@@ -0,0 +1,153 @@
...
+
+    .. code-block:: python
+
+        synced_list = RedisList('data')
+        synced_list.append("bar")
+        assert synced_list[0] == "bar"
+        assert len(synced_list) == 1
+        del synced_list[0]
+
+    .. warning::
```
bdice, line 127: Use a `Warnings` header. No indent needed.

signac/synced_collections/backends/collection_redis.py
```diff
@@ -0,0 +1,153 @@
...
+    Parameters
+    ----------
+    client: redis.Redis, optional
+        A Redis client (Default value = None).
+    key: str, optional
+        The key of the  collection (Default value = None).
+    data: non-str :py:class:`collections.abc.Sequence`, optional
+        The intial data pass to RedisList. Defaults to `list()`
+    parent: RedisCollection, optional
+        A parent instance of RedisCollection (Default value = None).
```
bdice, line 146:
```suggestion
    client : redis.Redis, optional
        A Redis client (Default value = None).
    key : str, optional
        The key of the  collection (Default value = None).
    data : non-str :py:class:`collections.abc.Sequence`, optional
        The initial data (Default Value = []).
    parent : RedisCollection, optional
        A parent instance of RedisCollection (Default value = None).
```

signac/synced_collections/backends/collection_zarr.py
```diff
@@ -0,0 +1,204 @@
...
+
+try:
+    import numcodecs
+
+    ZARR = True
+except ImportError:
+    ZARR = False
+
+
+class ZarrCollection(SyncedCollection):
```
bdice, line 17: Skipping review of this file for now. Several comments from other backends need to be applied here too.

bdice (changes requested):
I have reviewed half of the files in this PR (26/51). I'll submit what I have for now, and continue with a second review pass later.

signac/synced_collections/backends/collection_json.py
```diff
@@ -0,0 +1,387 @@
...
+            write_concern=write_concern,
+            data=data,
+            parent=parent,
+            *args,
+            **kwargs,
+        )
+
+
+class JSONList(JSONCollection, SyncedList):
+    """A non-string sequence interface to a persistent JSON file.
```
bdice, line 237: Make the first line less awkward. Something like:
```suggestion
    """A list-like data structure that synchronizes with a persistent JSON file.
    
    Only non-string sequences are supported by this class.
```

signac/synced_collections/backends/collection_json.py
```diff
@@ -0,0 +1,387 @@
...
+
+    .. code-block:: python
```
bdice, line 239: Use doctest style and an `Examples` block.

signac/synced_collections/backends/collection_json.py
```diff
@@ -0,0 +1,387 @@
...
+
+    Parameters
+    ----------
+    filename : str, optional
+        The filename of the associated JSON file on disk (Default value = None).
+    write_concern : bool, optional
+        Ensure file consistency by writing changes back to a temporary file
+        first, before replacing the original file (Default value = None).
+    data : non-str :class:`collections.abc.Sequence`, optional
+        The initial data pass to :class:`JSONList `. If ``None``, Defaults to
```
bdice, line 255:
```suggestion
        The initial data passed to :class:`JSONList `. If ``None``, defaults to
```

signac/synced_collections/backends/collection_json.py
```diff
@@ -0,0 +1,387 @@
...
+        ``[]`` (Default value = None).
+    parent : JSONCollection, optional
+        A parent instance of JSONCollection or None (Default value = None).
+
+    Warnings
+    --------
+    While the :class:`JSONList` object behaves like a list, there are important
```
bdice, line 262:
```suggestion
    While the :class:`JSONList` object behaves like a :class:`list`, there are important
```

signac/synced_collections/backends/collection_mongodb.py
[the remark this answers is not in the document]
bdice: @vyasr Maybe I'm misunderstanding. I thought that `uid` was meant to be a unique identifier like a hash or primary key integer, except as a dictionary like `{"_id": "abc123def456"}`. From my understanding, `uid=None` will raise an error because it's not a mapping. (Or perhaps it will return an arbitrary record... which would be a bug imo.) That's intentional, then?

bdice (changes requested):
An additional pass of review. I got through 2/3 of the files this time (including re-review of some files that were changed in the last round). Please take special note of my comments on the backend files -- many of the suggested changes apply to all backends, and I only wrote them one time. I'll try to finish reviewing the rest of the files before Sunday.

signac/synced_collections/buffers/memory_buffered_collection.py
```diff
@@ -0,0 +1,278 @@
...
+        Parameters
+        ----------
+        force : bool
+            If True, force a flush even in buffered mode (defaults to False). This
+            parameter is used when the buffer is filled to capacity.
+
+        Returns
+        -------
+        dict
+            Mapping of filename and errors occured during flushing data.
```
bdice, line 270:
```suggestion
            Mapping from filenames to errors that occurred while flushing data.
```

signac/synced_collections/buffers/serialized_file_buffered_collection.py
```diff
@@ -0,0 +1,309 @@
...
+    safety of an individual collection backend; the backend must support thread
+    safe writes to the underlying resource in order for a buffered version using
+    this class to be thread safe for general use. The thread safety guaranteed
+    by this class only concerns buffer reads, writes, and flushes. All these
+    operations are serialized because there is no way to prevent one collection
+    from triggering a flush while another still thinks its data is in the cache.
+
+    Parameters
+    ----------
+    filename: str, optional
```
bdice, line 56:
```suggestion
    filename : str, optional
```

signac/synced_collections/buffers/serialized_file_buffered_collection.py
```diff
@@ -0,0 +1,309 @@
...
+
+import hashlib
+import json
+
+from ..errors import MetadataError
+from ..utils import SyncedCollectionJSONEncoder
+from .file_buffered_collection import FileBufferedCollection
+
+
+class SerializedFileBufferedCollection(FileBufferedCollection):
```
bdice, line 19: Are we going to use the shared memory buffer (instead of the serialized file buffer) in signac 1.7, signac 2.0, or is this undecided yet?

bdice (changes requested):
Final round of comments.

bdice:
I ran benchmarks and profiles with this command: `./benchmark.py run -N 10 100 1000 -k 10 -s 1000 -o synced_collections.txt`

My observations are that this PR is about 2x slower than `master` for loading statepoints (for profiling enabled _and_ profiling disabled). I have a breakdown below.

## Synced Collections (this PR)
image

## master (close to v1.6.0)
image

The slowdown is coming from these two places. Correcting these two problems would make the new code just as fast as the old code, as far as I can tell.

1. It appears that data is validated after loading from the resource. For instance, the `_StatePointDict` reads from JSON, then validates that data is JSON-encodable during the call to `SyncedCollection._update`. Is this strictly necessary? The old code does not perform such a validation when loading from a resource. Perhaps we need a parameter for `SyncedCollection._update` that skips validation?
image

2. The method `SyncedAttrDict.__setattr__` is consuming a considerable portion of time. It has to traverse a deep hierarchy of classes and is probably called more times than is needed during instantiation. I think it might be possible to defer the "fancy" attribute access until after the instance is ready for use, but I could not find a good solution for this (attempted a "guard" attribute `_attribute_access_enabled` but I couldn't work around the issue; using a metaclass with [this hack](https://stackoverflow.com/a/34065724/) led to a conflict).

image

vyasr:
@bdice thanks for doing some benchmarking. I'll go through your comments later, for now here are my thoughts regarding the two points you brought up on performance.

1. I think it would be fine to remove validation entirely from `_update`. There are some API methods that call `_update` internally (`reset`, `update`, and `__setitem__`) for the `SyncedAttrDict` off the top of my head), and I think at least the first two rely on `_update` for validation currently (`__setitem__` does its own IIRC which is another unnecessary inefficiency), so we would need to add `self.validate` calls to those methods. We can generally rely on the backend implementations to detect if the user attempts to create a `SyncedCollection` that points to an invalid resource. The small trade-off here is that removing validation in `_update` would prevent any _additional_ validation of preexisting data. I don't think this would lead to destructive changes, only to unexpected behavior.
Here's an example (completely untested, just based on my recollection of all the implementation details). Consider the case where a user creates a JSON file manually that contains dots in keys e.g. `json.dump(f, {'foo.bar': 1}`. If we remove validation in `_update`, the user would be able to initialize a `JSONDict` pointing to this file and make modifications, and I think the result would basically be that the key would appear to not exist. If the user did `JSONDict(fn)['foo.bar'] = 2`, the JSON file will look like `{'foo.bar': 1, 'foo': {'bar': 2}}`. If we wanted to offer some protection against this behavior, we could add a constructor parameter `validate_on_init` to run `self._validate(self())` at the end of the constructor, but it would have to be implemented at the final class level (e.g. `JSONDict`) since it will only be possible after all backend and data type initialization is done. Alternatively, we could just document this behavior and leave it unhandled.
2. I'm kind of surprised that this is so slow, the only work that `__setattr__` should be doing that is affected by the deep hierarchy is constructing the set of protected keys for each class, but those are cached after the first time so it should only be expensive the very first time `__setattr__` is called for any instance of the class. Without further investigation I can't say for sure why this is slowing it down. However, I would like to propose an alternative solution in lieu of further benchmarking.
I have previously mentioned that I am uncomfortable with our overriding `__getattr__`, `__setattr__`, and `__delattr__`. While we're basically stuck with this decision for signac, I would not be unhappy if we could avoid introducing this into synced collections since I'm much more uncomfortable promising such a feature in a generic data structure class than I am with promising it within the context of a signac document. What if we removed this feature entirely (converted `SyncedAttrDict` to `SyncedDict`), then extend `JSONDict` in signac to just add those features? That would allow us to bypass all of these issues, and isolate this API decision just to signac.
If you are in favor of this decision, it does require that we address one other point which I had been intentionally ignoring up to this point since it's not critical to any implementation details so far and it's quite tricky to resolve. Right now, nested `SyncedCollections` are handled by using `is_base_type` within `from_base` to determine what type of collection to create based on a registry of available types. One limitation of this approach is that it assumes that the first type to return `True` from `is_base_type` is what the user wants. Currently, this is not an issue because there only exists one type for every combination of backend and data type, but implementing the change I discussed above would introduce a subclass of `JSONDict` that also satisfied the same requirements. We avoid this entirely with the `_StatePointDict` by not registering it, meaning that dicts nested inside the `_StatePointDict` are actually instances of `JSONDict`, but this is not a problem since `_load` and `_save` calls eventually get passed up to the top-level `_StatePointDict` instance. However, if we extended `JSONDict` into a `_JSONAttrDict` within signac, we would need that to also persist for nested dictionaries. Currently the resolution order in `from_base` is arbitrary (but deterministic) based on the order in which `SyncedCollection` types are registered on construction. Introducing a well-defined, user-controlled order as an API promise is one solution, but I don't think it's general enough because the registry is on the package level. This means that a new SyncedCollection type registered by the user or another package would also modify the resolution order within signac, and vice versa. I'm not sure what the best way to overcome that problem is, and would be interested to hear ideas.
We could also punt on making this decision now by just overriding the behavior of `from_base` in the new `JSONAttrDict`, so we don't _have_ to resolve this now, but I wanted to bring it up in case someone has ideas for a better approach.

signac/synced_collections/backends/collection_mongodb.py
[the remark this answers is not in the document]
vyasr: You're right that `uid` is meant to be a unique identifier. However, a `MongoDBCollection` can contain another `MongoDBCollection` (ad infinitum) and only the top-level collection has a uid, while the uid is None for all others. This is not different than the filename only being set for a top-level `JSONDict`.

signac/synced_collections/backends/collection_json.py
```diff
@@ -0,0 +1,395 @@
...
+from .. import SyncedAttrDict, SyncedCollection, SyncedList
+from ..buffers.memory_buffered_collection import SharedMemoryFileBufferedCollection
+from ..buffers.serialized_file_buffered_collection import (
+    SerializedFileBufferedCollection,
+)
+from ..utils import SyncedCollectionJSONEncoder
+from ..validators import json_format_validator
+
+
+# TODO: This method should be removed in signac 2.0.
```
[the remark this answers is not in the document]
vyasr: Yes, I'm also in favor of that. It would also allow us to do things like remove attribute-based dict access in synced_collections and only support it in internal signac classes.

signac/synced_collections/backends/collection_json.py
```diff
@@ -0,0 +1,387 @@
...
+    replacement of the original file. The file replacement operation uses
+    :func:`os.replace`, which is guaranteed to be atomic by the Python standard.
+
+    Parameters
+    ----------
+    filename : str
+        The filename of the associated JSON file on disk.
+    write_concern : bool, optional
+        Ensure file consistency by writing changes back to a temporary file
+        first, before replacing the original file (Default value = False).
```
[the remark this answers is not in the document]
vyasr: The reason to include `*args` and `**kwargs` here is that it's the only way to properly handle the arbitrary multiple inheritance patterns that synced collections are built on, because all arguments are passed to all constructors by super. I don't think documenting these makes sense, because a user should only see the docstrings for arguments that a given class knows how to parse (based on its inheritance hierarchy). For example, `JSONDict` documents `filename` and `write_concern` (handled by `JSONCollection`), `parent` (handled by `SyncedCollection`), and `data` (handled by `SyncedDict`).

signac/synced_collections/backends/collection_json.py
```diff
@@ -0,0 +1,387 @@
...
+    """A dict-like mapping interface to a persistent JSON file.
+
+    Examples
+    --------
+    >>> doc = JSONDict('data.json', write_concern=True)
+    >>> doc['foo'] = "bar"
+    >>> assert doc.foo == doc['foo'] == "bar"
+    >>> assert 'foo' in doc
+    >>> del doc['foo']
+
```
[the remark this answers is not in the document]
vyasr: Nope, these were two separate examples previously and I removed the extra code block directives but not the newline. Thanks.

signac/synced_collections/backends/collection_json.py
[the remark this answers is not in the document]
bdice: The behavior of what happens with args and kwargs is not obvious. In most classes like this, I write explicit docstrings that say args and kwargs are forwarded to the parent constructor.

signac/synced_collections/utils.py
```diff
@@ -0,0 +1,219 @@
...
+            for data_type, id_func in self.abstract_type_identifiers.items():
+                if id_func(obj):
+                    enum_type = self.type_map[obj_type] = data_type
+                    break
+            self.type_map[obj_type] = enum_type
+
+        return enum_type
+
+
+def default(o: Any) -> Dict[str, Any]:  # noqa: D102
```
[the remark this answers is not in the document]
vyasr: 1. This function does not _need_ to be public, but it is the API that would be used to encode a SyncedCollection using `orjson` so if we're exposing the `SyncedCollectionJSONEncoder` this would be nice to expose as well.
2. For the same reason (it's the `orjson` name) it would be helping to retain the association. Perhaps `synced_collection_default` to match `SyncedCollectionJSONEncoder`?
3. `default` is the method you have to override, so this cannot be changed.

signac/synced_collections/utils.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def __enter__(self):
+        pass
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        pass
+
+    def __call__(self):
+        """Allow usage of the context in a function-like manner."""
+        return self
```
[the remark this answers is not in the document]
vyasr: This is required because currently signac exposes the buffered context via `with collection.buffered()`, whereas the new version exposes it as `collection.buffered` (it's an attribute, not a method) and the same goes for "global" buffering. I did mean to ask whether we want to retain that backwards compatibility in signac 2.0; if not, we can mark this as deprecated.

signac/synced_collections/buffers/buffered_collection.py
```diff
@@ -0,0 +1,187 @@
...
+            else:
+                self._root._load()
+
+    def _save_to_buffer(self):
+        """Store data in buffer.
+
+        By default, this method simply calls :meth:`~._save_to_resource`. Subclasses
+        must implement specific buffering strategies.
+        """
+        self._save_to_resource()
```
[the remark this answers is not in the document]
vyasr: You're right, that's what's missing, but I don't really want to write a new class just for testing something that we don't need right now. We can come back to it whenever we get around to enabling the interchange of backends within signac, since that will require buffered versions of whatever collection type that is.

signac/synced_collections/buffers/serialized_file_buffered_collection.py
vyasr: I hadn't decided on a specific timeline. Definitely by signac 2.0, I just wanted to get this pull request merged and get some testing in the wild before switching that on master. Depending on how long we wait until 1.7, this could be either in 1.7 or in 2.0.

signac/synced_collections/backends/collection_json.py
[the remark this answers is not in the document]
vyasr: That's fine, but it will need to be copy-pasted to every single subclass of `SyncedCollection`.

signac/synced_collections/data_types/synced_list.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+
+    def __init__(self, data=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if data is None:
+            self._data = []
+        else:
+            self._validate(data)
+            if NUMPY and isinstance(data, numpy.ndarray):
+                data = data.tolist()
```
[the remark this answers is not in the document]
vyasr: For 0d numpy arrays `data.item()` and `data.tolist()` are equivalent (they return Python scalars if an appropriate conversion exists, otherwise a numpy scalar). For 1-element numpy arrays, they behave differently because `tolist()` will preserve the list while `item` will squeeze out all singleton dimensions. If a user manually constructs a `SyncedList` with a scalar or a 0d array, that will fail as expected because it is not a sequence. A nested assignment of a numpy scalar to a SyncedCollection will never trigger this path, because 0d numpy arrays are not identified as one of the base types for SyncedList (see the `_sequence_resolver` above).

#503 helps isolate this logic, hopefully making it clearer in the future.

signac/synced_collections/utils.py
```diff
@@ -0,0 +1,219 @@
...
+        str
+            The name of the type, where valid types are the keys of the dict
+            argument to the constructor. If the object's type cannot be identified,
+            will return ``None``.
+
+        """
+        # 0-d NumPy arrays must be handled the right way here.
+        if NUMPY and isinstance(obj, numpy.ndarray):
+            if obj.shape == ():
+                obj = obj.item()
```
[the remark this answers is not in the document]
vyasr: This logic is now handled more cleanly in #503.

vyasr:
@bdice here's what your benchmark looks like on the last few optimizations. I used timeit with 10 iterations to get rid of any profiling overhead and get a reasonable estimate of average performance (`%%timeit -n 10` of `load_code()` in a Jupyter notebook):

master: 9.46 ms ± 809 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
9f24 (last commit before any optimizations): 33.1 ms ± 2.08 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
adddc (define validators at class-definition time): 27.5 ms ± 1.75 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
e71942 (no os path join): 25.2 ms ± 1.77 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
48863 (disabling recursive validation): 24.7 ms ± 2.11 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
f0537 (new JSON validator): 19.2 ms ± 1.03 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
28cfe (using sets for protected keys): 18.1 ms ± 876 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

If your benchmarks on #497 still hold (I suspect that it will still help, but less since validation is now much faster) we should get pretty close to the performance of master.

vyasr:
@bdice here's an updated set of benchmarks, including #512 both with and without #514:

### Benchmarks without #514 

master: 9.46 ms ± 809 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
9f24 (last commit before any optimizations): 33.1 ms ± 2.08 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
adddc (define validators at class-definition time): 27.5 ms ± 1.75 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
e71942 (no os path join): 25.2 ms ± 1.77 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
48863 (disabling recursive validation): 24.7 ms ± 2.11 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
f0537 (new JSON validator): 19.2 ms ± 1.03 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
28cfe (using sets for protected keys): 18.1 ms ± 876 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
e5b80 (defer statepoint init): 17.5 ms ± 1.04 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
92593 (optimize SyncedCollection init and fast exit from _from_base): 15.6 ms ± 2.27 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
ca3397 (optionally skip validation on load from disk): 15.7 ms ± 907 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

### Benchmarks with #514 

master: 9.27 ms ± 3.14 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
9f24: 32.6 ms ± 3 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
adddc: 26.1 ms ± 3.26 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
e71942: 24.8 ms ± 2.94 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
48863: 23.7 ms ± 3.17 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
f0537: 19.5 ms ± 2.88 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
28cfe: 18.5 ms ± 3.1 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
e5b80: 17.6 ms ± 2.99 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
92593: 15.9 ms ± 3.21 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
ca3397: 15.3 ms ± 3.25 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

So #514 doesn't really affect performance much here. However, all of these numbers are for running the benchmark the very first time through. If I rerun the benchmark on the same project (I'm in a Jupyter notebook, so this means not resetting the kernel _and_ not calling `signac.get_project` again), then the last commit time (from PR #512) drops to ~14.5 on subsequent loads. This improvement is reproducible, and I don't observe it on previous commits, so the fast iteration path is helped somewhat by #512.

bdice (approved):
This is ready to merge. Thank you @vyasr!

PR #494, opened 2021-02-01
auto merge patches and minor versions

mikemhenry:
This will auto-merge dependbot PRs (not major bumps but minor and patch) if the tests pass.

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -1,0 +1,14 @@
+name: auto-merge
+
+on:
+  pull_request:
+
+jobs:
+  auto-merge:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ahmadnassri/action-dependabot-auto-merge@v2
+        with:
+          target: minor
+          github-token: ${{ secrets.AUTO_MERGE_DEPENDBOT }}
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -11,4 +11,5 @@ jobs:
       - uses: ahmadnassri/action-dependabot-auto-merge@v2
         with:
           target: minor
+          approve: true
           github-token: ${{ secrets.AUTO_MERGE_DEPENDABOT }}
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -11,4 +11,4 @@ jobs:
       - uses: ahmadnassri/action-dependabot-auto-merge@v2
         with:
           target: minor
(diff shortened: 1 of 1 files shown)


mikemhenry:
Reference documentation here: https://github.com/marketplace/actions/dependabot-auto-merge
If this looks good, I was thinking we merge it and see if takes care of the PRs. I'll merge them if it doesn't and we should be set for next month!

.github/workflows/auto-merge.yml
```diff
@@ -0,0 +1,14 @@
...
+
+jobs:
+  auto-merge:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ahmadnassri/action-dependabot-auto-merge@v2
+        with:
+          target: minor
+          github-token: ${{ secrets.AUTO_MERGE_DEPENDBOT }}
```
bdice, line 14: Is this a typo?
```suggestion
          github-token: ${{ secrets.AUTO_MERGE_DEPENDABOT }}
```

mikemhenry: shoot it is, but I copied and pasted the token name :facepalm: so it "works" but I don't think I can rename the token without having to remake it, I'll fix it even if I have to remake

.github/workflows/auto-merge.yml
```diff
@@ -0,0 +1,14 @@
...
+  pull_request:
+
+jobs:
+  auto-merge:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ahmadnassri/action-dependabot-auto-merge@v2
+        with:
+          target: minor
```
bdice, line 13: I'd like to make the "approve by default" option explicit.
```suggestion
          target: minor
          approve: true
```

mikemhenry: I'm in favor of making that more explicit

.github/workflows/auto-merge.yml
mikemhenry: okay fixed the name in the code and also the token name in our secrets page

bdice (approved):
Looks good enough to me! I don't know how else to test this, so I approve and we can see if it works next month.

mikemhenry:
Sounds good! I turned on the auto-merge (squash) so once it gets merged in we will see if it works retroactively. The part I was most worried about not working was the token, but it looks like it works since we get this warning:
`Warning: expected PR by "dependabot[bot]", found "mikemhenry" instead`
Which is what I would expect for this PR, so I think the token has the right scope, but we will see if it works when it has to actually perform the merge.

vyasr (approved):
Agreed, let's just test this in action. Worst case scenario we have to revert a PR at some point when it gets merged prematurely.

Issue #473, opened 2021-01-19
Improve test coverage of exception cases in job.py


vyasr:
In #472 @bdice requested improving coverage since that PR modifies various job internals. The relevant untested cases are not new, they are the same cases that are currently untested on master, so we should endeavor to address these on master first since that will be simpler to identify issues without having to deal with the complexities of the new synced collection infrastructure. Lines of code missing coverage can be see on [the Codecov page for job.py](see https://codecov.io/gh/glotzerlab/signac/src/master/signac/contrib/job.py).

bdice:
One of these failure cases to test: try to edit a state point file that can't be moved because its directory is read-only. I think that will raise an error that isn't caught here. https://github.com/glotzerlab/signac/pull/472/files#diff-7ca53886bf1aa90f8bf61c441c23a33b6a7ac67d61adad647ee141f99d5deafaR105

bdice:
Another failure case: failure to call `job.init()` because the `workspace` directory is read-only and a new job workspace directory cannot be created.

mikemhenry:
A quick google didn't find anything, but I imagine there is a way to simulate disk I/O issues with pytest (such as a read only file system)

bdice:
Most of these test cases should be possible to construct with just the `os` module.

vyasr:
Still relevant

PR #497, opened 2021-02-06
Defer statepoint instantiation, unify reset_statepoint logic

bdice:
## Description
@vyasr this is an attempt to defer instantiation of the `_StatePointDict` class. Related to https://github.com/glotzerlab/signac/pull/484#issuecomment-774398330.

~I also added an option to disable input data validation in `SyncedCollection._update`.~ I am moving the option to disable data validation into a separate PR.

## Motivation and Context
Speeds up `Project.open_job` by deferring `_StatePointDict.__init__` until `Job.statepoint` access.
This is a performance improvement for common access patterns (project iteration) for the same reasons as lazy loading, but benchmarks measuring "time to read statepoints" will see no change.

With deferred instantiation of the `_StatePointDict`, the profile of `open_job` looks identical for the current state of PR #484 and v1.6.0. As a consequence, we can conclude that the slowdown of `open_job` from v1.6.0 to the current state of PR #484 is _entirely_ due to the hierarchical access of `SyncedAttrDict.__setattr__`.

## Effect of the changes

With data validation disabled during `_StatePointDict.load`, the performance is closer to v1.6.0 but it won't match until we figure out how to improve `SyncedAttrDict.__setattr__`. Profiling a workspace of 1000 jobs, where I load the state point of every job:
- v1.6.0: 0.055 seconds
- PR #484 (currently): 0.113 seconds
- This PR: 0.081 seconds

The only other item causing slowdown relative to v1.6.0 is the call to `Job._statepoint_filename`, which appears to call `os.path.join` twice (once for the workspace, once for the state point filename) instead of once (directly computing the state point filename). Unless someone sees a way around that, I'm fine with ignoring that small impact.

### DIFF
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -380,7 +380,18 @@ class Job:
             The job's new state point.
 
         """
-        self.statepoint.reset(new_statepoint)
+        if self._statepoint_requires_init:
+            # Instantiate state point data lazily - no load is required, since
+            # we are provided with the new state point data.
+            self._statepoint = _StatePointDict(
+                jobs=[self], filename=self._statepoint_filename, data=new_statepoint
+            )
+
+            # Update the project's state point cache when loaded lazily
+            self._project._register(self.id, new_statepoint)
+            self._statepoint_requires_init = False
+        else:
+            self.statepoint.reset(new_statepoint)
 
     def update_statepoint(self, update, overwrite=False):
         """Change the state point of this job while preserving job data.
@@ -424,9 +435,12 @@ class Job:
         if not overwrite:
             for key, value in update.items():
                 if statepoint.get(key, value) != value:
-                    raise KeyError(key)
+                    raise KeyError(
+                        f"Key {key} was provided but already exists in the "
+                        "mapping with another value."
+                    )
         statepoint.update(update)
-        self.statepoint.reset(statepoint)
+        self.reset_statepoint(statepoint)
 
     @property
     def statepoint(self):
@@ -474,7 +488,7 @@ class Job:
             The new state point to be assigned.
 
         """
-        self.statepoint.reset(new_statepoint)
+        self.reset_statepoint(new_statepoint)
 
     @property
     def sp(self):
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -275,9 +275,6 @@ class Job:
         else:
             # Only an id was provided. State point will be loaded lazily.
             self._id = _id
-            self._statepoint = _StatePointDict(
-                jobs=[self], filename=self._statepoint_filename
-            )
             self._statepoint_requires_init = True
 
     def _initialize_lazy_properties(self):
@@ -383,7 +380,7 @@ class Job:
             The job's new state point.
 
         """
-        self._statepoint.reset(new_statepoint)
+        self.statepoint.reset(new_statepoint)
 
     def update_statepoint(self, update, overwrite=False):
         """Change the state point of this job while preserving job data.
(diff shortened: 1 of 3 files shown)


vyasr:
I think something along the lines of this change is fine. Adding a parameter to `_update` makes sense to me, but rather than implementing the parameter passing at the Job level, I think what would make sense is to add a parameter `trust_source` or similar to SyncedCollection that controls the argument to `_update` in `_load`. This parameter would indicate that when initializing a collection pointing to a pre-existing resource, we trust that the contents of that resource are well-formed. This will never be 100% true since, by construction, `_load_from_resource` will fail on a malformed underlying resource, but this parameter could control whether we need to validate data loaded from the resource.

The same change (whichever we decide on) should also be implemented for SyncedList.

Since this PR is meant to be an optimization, I should point out that some of the changes of `_statepoint` to `statepoint` will be less efficient with your changes. Any operation where a state point is modified before its data is loaded will be slower now because the `load` method will be called, whereas before it would not have been. This probably isn't typically performance-critical, on the other hand I don't see the minor consistency improvement of using the public property as a good enough reason to slow this down. This kind of optimization is precisely what typically justifies internal usage of an attribute rather than the public-facing property IMO.

bdice:
I discussed more with @vyasr on a call. Takeaways:
- Deferring the construction of `_statepoint` is fine, and it's okay to use only the public property `statepoint` in the `Job.init` method.
- I need to refactor `update_statepoint` and the statepoint property setter to call `reset_statepoint`.
- The `reset_statepoint` method needs to use the private `_statepoint` (and construct it, conditional on `_statepoint_requires_init`) to avoid triggering an unnecessary load.
- In `synced_collections`, rename `_update` parameter to `trust_source` which will skip validation.
- Refactor SyncedList to have a similar parameter `trust_source`.

Meanwhile, @vyasr will refactor the "attribute" accessors and make them only usable for a new child class `JSONAttrDict`.

signac/contrib/job.py
```diff
@@ -383,7 +380,18 @@ def reset_statepoint(self, new_statepoint):
             The job's new state point.
 
         """
-        self._statepoint.reset(new_statepoint)
+        if self._statepoint_requires_init:
+            # Instantiate state point data lazily - no load is required, since
+            # we are provided with the new state point data.
+            self._statepoint = _StatePointDict(
```
vyasr, line 386: Do we not have any tests that open a job by id (without a state point)? There's some surprising missing coverage here, was this one of the issues you noted on a previous PR? This seems new (most of the other ones were exception cases in `init`). I'm fine addressing in a separate PR to keep our optimization push moving, but if this is actually a mising case can you note it on #473?

vyasr (approved):
LGTM. One comment, but not something that needs to be addressed before this is merged.

signac/contrib/job.py
bdice: We definitely have tests for opening a job by id. The missing coverage is for "resetting a statepoint when opened by id, without accessing the state point or calling `job.init()`. I will add a test for that case.

bdice: I added a test and fixed a bug. Thanks for the catch @vyasr 😄

vyasr: Thanks! Good fix.

tests/test_job.py
```diff
@@ -1016,6 +1016,39 @@ def test_reset_statepoint_job(self):
         with pytest.raises(DestinationExistsError):
             src_job.reset_statepoint(dst)
 
+    def test_reset_statepoint_job_lazy_access(self):
```
bdice, line 1019:
```suggestion
    @pytest.mark.skipif(not H5PY, reason="test requires the h5py package")
    def test_reset_statepoint_job_lazy_access(self):
```

PR #499, opened 2021-02-12
Revert "Implement feature to search jobs with a sp- and doc-integrated filter "

mikemhenry:
Reverts glotzerlab/signac#332


bdice:
This temporarily resolves #500. We'll need to revisit and fix the API breakage, adding deprecations where appropriate and perhaps some new tests to cover the missing chunks of uncovered API if relevant.

bdice (approved):

PR #503, opened 2021-02-15
Feature/synced collection/numpy

### LINKED ISSUE #383: Handle NumPy extended precision values.

vyasr:
## Description

This PR isolates all numpy-specific logic into a single new `numpy_utils` module. It standardizes handling of numpy scalars, particularly with respect to 0d arrays, and ensures that each check does the bare minimum expected to reduce costs since these functions will be called frequently.

## Motivation and Context

Resolves #383 and addresses some open questions on #484. This PR will make it much easier to address #380 in the future, but that is out of scope for now.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:


signac/synced_collections/utils.py
```diff
@@ -111,18 +103,26 @@ def default(o: Any) -> Dict[str, Any]:  # noqa: D102
...
-        elif isinstance(o, numpy.number):
-            return o.item()
-    try:
-        return o._data
-    except AttributeError as e:
-        raise TypeError from e
+    # Numpy converters return the data unchanged.
+    converted_o = _convert_numpy(o)
+
+    # Numpy arrays will be converted to lists, then recursively parsed by the
```
bdice, line 109:
```suggestion
    # NumPy arrays will be converted to lists, then recursively parsed by the
```

signac/synced_collections/utils.py
```diff
@@ -111,18 +103,26 @@ def default(o: Any) -> Dict[str, Any]:  # noqa: D102
...
+    # JSON encoder, so we only have to handle the case where we have a scalar
+    # type at the bottom level that can't been converted to a Python scalar.
```
bdice, line 111:
```suggestion
    # type at the bottom level that can't be converted to a Python scalar.
```

signac/synced_collections/utils.py
```diff
@@ -111,18 +103,26 @@ def default(o: Any) -> Dict[str, Any]:  # noqa: D102
...
+    if _is_numpy_scalar(converted_o):
+        raise ValueError(
+            "Only NumPy types that can be converted to raw Python types can be"
+            "JSON-encoded. All other types, such as NumPy extended-precision types, "
+            "must be converted by the user."
```
bdice, line 116:
```suggestion
            "Only NumPy types with corresponding Python types can be JSON-encoded. "
            "All other types, such as NumPy extended precision types, must be "
            "converted by the user."
```

signac/synced_collections/validators.py
```diff
@@ -112,10 +107,8 @@ def json_format_validator(data):
         for value in data:
             json_format_validator(value)
     elif switch_type == "NUMPY":
-        if numpy.iscomplex(data).any():
-            raise TypeError(
-                "NumPy object with complex value(s) is not JSON serializable"
-            )
+        if _is_numpy_scalar(data.item()):
```
bdice, line 110: I see that the "complex" check was removed here, but I don't see a corresponding place where that was copied elsewhere. Complex values _do_ have corresponding Python types but are not JSON-encodable.

signac/synced_collections/utils.py
bdice: Exception to note: Complex values _do_ have corresponding Python types but are not JSON-encodable. See comment below.

tests/test_synced_collections/synced_collection_test.py
```diff
@@ -441,9 +521,68 @@ def set_value(sd):
...
+    def test_set_get_numpy_float_data(self, synced_collection, dtype, shape):
+        """Test setting scalar float types, which work if a raw Python analog exists."""
+        # Explicitly get an array with a shape so we can
+        value = dtype(np.random.random_sample(shape))
+
+        # If casting via item does not give a base Python type, the number
+        # should fail to set correctly.
+        raw_value = value.item() if shape is None else value.tolist()
+        test_value = value[0].item() if isinstance(raw_value, list) else raw_value
+        should_fail = isinstance(test_value, (np.number, np.bool_))
```
bdice, line 552: Rename `should_fail` to something like `has_corresponding_python_type`. It makes the reason for failure much clearer.

tests/test_synced_collections/synced_collection_test.py
```diff
@@ -441,9 +521,68 @@ def set_value(sd):
...
+    @pytest.mark.parametrize("shape", (None, (1,), (2,)))
+    def test_set_get_numpy_complex_data(self, synced_collection, dtype, shape):
+        """Test setting scalar complex types, which should always fail."""
+        # Note that the current behavior of this test is based on the fact that
+        # all backends rely on JSON-serialization (at least implicitly), even
+        # non-JSON backends. This test may have to be generalized if we add any
+        # backends that support other data, or if we want to test cases like
+        # ZarrCollection with a non-JSON codec (alternatives are supported, but
+        # not a priority to test here).
+        # Explicitly get an array with a shape so we can
```
bdice, line 575: This comment seems incomplete.

tests/test_synced_collections/synced_collection_test.py
```diff
@@ -678,3 +801,95 @@ def append_value(sl):
             list(executor.map(append_value, [synced_collection] * num_elements))
 
         assert len(synced_collection) == num_elements
+
+    @pytest.mark.skipif(not NUMPY, reason="This test requires the numpy package.")
+    @pytest.mark.parametrize("dtype", NUMPY_INT_TYPES)
+    @pytest.mark.parametrize("shape", (None, (1,), (2,)))
```
bdice, line 807: Do we need to test multi-dimensional shapes? e.g. `(3, 3)`. Since the same shapes are re-used in many tests, it would be nice to specify them once at the top as `NUMPY_SHAPES = (None, (1,), (2,), (3, 3))`.

tests/test_synced_collections/synced_collection_test.py
```diff
@@ -678,3 +801,95 @@ def append_value(sl):
...
+        # Test assignment after append.
+        with pytest.warns(NumpyConversionWarning):
+            synced_collection[-1] = value
+
+    @pytest.mark.skipif(not NUMPY, reason="This test requires the numpy package.")
+    @pytest.mark.parametrize("dtype", NUMPY_FLOAT_TYPES)
+    @pytest.mark.parametrize("shape", (None, (1,), (2,)))
+    def test_set_get_numpy_float_data(self, synced_collection, dtype, shape):
+        """Test setting scalar float types, which work if a raw Python analog exists."""
+        # Explicitly get an array with a shape so we can
```
bdice, line 830: Incomplete comment.

tests/test_synced_collections/synced_collection_test.py
```diff
@@ -441,9 +521,68 @@ def set_value(sd):
...
+            synced_collection["numpy_dtype_val"] = value
+        raw_value = value.item() if shape is None else value.tolist()
+        assert synced_collection["numpy_dtype_val"] == raw_value
+
+    @pytest.mark.skipif(not NUMPY, reason="This test requires the numpy package.")
+    @pytest.mark.parametrize("dtype", NUMPY_FLOAT_TYPES)
+    @pytest.mark.parametrize("shape", (None, (1,), (2,)))
+    def test_set_get_numpy_float_data(self, synced_collection, dtype, shape):
+        """Test setting scalar float types, which work if a raw Python analog exists."""
+        # Explicitly get an array with a shape so we can
```
bdice, line 545: Incomplete comment.

tests/test_synced_collections/synced_collection_test.py
```diff
@@ -441,9 +521,68 @@ def set_value(sd):
...
+        value = dtype(np.random.random_sample(shape))
```
bdice, line 546: Do we need to use a better random method to avoid future deprecations? I think that NumPy recommends using a Generator nowadays.

tests/test_synced_collections/synced_collection_test.py
bdice, line 575: Incomplete comment.

tests/test_synced_collections/synced_collection_test.py
bdice, line 830: Incomplete comment.

tests/test_synced_collections/synced_collection_test.py
```diff
@@ -678,3 +801,95 @@ def append_value(sl):
...
+    @pytest.mark.parametrize("shape", (None, (1,), (2,)))
+    def test_set_get_numpy_complex_data(self, synced_collection, dtype, shape):
+        """Test setting scalar complex types, which should always fail."""
+        # Note that the current behavior of this test is based on the fact that
+        # all backends rely on JSON-serialization (at least implicitly), even
+        # non-JSON backends. This test may have to be generalized if we add any
+        # backends that support other data, or if we want to test cases like
+        # ZarrCollection with a non-JSON codec (alternatives are supported, but
+        # not a priority to test here).
+        # Explicitly get an array with a shape so we can
```
bdice, line 864: Incomplete comment.

tests/test_synced_collections/test_mongodb_collection.py
```diff
@@ -29,49 +28,73 @@
...
 
 
+try:
+    import numpy
+
+    NUMPY = True
+
+    from numpy_conversion_tests import NUMPY_INT_TYPES
+
+    # BSON does not support >8-byte ints. We Programatically remove larger
```
bdice, line 38:
```suggestion
    # BSON does not support >8-byte ints. We remove larger
```

tests/test_synced_collections/test_mongodb_collection.py
```diff
@@ -29,49 +28,73 @@
     PYMONGO = False
 
 
+try:
+    import numpy
```
bdice, line 32: Let's be consistent in using `numpy` or `np`, but not both. It makes a huge difference when refactoring. At one point, I refactored all uses of `np` to `numpy`, to make the codebase consistent with its pre-existing choice of `numpy`. I noticed a few `np` uses earlier in this PR.

bdice (changes requested):
Overall looks good. I have some comments attached that should be addressed or rebutted.

There were some changes to buffering tests that I ignored and assumed are safe (code looked okay but I didn't trace the behavior to manually verify any of it).

signac/synced_collections/backends/collection_redis.py
```diff
@@ -69,6 +73,15 @@ def __deepcopy__(self, memo):
         raise TypeError("RedisCollection does not support deepcopying.")
 
 
+# TODO: Currently all the backends implicitly or explicitly rely on JSON
```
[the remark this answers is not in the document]
vyasr: I'm going to apply this to MongoDB but not Zarr. The reason is that users can specify alternate codecs for Zarr that won't be JSON-based. The same goes for Redis, we just happen to use JSON encoding for now. I'll add this to the docstring, and we can make this more flexible in the future.

signac/synced_collections/data_types/synced_attr_dict.py
```diff
@@ -186,7 +186,10 @@ def __setitem__(self, key, value):
         # should not have side effects once that backwards compatibility layer
         # is removed, so we can validate a temporary dict {key: value} and
         # directly set using those rather than looping over data.
+
         data = {key: value}
+        # TODO: May need to convert from base prior to validation to handle
```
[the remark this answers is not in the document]
vyasr: No, the implemented validators themselves handle this correctly. I think it makes sense to do this beforehand because validators should have the flexibility to respond to special types in different ways before they are coerced, if that is necessary.

signac/synced_collections/data_types/synced_list.py
```diff
@@ -10,19 +10,16 @@
...
-
+from .synced_collection import SyncedCollection, _sc_resolver
 
 # Identifies sequences, which are the base type for this class.
 _sequence_resolver = AbstractTypeResolver(
     {
         "SEQUENCE": (
             lambda obj: (isinstance(obj, Sequence) and not isinstance(obj, str))
-            or (NUMPY and (isinstance(obj, numpy.ndarray) and obj.shape != ()))
+            or _is_nonscalar_numpy_array(obj)
```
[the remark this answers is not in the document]
vyasr: I went with `_is_atleast_1d_numpy_array`.

signac/synced_collections/numpy_utils.py
```diff
@@ -0,0 +1,87 @@
+# Copyright (c) 2020 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+"""Define utilities for handling NumPy arrays."""
+
+
+class NumpyConversionWarning(UserWarning):
+    """Warning that NumPy data was converted."""
```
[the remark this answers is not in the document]
vyasr: I don't think the implicitly is necessary since the whole point of the warning is to be more open about it. Technically it's probably fine language to use, but I don't think it adds anything to the warning. The rest of the rewording is good though.

signac/synced_collections/utils.py
vyasr: I applied your changes. I'll leave this comment unresolved for the moment and we can discuss on the other thread.

signac/synced_collections/numpy_utils.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        return isinstance(data, (numpy.number, numpy.bool_)) or (
+            allow_zero_d_array and isinstance(data, numpy.ndarray) and data.shape != ()
+        )
+
+
+except ImportError:
+
+    def _convert_numpy(data):
+        """Trivial implementation if numpy is not present."""
```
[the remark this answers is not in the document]
vyasr: When first writing the module I thought doing it this way would be cleaner because there was a lot more special-purpose logic. Now that I've consolidated and simplified the logic, I agree and have made this change.

signac/synced_collections/validators.py
vyasr: Yes, and for that reason this validator should not throw a numpy-related error. Complex types will instead fall into the generic `else` clause and print `TypeError: Object of type complex is not JSON serializable`. Do you think that's insufficient or confusing? Does knowing that change what you think the error in your previous comment should say?

tests/test_synced_collections/synced_collection_test.py
vyasr: Good ideas.

tests/test_synced_collections/synced_collection_test.py
vyasr: They do, I don't anticipate `numpy.random` being removed any time in the near future though, I'd say ~3-5 years since their deprecation cycle is incredibly slow and the old `numpy.random` namespace isn't even deprecated right now. No harm in changing though.

signac/synced_collections/numpy_utils.py
```diff
@@ -0,0 +1,87 @@
...
+        -------
+        bool
+            Whether or not the input is a numpy array.
+        """
+        return isinstance(data, numpy.ndarray) and data.shape != ()
+
+    def _is_numpy_scalar(data, allow_zero_d_array=False):
+        """Check if an object is a numpy scalar.
+
+        In certain cases a 0d numpy array is an acceptable surrogate for a scalar,
```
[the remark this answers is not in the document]
vyasr: This is a very good question, and actually leads to a more complicated issue. The immediate answer to your question is that, everywhere that you expect 0d arrays to work as scalars, they do, because we first call `_convert_numpy` in all of those cases. So actually this function is never called on a 0d array, and that entire branch of this function can be removed.

The thornier issue is _why_ this works. Right now, the reason is that `_convert_numpy` is being called by the `AbstractTypeResolver` internally, which means that 0d arrays are already being converted to scalars before any of the places where this function would be called. I use resolvers code base without making this assumption, so the `_is_numpy_scalar` function is being called in places where I anticipated having to handle 0d arrays, but then found that that case never arose.

The problem is that the resolver is basically a dictionary mapping from types->strings. While it uses callables to establish the mapping the first time, afterwards it does a simple dict lookup based on the type. This means that if the first numpy array it sees is >0d, it will forever map 0d numpy arrays to the same thing as other numpy arrays (and vice versa). Currently, this works because the resolver internally converts numpy arrays, so the resulting types end up being what we want them to be.

The easiest solution for now is to simply document this behavior of the class with respect to numpy arrays. I only think it's really an issue if someone tried to use it out of the context of synced collections, in which case this behavior might be unexpected. Frankly I don't _like_ that it handles this internally, but I don't see a good alternative. If you can think of a better way to do this then I'm all for it, otherwise I'd vote to punt in the interest of getting this merged. I'd like a cleaner solution, but it's a more subtle issue that requires some thinking and I'm bandwidth-limited so I'm not going to spend the time unless an easy answer comes to me.

tcmoore3 (approved):
Great work. I have no comments beyond what Bradley already mentioned.

signac/synced_collections/numpy_utils.py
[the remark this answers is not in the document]
bdice: > So actually this function is never called on a 0d array, and that entire branch of this function can be removed.

...so can we remove it or not? I read the last 3 paragraphs 3 times but I never quite understood what you're getting at wrt how we should act here. Ultimately I'm fine with just dropping this conversation and letting it be, if that's easiest and lets us make forward progress. Resolve this however you see fit (whether with some change or no action).

signac/synced_collections/validators.py
bdice: I'm happy as long as some error is raised, regardless of the source. This can be resolved, it seems.

signac/synced_collections/utils.py
bdice: Please add some clarifying description because complex dtypes do have corresponding types in Python but are not JSON-encodable, then this can be resolved.

signac/synced_collections/numpy_utils.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+NUMPY_CONVERSION_WARNING = (
+    "Any numpy types provided will be transparently converted to the "
+    "closest base Python equivalents."
+)
+
+
+def _convert_numpy(data):
+    """Convert a numpy data types to the corresponding base data types.
```
bdice, line 27:
```suggestion
    """Convert numpy data types to the corresponding base data types.
```

signac/synced_collections/numpy_utils.py
```diff
@@ -0,0 +1,80 @@
...
+
+    0d numpy arrays and numpy scalars are converted to their corresponding
+    primitive types, while other numpy arrays are converted to lists. If data
+    not a numpy data type, this function is a no-op.
```
bdice, line 31:
```suggestion
    is not a numpy data type, this function is a no-op.
```

signac/synced_collections/numpy_utils.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    if NUMPY:
+        if isinstance(data, numpy.ndarray):
+            # tolist will return a scalar for 0d arrays, so there's no need to
+            # special-case that check. 1-element 1d arrays should remain
+            # arrays, i.e. np.array([1])->[1], not 1.
```
bdice, line 37:
```suggestion
            # arrays, i.e. np.array([1]) should become [1], not 1.
```

signac/synced_collections/numpy_utils.py
```diff
@@ -0,0 +1,80 @@
...
+            return data.item()
+    return data
+
+
+def _is_atleast_1d_numpy_array(data):
+    """Check if an object is a nonscalar numpy array.
+
+    The need to ignore 0d numpy arrays is in typical in the synced
+    collections framework, since >0d arrays are mapped to (synced) lists
+    while 0d arrays are mapped to scalars.
```
bdice, line 51: This line doesn't make sense. Do you mean:
```suggestion
    The synced collections framework must detect 0d numpy arrays, since >0d
    arrays are mapped to (synced) lists while 0d arrays are mapped to scalars.
```

signac/synced_collections/utils.py
```diff
@@ -111,18 +103,26 @@ def default(o: Any) -> Dict[str, Any]:  # noqa: D102
...
-        raise TypeError from e
+    # NumPy converters return the data unchanged.
+    converted_o = _convert_numpy(o)
+
+    # NumPy arrays will be converted to lists, then recursively parsed by the
+    # JSON encoder, so we only have to handle the case where we have a scalar
+    # type at the bottom level that can't be converted to a Python scalar.
+    if _is_numpy_scalar(converted_o):
+        raise ValueError(
+            "Only NumPy types with corresponding Python types can be JSON-encoded"
```
bdice, line 114:
```suggestion
            "Only NumPy types with corresponding Python types can be JSON-encoded. "
```

tests/test_synced_collections/test_mongodb_collection.py
```diff
@@ -29,49 +28,73 @@
...
 
 @pytest.mark.skipif(
     not PYMONGO, reason="test requires the pymongo package and mongodb server"
 )
 class TestMongoDBList(MongoDBCollectionTest, SyncedListTest):
-
     _collection_type = MongoDBList
+
+    @pytest.mark.parametrize("dtype", NUMPY_INT_TYPES)
+    @pytest.mark.parametrize("shape", (None, (1,), (2,)))
```
bdice, line 97: This test can import and use `NUMPY_SHAPES`, I think? I don't know a reason why a 2d list would fail here, so let's just test it for completeness.

bdice (approved):
A few minor change requests, otherwise good!

tests/test_synced_collections/test_mongodb_collection.py
vyasr: 👍 No reason, I just didn't import it

signac/synced_collections/utils.py
vyasr: I rewrote this error to explicitly state that the existence of a conversion is a necessary but not sufficient condition.

signac/synced_collections/numpy_utils.py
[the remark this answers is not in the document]
vyasr: Sorry, I rambled. I've resolved this in a way that makes me slightly happier. Yes, I removed this branch entirely; this function will no longer return true for 0d numpy arrays. I also added an optional preprocessor argument to the AbstractTypeResolver, allowing places that need it to specify that the input data needs to be converted in some way prior to type resolution, avoiding any implicit behavior wrt numpy.

PR #504, opened 2021-02-16
Feature/synced collection/remove attr access

vyasr:
## Description

This change removes attribute-based access from `SyncedDict` and confines it to a new mixin class `AttrDict`. This class is no longer inherited from by most backends, only a few newly created JSON classes that will use this for signac.

This PR will have some conflicts with #503, but hopefully not too many.

## Motivation and Context

This change was precipitated by some the [benchmarks](https://github.com/glotzerlab/signac/pull/484#issuecomment-774398330) of #484 implicating the deep hierarchy and a slow `__setattr__` implementation in Python with slowing down certain performance-critical operations like opening many signac jobs by id. Additionally, overriding attribute access in this manner is error-prone and makes it harder to extend these classes, so it's in our best interest to avoid this if possible. We must retain the feature in signac because of its history, but we can avoid making such promises in synced collections in the future.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/__init__.py
+++ b/signac/__init__.py
@@ -30,10 +30,8 @@ from .core.h5store import H5Store, H5StoreManager
 from .core.jsondict import flush_all as flush
 from .db import get_database
 from .diff import diff_jobs
-from .synced_collections.backends.collection_json import (
-    BufferedJSONAttrDict,
-    JSONAttrDict,
-)
+from .synced_collections.backends.collection_json import BufferedJSONAttrDict
+from .synced_collections.backends.collection_json import JSONAttrDict as JSONDict
 from .version import __version__
 
 # Alias some properties related to buffering into the signac namespace.
@@ -77,7 +75,7 @@ __all__ = [
     "get_buffer_size",
     "get_buffer_load",
     "set_buffer_size",
-    "JSONAttrDict",
+    "JSONDict",
     "H5Store",
     "H5StoreManager",
     "testing",
--- a/signac/__init__.py
+++ b/signac/__init__.py
@@ -30,16 +30,17 @@ from .core.h5store import H5Store, H5StoreManager
 from .core.jsondict import flush_all as flush
 from .db import get_database
 from .diff import diff_jobs
-from .synced_collections.backends.collection_json import BufferedJSONAttrDict
-from .synced_collections.backends.collection_json import JSONAttrDict as JSONDict
+from .synced_collections.backends.collection_json import (
+    BufferedJSONAttrDict as JSONDict,
+)
 from .version import __version__
 
 # Alias some properties related to buffering into the signac namespace.
-buffered = BufferedJSONAttrDict.buffer_backend
-is_buffered = BufferedJSONAttrDict.backend_is_buffered
-get_buffer_load = BufferedJSONAttrDict.get_current_buffer_size
-get_buffer_size = BufferedJSONAttrDict.get_buffer_capacity
-set_buffer_size = BufferedJSONAttrDict.set_buffer_capacity
+buffered = JSONDict.buffer_backend
+is_buffered = JSONDict.backend_is_buffered
+get_buffer_load = JSONDict.get_current_buffer_size
+get_buffer_size = JSONDict.get_buffer_capacity
+set_buffer_size = JSONDict.set_buffer_capacity
 
 __all__ = [
     "__version__",
--- a/signac/__init__.py
+++ b/signac/__init__.py
@@ -30,15 +30,18 @@ from .core.h5store import H5Store, H5StoreManager
 from .core.jsondict import flush_all as flush
 from .db import get_database
 from .diff import diff_jobs
-from .synced_collections.backends.collection_json import BufferedJSONDict, JSONDict
+from .synced_collections.backends.collection_json import (
+    BufferedJSONAttrDict,
+    JSONAttrDict,
+)
 from .version import __version__
 
 # Alias some properties related to buffering into the signac namespace.
-buffered = BufferedJSONDict.buffer_backend
-is_buffered = BufferedJSONDict.backend_is_buffered
-get_buffer_load = BufferedJSONDict.get_current_buffer_size
-get_buffer_size = BufferedJSONDict.get_buffer_capacity
-set_buffer_size = BufferedJSONDict.set_buffer_capacity
+buffered = BufferedJSONAttrDict.buffer_backend
+is_buffered = BufferedJSONAttrDict.backend_is_buffered
+get_buffer_load = BufferedJSONAttrDict.get_current_buffer_size
+get_buffer_size = BufferedJSONAttrDict.get_buffer_capacity
+set_buffer_size = BufferedJSONAttrDict.set_buffer_capacity
 
 __all__ = [
     "__version__",
@@ -74,7 +77,7 @@ __all__ = [
     "get_buffer_size",
     "get_buffer_load",
     "set_buffer_size",
-    "JSONDict",
+    "JSONAttrDict",
     "H5Store",
     "H5StoreManager",
     "testing",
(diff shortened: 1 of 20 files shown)


bdice (changes requested):
Looks good overall. However this separation doesn't solve the core performance issues of the custom `__setattr__` method. The profile in this image shows the same deep nested access that executes `__setattr__`, and it's only about 13% faster than my previous benchmarks of synced collections (probably explained by one or two fewer calls?).
image

signac/synced_collections/backends/collection_json.py
```diff
@@ -10,13 +10,41 @@
...
-from ..validators import json_format_validator
+from ..validators import json_format_validator, no_dot_in_key
+
+###############################################################################
+# There are many classes defined in this file. Most of the definitions are    #
+# trivial since logic is largely inherited, but the large number of classes   #
+# and the extensive docstrings can be intimidating and make the source        #
+# difficult to parse. Section headers like these are used to organize the     #
+# code to reduce this barrier.                                                #
+###############################################################################
```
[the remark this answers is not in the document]
vyasr: I personally find triple-quoted strings that aren't being used as docstrings (class, module, whatever) to look kind of ugly, and find block comments like this to be clearer headers (the triple-quoted strings just look out of place not being attached to something), but that's just my opinion 🤷 .

signac/synced_collections/backends/collection_json.py
```diff
@@ -398,3 +473,102 @@ def __init__(
...
+    """
+
+    _backend = __name__ + ".attr"  # type: ignore
+
+
+JSONAttrDict.add_validator(no_dot_in_key)
+
+
+class JSONAttrList(JSONList):
+    r"""A :class:`JSONList` whose children will be of type :class:`JSONAttrDict`."""
```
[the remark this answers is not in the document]
vyasr: Actually it does because it documents *args and **kwargs.

signac/synced_collections/validators.py
```diff
@@ -101,12 +133,8 @@ def json_format_validator(data):
...
     elif switch_type == "MAPPING":
         for key, value in data.items():
-            # Support for non-str keys will be removed in version 2.0.
-            # See issue: https://github.com/glotzerlab/signac/issues/316.
-            if not isinstance(key, (str, int, bool, type(None))):
-                raise KeyTypeError(
-                    f"Keys must be str, int, bool or None, not {type(key).__name__}"
-                )
+            if not isinstance(key, str):
+                raise KeyTypeError(f"Keys must be str, not {type(key).__name__}")
```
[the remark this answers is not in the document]
vyasr: No, because in the relevant cases where we need to support this the `_convert_key_to_str` is now handling the conversion upstream. That is what the ordering comment regarding validators in `collection_json` is referring to.

vyasr:
@bdice OK that tracks with what suggested in our meeting, which is that this probably wouldn't be sufficient to speed up the code if it really is just spending all this time in `__setattr__`.  I suspect that the mild speedup is a combination of fewer calls like you said, plus no longer needing to collect the protected keys from the class hierarchy for each type.

To me, this suggests that the main slowdown isn't the class hierarchy per se, but the large number of new attributes that are added by the new classes. In particular, there are multiple context managers created on construction and assigned for the purpose of thread-safety, buffer-safety, and sync-suspension. The reason I'm suspecting this is because `__setattr__` is defined the same way now as it was for the old `SyncedAttrDict`, so my first guess would be that it is just being called many more times. 

The easiest thing I can imagine trying for this would be adding a constructor to `AttrDict` that sets some variable `_attrdict_initialized = True` after calling `super().__init__(...)`, then using that as an extra flag in the conditional in `__setattr__` to immediately call `super().__setattr__(...)`.  If you made sure to always put `AttrDict` first in the inheritance order for subclasses, then it should run all the superclass constructors first (using object.__setattr__` instead if `AttrDict.__setattr__`). Can you pull this branch, make that change, and rerun your benchmark?

If that is still slow, then we're probably really running into the overhead of a pure Python `__setattr__` vs a C implementation; we can cross that bridge when we come to it, maybe benchmarking more closely will reveal a more obvious culprit, at that point I would suggest looking at the output of something like `line_profiler` instead of `snakeviz`. I would also advocate merging this PR regardless of the outcome of that attempted optimization, just so we can move forward on the other related tasks while working on the performance issue in parallel.

signac/synced_collections/backends/collection_json.py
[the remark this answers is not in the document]
bdice: ...what? There are no parameters listed in this docstring.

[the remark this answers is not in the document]
vyasr: Oh lol I misread where you made this comment, but I recognized the problem and fixed it in the place you actually meant. So we're good :)

signac/__init__.py
```diff
@@ -30,15 +30,16 @@
 from .core.jsondict import flush_all as flush
 from .db import get_database
 from .diff import diff_jobs
-from .synced_collections.backends.collection_json import BufferedJSONDict, JSONDict
+from .synced_collections.backends.collection_json import BufferedJSONAttrDict
+from .synced_collections.backends.collection_json import JSONAttrDict as JSONDict
```
bdice, line 34: Don't we need the buffered class here, for backwards compatibility? In my understanding, the signac 1.6.0 implementation will also buffer arbitrary `JSONDict` instances created by the user, but this would not. We would need to `import BufferedJSONAttrDict as JSONDict` if that's correct.

signac/contrib/job.py
```diff
@@ -484,20 +493,20 @@ def document(self):
...
             For more information, see
-            :class:`~signac.synced_collections.collection_json.BufferedJSONDict`.
+            :class:`~signac.synced_collections.collection_json.BufferedJSONAttrDict`.
 
         See :ref:`signac document <signac-cli-document>` for the command line equivalent.
 
         Returns
         -------
-        :class:`~signac.JSONDict`
+        :class:`~signac.BufferedJSONAttrDict`
```
bdice, line 502: If we import the `BufferedJSONAttrDict as JSONDict` above, then we should revert to the old name `signac.JSONDict`.

signac/contrib/job.py
```diff
@@ -528,7 +537,7 @@ def doc(self):
 
         Returns
         -------
-        :class:`~signac.synced_collections.backends.collection_json.BufferedJSONDict`
+        :class:`~signac.synced_collections.backends.collection_json.BufferedJSONAttrDict`
```
bdice, line 540: If we import the `BufferedJSONAttrDict as JSONDict` above, then we should revert to the old name `signac.JSONDict`.

signac/synced_collections/data_types/attr_dict.py
```diff
@@ -0,0 +1,61 @@
...
+__delattr__.  While we do not want to offer this API generally for all
+SyncedDict objects, some applications may want to add this feature, so this
+simple mixin can be combined via inheritance without causing much difficulty.
+"""
+
+from typing import Tuple
+
+
+class AttrDict:
+    r"""A class that redirects attribute access methods to __getitem__.
```
bdice, line 16:
```suggestion
    """A class that redirects attribute access methods to __getitem__.
```

signac/synced_collections/validators.py
```diff
@@ -69,6 +70,37 @@ def no_dot_in_key(data):
...
+
+    Parameters
+    ----------
+    data
+        Data to validate.
+
+    Raises
+    ------
+    KeyTypeError
+        If key type is not not a string.
```
bdice, line 86:
```suggestion
        If key type is not a string.
```

bdice (approved):
Approving conditional on a few last fixes.

signac/__init__.py
vyasr: ... yes. Annoyingly :P

signac/synced_collections/validators.py
vyasr: ...but what if the key is not not a string???

PR #507, opened 2021-02-19
Remove add_validators and specify validators in class definition

vyasr:
## Description

Remove `add_validator` and instead force classes to define validators at class definition time via clearly defined class variables.

## Motivation and Context

This change prevents issues that could arise if different applications using synced collections try to register different validators to the same SyncedCollection class, because there is no way to namespace these if they can modify the class directly. The change forces users to subclass the desired class to add the new validators.

This change also allows reliable registration of validators at class-definition time, rather than collection of validators from the MRO every time the property is requested. This is a minor optimization. More importantly, the addition of the resulting `_all_validators` hook makes it easy for subclasses to override the set list of validators entirely, which makes it much easier to cleanly optimize the validation process where appropriate.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/synced_collections/data_types/synced_collection.py
+++ b/signac/synced_collections/data_types/synced_collection.py
@@ -7,7 +7,7 @@ from collections import defaultdict
 from collections.abc import Collection
 from inspect import isabstract
 from threading import RLock
-from typing import Any, Callable, DefaultDict, List
+from typing import Any, DefaultDict, List
 
 from ..numpy_utils import _convert_numpy
 from ..utils import AbstractTypeResolver, _CounterContext, _NullContext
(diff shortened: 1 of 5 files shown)


signac/synced_collections/data_types/synced_collection.py
```diff
@@ -220,18 +272,8 @@ def disable_multithreading(cls):
...
-        """
-        validators = []
-        # Classes inherit the validators of their parent classes.
-        for base_cls in type(self).__mro__:
-            if hasattr(base_cls, "_validators"):
-                validators.extend(
-                    [v for v in base_cls._validators if v not in validators]
-                )
-        return validators
+        """List[Callable]: The validators that will be applied."""
```
bdice, line 275: Tuple, not list.

bdice (approved):
All good. One minor fix.

PR #510, opened 2021-02-19
Feature/synced collection/optimize protected key lookup

vyasr:
## Description

**This PR should be merged after #508.**

It changes the `_PROTECTED_KEYS` attribute of AttrDict subclasses into sets to speed up checks for when those are being accessed. It also changes the order of checks in __setattr__ assuming that dunder attributes are set far less frequently than protected keys.

## Motivation and Context

Optimization for synced collections.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/synced_collections/backends/collection_json.py
+++ b/signac/synced_collections/backends/collection_json.py
@@ -9,9 +9,8 @@ import os
 import uuid
 import warnings
 from collections.abc import Mapping, Sequence
-from typing import Callable
+from typing import Callable, FrozenSet
 from typing import Sequence as Sequence_t
-from typing import Tuple
 
 from .. import SyncedCollection, SyncedDict, SyncedList
 from ..buffers.memory_buffered_collection import SharedMemoryFileBufferedCollection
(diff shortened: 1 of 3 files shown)


signac/synced_collections/data_types/attr_dict.py
```diff
@@ -49,13 +49,17 @@ def __setattr__(self, key, value):
...
+        # much more common than setting dunder attributes.
+        if key in self._PROTECTED_KEYS or key.startswith("__"):
             super().__setattr__(key, value)
         else:
             self.__setitem__(key, value)
 
     def __delattr__(self, key):
-        if key.startswith("__") or key in self._PROTECTED_KEYS:
+        # The order of these checks assumes that setting protected keys will be
+        # much more common than setting dunder attributes.
```
bdice, line 61: This whole comment is kind of irrelevant because we don't expect deletion of protected keys or dunder attributes to be allowed... but whatever.
```suggestion
        # The order of these checks assumes that deleting protected keys will be
        # much more common than deleting dunder attributes.
```

bdice (approved):

signac/synced_collections/backends/collection_json.py
```diff
@@ -278,24 +277,26 @@ def _lock_id(self):
...
+        "_load_and_save",
+        "_suspend_sync",
+        "_supports_threading",
+        "_LoadSaveType",
+        "registry",
+        # These keys are specific to the JSON backend.
+        "_filename",
+        "_write_concern",
+    ]
 )
```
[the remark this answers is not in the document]
vyasr: Good call on the mutability. I don't think the benchmarks is super relevant since this will only happen one time ever (when the class is defined, rather than when instances are created), but the change is a good one for clarity.

signac/synced_collections/data_types/attr_dict.py
vyasr: Yeah I mean it's possible that some internal method at some point will want do `del self._data` or something temporarily 🤷 .

PR #511, opened 2021-02-19
Don't use os.path.join where not needed.

vyasr:
## Description

Rather than using `os.path.join`, directly concatenating using `os.sep` in `job.workspace` and `job._statepoint_filename`. All the inputs are completely controlled by signac, so we should not need any of the fancier handling done by `os.path.join` to avoid duplicate slashes, handle multiple absolute paths, etc. For the straightforward cases we're dealing with, the simple string concatenation with `os.sep` should be faster without loss of generality.

## Motivation and Context

Optimization for synced collections.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -346,7 +346,7 @@ class Job:
         """
         if self._wd is None:
             # We can rely on the project workspace to be well-formed, so just
-            # use string-concatenation with os.sep.
+            # use string-concatenation with os.sep instead of os.path.join for speed.
             self._wd = self._project.workspace() + os.sep + self.id
         return self._wd
 
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -345,13 +345,15 @@ class Job:
 
         """
         if self._wd is None:
-            self._wd = os.path.join(self._project.workspace(), self.id)
+            # We can rely on the project workspace to be well-formed, so just
+            # use string-concatenation with os.sep.
+            self._wd = self._project.workspace() + os.sep + self.id
         return self._wd
 
     @property
     def _statepoint_filename(self):
         """Get the path of the state point file for this job."""
-        return os.path.join(self.workspace(), self.FN_MANIFEST)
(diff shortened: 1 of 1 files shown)


signac/contrib/job.py
```diff
@@ -345,13 +345,15 @@ def workspace(self):
 
         """
         if self._wd is None:
-            self._wd = os.path.join(self._project.workspace(), self.id)
+            # We can rely on the project workspace to be well-formed, so just
+            # use string-concatenation with os.sep.
```
bdice, line 349:
```suggestion
            # use string-concatenation with os.sep instead of os.path.join for speed.
```

bdice (approved):
LGTM, but the comment is not very helpful unless you mention that it's meant to be an optimization.

mikemhenry:
Does this speed things up by not resolving absolute paths and sym links?
What does the speedup look like?

Also while we are at it, have we looked at seeing if [pathlib](https://docs.python.org/3/library/pathlib.html) would be a good replacement for all the `os` stuff we use?

bdice:
> 
> 
> Does this speed things up by not resolving absolute paths and sym links?
> What does the speedup look like?
> 
> Also while we are at it, have we looked at seeing if [pathlib](https://docs.python.org/3/library/pathlib.html) would be a good replacement for all the `os` stuff we use?

@mikemhenry I have looked at `pathlib` before and benchmarked it (informally, no numbers handy). `pathlib` creates Python objects to represent paths, and they're slower to instantiate and manipulate than strings with `os.path`. We need to use the lowest-level features available to us for speed in signac.

This change from `os.path.join` to manual construction with `os.sep` will not change to absolute paths or resolve symlinks, it only skips over some path-validation steps (@vyasr mentions duplicate slashes, handling multiple absolute paths). This is fine for these places in signac because we know the input paths from the project/workspace are valid and don't need those checks for safety.

PR #513, opened 2021-02-20
Feature/synced collection/optimize

vyasr:
## Description

This PR applies a number of minor optimizations, in particular a faster exit from the `_from_base` method and a more efficient `SyncedCollection.__init__`.

## Motivation and Context

Optimizing synced collections.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/synced_collections/numpy_utils.py
+++ b/signac/synced_collections/numpy_utils.py
@@ -32,7 +32,10 @@ def _convert_numpy(data):
     primitive types, while other numpy arrays are converted to lists. If ``data``
     is not a numpy data type, this function is a no-op.
     """
-    if NUMPY:
+    # Initially performing one isinstance check is faster since most of the
+    # time the inputs are not numpy arrays, preventing additional function
+    # calls inside the if statement.
+    if NUMPY and isinstance(data, (numpy.ndarray, numpy.number, numpy.bool_)):
         if isinstance(data, numpy.ndarray):
             # tolist will return a scalar for 0d arrays, so there's no need to
             # special-case that check. 1-element 1d arrays should remain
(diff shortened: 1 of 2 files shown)


signac/synced_collections/numpy_utils.py
```diff
@@ -32,7 +32,10 @@ def _convert_numpy(data):
     primitive types, while other numpy arrays are converted to lists. If ``data``
     is not a numpy data type, this function is a no-op.
     """
-    if NUMPY:
+    # Initially performing one isinstance check is faster since most of the
+    # time the inputs are not numpy arrays, preventing additional function
+    # calls inside the if statement.
+    if NUMPY and isinstance(data, (numpy.ndarray, numpy.number, numpy.bool_)):
```
bdice, line 38: Can the second "elif" just be an "else", by elimination? Otherwise the "else" would be unreachable.

It's too far below this change for me to make a suggestion.

bdice: Also should we support _all_ array scalars? We could use `numpy.generic` instead of `(numpy.number, numpy.bool_)`? You can have strings and dicts and other supported (JSON-encodable) types that are simply wrapped as array scalars. https://numpy.org/doc/stable/reference/arrays.scalars.html

bdice (commented):
Questions in comment.

signac/synced_collections/numpy_utils.py
vyasr: For your first question, yes, good call.

For your second question, I'm not opposed to doing that, but I'd rather punt on it for now. I intentionally added `np.bool_` here but not `flexible` or `object_` since I figured that we'd need to be much more thorough about testing those cases and deciding where errors should be thrown. We can definitely do this, I just don't want to think about it as part of #484.

bdice: Sounds good.

bdice (approved):
Merging.

PR #514, opened 2021-02-20
Treat empty filters {} the same as None.

bdice:
## Description
I have been running benchmarks for synced collections (#484) and noticed that my project was loading state point data even when not requested.

I traced this down to the recent changes in #332 that changed the default `filter` behavior from `None` to `{}`. This prevented some of the "fast path" checks from triggering, and caused a full index for the project to be constructed.

## Motivation and Context
Bug fix.

I don't know how to test this effectively, because it only affects internal APIs and performance, not correctness.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -13,7 +13,7 @@ next
 Added
 +++++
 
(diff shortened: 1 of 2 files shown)


vyasr:
I'm fine with not testing this for now.I didn't notice that the innocuous-looking call to `project._find_job_ids` in `JobsCursor.__iter__` that resulting from `filter=None` would be the difference between slow and fast iteration, so in lieu of specific testing I'm going to add some documentation of this behavior in a few places so that it's easier for developers to be aware.

vyasr (approved):

PR #515, opened 2021-02-20
Optimize path joins with os.sep.join(...).

bdice:
## Description
Extends on work started in #511. In a few performance-critical locations that use internal paths (not user-provided paths), we can sacrifice the safety benefits of `os.path.join(*paths)` in favor of the speedier `os.sep.join(paths)`.

This yields about a 20-25% speedup when iterating over a project without accessing job state points.

We can only apply this optimization in places that deal with _internal paths_, like "project_workspace / job_id" but not like `job.fn(some_user_path)`. User paths must be accessed through `os.path.join`, which includes the project workspace (which is configurable). Regardless, we are able to safely handle the most important cases (the job workspace and the job state point filename) that are called O(N) times.

## Benchmarks

Below I have benchmarks of the following script, which just iterates over the project without accessing any job properties. There are 1000 jobs in the workspace.

```python
import signac

def no_load_data():
    data = [job for job in project]

no_load_data()
```

### Before

The call to `os.path.join` (displayed as `posixpath.py:71(join)`) takes a chunk of time, which is eliminated in the "After".
image

### After
The bulk of the iteration is `stat` calls, which cannot be further optimized. 👍 
image

## Motivation and Context
Seeking some final obvious optimizations for synced collections (#484).

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


kidrahahjo (approved):
This is an interesting find, the changes looks good!

vyasr (approved):

PR #530, opened 2021-03-12
Avoid deadlock in SyncedCollection

joaander:
## Description

Add locks needed to avoid deadlocks in `SyncedCollection`

## Motivation and Context

This MWE deadlocks because `SycedCollections._locks[lock_id]` gets overwritten while it is being used:

```
import json
import tempfile
from concurrent.futures import ThreadPoolExecutor, ProcessPoolExecutor
from signac.synced_collections.backends.collection_json import JSONAttrDict
def update_status(data):
    s, filename = data
    d = JSONAttrDict(filename=filename)
    d.update(dict(a=1, b=s-1))
    return False
if __name__ == "__main__":
    N = 1
    n_tasks = 20
    with tempfile.TemporaryDirectory() as d:
        for s in range(N):
            with open(f'{d}/{s}.json', 'w') as f:
                json.dump(dict(a=s, b=0), f)
        tasks = []
        for s in range(N):
            for t in range(n_tasks):
                tasks.append((s, f'{d}/{s}.json'))
        with ThreadPoolExecutor() as e:
            e.map(update_status, tasks)
        print(JSONAttrDict._locks)
```

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/synced_collections/data_types/synced_collection.py
+++ b/signac/synced_collections/data_types/synced_collection.py
@@ -188,7 +188,9 @@ class SyncedCollection(Collection):
             self._load_and_save = self._LoadSaveType(self)
 
         if self._supports_threading:
-            self._locks[self._lock_id] = RLock()
+            with self._cls_lock:
+                if self._lock_id not in self._locks:
+                    self._locks[self._lock_id] = RLock()
 
     @classmethod
     def _register_validators(cls):
@@ -235,6 +237,7 @@ class SyncedCollection(Collection):
 
         # Monkey-patch subclasses that support locking.
         if cls._supports_threading:
+            cls._cls_lock = RLock()
             cls._locks = {}
             cls.enable_multithreading()
         else:


bdice (approved):
Looks good to me. Thanks again for your help, @joaander. If you can think of a test that would _fail_ (but not deadlock), that would be a good addition. I don't know how I would construct such a case unless there is a way to force a timeout.

joaander:
> Looks good to me. Thanks again for your help, @joaander. If you can think of a test that would _fail_ (but not deadlock), that would be a good addition. I don't know how I would construct such a case unless there is a way to force a timeout.

I tried to get failing behavior in my first attempt on the way to this MWE but wan't successful. I had tried to trigger conflicting writes to the file.

If inspection of private variables is allowed in the test, a test that would fail before this change would construct two `JSONAttrDict`s with the same file and assert that the identify of the lock `JSONAttrDict._locks[filename]` stays the same.

cbkerr (approved):
With help from @bdice, I was able to reproduce the problem and fix even though I didn't experience a deadlock by printing `threading.get_ident()` and "done" inside the `update_status` function.

vyasr (approved):

vyasr:
In the interest of getting things merged I'll look into constructing a test later.

signac/synced_collections/data_types/synced_collection.py
```diff
@@ -188,7 +188,9 @@ def __init__(self, parent=None, *args, **kwargs):
             self._load_and_save = self._LoadSaveType(self)
 
         if self._supports_threading:
-            self._locks[self._lock_id] = RLock()
+            with self._cls_lock:
+                if self._lock_id not in self._locks:
+                    self._locks[self._lock_id] = RLock()
```
csadorf, line 193: @vyasr I'm wondering whether `defaultdict(RLock)`  might be thread-safe?
https://stackoverflow.com/a/17682555

vyasr: In my opinion it is usually inadvisable to rely on the thread-safety of Python built-ins. Their thread-safety is an implementation detail, and it's also often quite difficult to predict exactly how a given line of Python code will translate to Python bytecode (and Python is only thread-safe at the level of a single bytecode instruction). For example, one thread executing line 193 could create an RLock, but then be preempted by another thread before actually assigning it. We're better off adding a lock to guarantee safety, especially since it's a relatively cheap operation that only occurs on construction.

csadorf: Good point!

PR #532, opened 2021-03-25
Add configuration for stale bot

csadorf:
## Description

Adds configuration for the [stale bot](https://github.com/apps/stale).

I increased the default stale time from 60 days to 90 days to be a bit more conservative.
We can consider to make it more aggressive after some testing time.

Issues labeled 'pinned', 'security', or 'blocked' are ignored.

## Motivation and Context

There are many, many open issues that have become stale that clutter up the issue list and also waste the maintainers time. Using a bot will help de-clutter and make sure that issues that appear to not be as important as originally thought get closed over time.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -8,7 +8,7 @@ exemptLabels:
   - security
   - blocked
 # Label to use when marking an issue as stale
-staleLabel: wontfix
+staleLabel: stale
 # Comment to post when marking an issue as stale. Set to `false` to disable
 markComment: >
   This issue has been automatically marked as stale because it has not had
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,18 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 90
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+  - blocked
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: false
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,7 +1,7 @@
 # Number of days of inactivity before an issue becomes stale
-daysUntilStale: 90
+daysUntilStale: 60
 # Number of days of inactivity before a stale issue is closed
-daysUntilClose: 7
+daysUntilClose: 14
 # Issues with these labels will never be considered stale
 exemptLabels:
   - pinned


csadorf:
Once this PR is approved, I will create similar configuration for -flow, -dashboard, and -docs.

.github/stale.yml
```diff
@@ -0,0 +1,18 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 90
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
```
bdice, line 4: I think we want the first "ping" (becoming stale) to be earlier, and the final closure to be later.
```suggestion
daysUntilStale: 30
# Number of days of inactivity before a stale issue is closed
daysUntilClose: 60
```

.github/stale.yml
```diff
@@ -0,0 +1,18 @@
...
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+  - blocked
+# Label to use when marking an issue as stale
+staleLabel: wontfix
```
bdice, line 11: I don't think "wontfix" means the same thing as "stale" -- "wontfix" sounds like a refusal to ever consider addressing the issue, whereas "stale" just means it hasn't been prioritized.
```suggestion
staleLabel: stale
```

bdice:
> 
> 
> Once this PR is approved, I will create similar configuration for -flow, -dashboard, and -docs.

Could you also add signac-examples? Thank you so much, this is a very helpful maintenance task. 🚀

.github/stale.yml
csadorf: I think 30 days until stay is way too short for the current state and respecting @vyasr wish to be conservative about this. And then 60 days until actually closing it is way too long. How about: 60-14 ?

csadorf: I think 14 days should be more than enough time for anybody involved to respond to this and "prevent" the close. Also keep in mind that we can re-open an issue if needed, it's not like they get deleted.

.github/stale.yml
csadorf: Yeah, true. That also irked me a bit when I copied the config. I think having a dedicated 'stale' label is much better, agreed.

mikemhenry:
Thanks! This is great!

mikemhenry (approved):

.github/stale.yml
vyasr: I agree with @csadorf here, the time between when an issue is marked stale and when it's closed should be relatively short. 60-14 seems like a reasonable compromise. A total of 2.5 months of inactivity is a reasonable limit for when to close.

I agree that we can re-open issues in theory, but my concern has always been that this happens infrequently in practice. Github's search defaults to open issues, and unless people know to search closed ones (or have enough institutional knowledge to remember old closed issues) it's hard to remember. The overhead of open issues is definitely big enough to make this PR worthwhile, just pointing out why I've always been a bit hesitant to rely on the fact that closed issues don't vanish.

csadorf: > I agree with @csadorf here, the time between when an issue is marked stale and when it's closed should be relatively short. 60-14 seems like a reasonable compromise. A total of 2.5 months of inactivity is a reasonable limit for when to close.

Ok, would you prefer to have a prolonged discussion about this then? Otherwise I'd request you approve the PR.

> I agree that we can re-open issues in theory, but my concern has always been that this happens infrequently in practice. Github's search defaults to open issues, and unless people know to search closed ones (or have enough institutional knowledge to remember old closed issues) it's hard to remember. The overhead of open issues is definitely big enough to make this PR worthwhile, just pointing out why I've always been a bit hesitant to rely on the fact that closed issues don't vanish.

I don't think this is a problem. Either a particular issue crops up again and is important enough that there is either some institutional knowledge about it or that people go through the "effort" of also searching closed issues [1] or it is not that important and there simply is a little bit of duplication.

[1] I honestly don't understand that. Whenever I don't find what I'm looking for among the open issues, I of course also search the closed issues to check whether the issue has come up before or has already been resolved. I've certainly encountered auto-closed (stale) issues in this way on which I then simply commented and asked for re-opening.

bdice (approved):

.github/stale.yml
vyasr: Then you're better than me :) That wasn't true for me until maybe a couple of years ago, certainly not during the first few years of my dev life. Unless I knew that an issue previously existed and was closed, I wouldn't have thought to search through closed issues even though I knew it was a possibility. And I certainly see enough duplicate issues (usually of closed issues, but also occasionally of open ones) that this is definitely not a problem that's restricted to me.

vyasr, line 4:
```suggestion
daysUntilStale: 60
# Number of days of inactivity before a stale issue is closed
daysUntilClose: 14
```

vyasr (approved):

PR #543, opened 2021-04-27
Skip pre-checks in favor of pre-commit CI.

bdice:
## Description
I recently enabled pre-commit CI. This can replace the pre-checks currently being run on CircleCI.

## Motivation and Context
Removes redundancy and potentially speeds up CI.

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -268,31 +268,15 @@ workflows:
   version: 2
   test:
     jobs:
-      - pre-checks
-      - linux-python-39:
-          requires:
-            - pre-checks
-      - linux-python-39-minimal:
-          requires:
-            - pre-checks
-      - linux-python-38:
-          requires:
-            - pre-checks
-      - linux-python-37:
-          requires:
-            - pre-checks
-      - linux-python-36-oldest:
-          requires:
-            - pre-checks
-      - linux-pypy-37:
-          requires:
-            - pre-checks
-      - windows-python-38:
-          requires:
-            - pre-checks
-      - macos-python-3:
-          requires:
-            - pre-checks
+#     - pre-checks
+      - linux-python-39
+      - linux-python-39-minimal
+      - linux-python-38
+      - linux-python-37
+      - linux-python-36-oldest
+      - linux-pypy-37
+      - windows-python-38
+      - macos-python-3
       - check-metadata:
           filters:
             branches:


tommy-waltmann:
Does the other PR use github actions to do pre-commit CI? If CircleCI is handling the pre-commit, we may want to require the pre-commit checks pass before running tests to conserve CI time.

bdice:
> 
> 
> Does the other PR use github actions to do pre-commit CI? If CircleCI is handling the pre-commit, we may want to require the pre-commit checks pass before running tests to conserve CI time.

Pre-commit CI was added as an application, so there is no corresponding PR. You can see it here (may require some kind of permissions, I'm not sure): https://github.com/glotzerlab/signac/settings/installations

The Pre-commit CI service is better than running the pre-commit hooks on CircleCI because it can also update the PR to fix any formatting issues and also keeps the pre-commit configuration updated.

Your comment about conserving CI time is possible, but I think that on average we will see a reduction in time spent waiting for CI if we remove the first step (style checks) that was a dependency for all other steps. The time I would measure is "time from pushing a commit to all checks passing." Concurrency limits on CircleCI (4 runners for the glotzerlab organization) may be a limiting factor some of the time, e.g. if pushes occur to signac, signac-flow, and freud at the same time, but I believe the majority of commits to the repository do pass the style checks so parallelizing that step is helpful. Also, if pre-commit CI catches any style problems, it will push to the branch, which will cancel any queued tests on CircleCI for the "unstylish" commit and queue up tests for the fixed commit.

.circleci/config.yml
```diff
@@ -268,31 +268,15 @@ workflows:
...
-      - linux-pypy-37:
-          requires:
-            - pre-checks
-      - windows-python-38:
-          requires:
-            - pre-checks
-      - macos-python-3:
-          requires:
-            - pre-checks
+#     - pre-checks
```
csadorf, line 271: Why do we keep this comment?

csadorf (approved):
One question, but LGTM overall.

joaander:
> Your comment about conserving CI time is possible, but I think that on average we will see a reduction in time spent waiting for CI if we remove the first step (style checks) that was a dependency for all other steps. The time I would measure is "time from pushing a commit to all checks passing." Concurrency limits on CircleCI (4 runners for the glotzerlab organization) may be a limiting factor some of the time, e.g. if pushes occur to signac, signac-flow, and freud at the same time, but I believe the majority of commits to the repository do pass the style checks so parallelizing that step is helpful. Also, if pre-commit CI catches any style problems, it will push to the branch, which will cancel any queued tests on CircleCI for the "unstylish" commit and queue up tests for the fixed commit.

CircleCI still limits 4 running jobs? If this is every a limit, we should switch to GitHub Actions. Concurrency on GitHub Actions is 60 (though only 5 on mac) for our free Team account: https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#usage-limits

tommy-waltmann (approved):

.circleci/config.yml
bdice: I was planning to keep this temporarily -- I don't want to remove the `pre-checks` job configuration just yet, until we keep pre-commit CI enabled for a little bit longer.

PR #563, opened 2021-06-03
Fix incorrect docstring, resolve #506.

### LINKED ISSUE #506: Incorrect docstring in Job.update_statepoint
The docstring modified by @cbkerr in #444 has a small error / confusing wording. Here, the `Job.update_statepoint` method claims that setting `overwrite=True` makes the method `update_statepoint` act like the method `reset_statepoint`. This is not quite true, and nearly caused me to re-implement the behavior incorrectly while refactoring #497.
https://github.com/glotzerlab/signac/blob/9bb8887485f9dca6d5c012e8d20ad6303eb97334/signac/contrib/job.py#L407-L408

For example, see this minimal reproduction:
```python
import signac
project = signac.init_project("UpdateStatepointDocstringBug")
job = project.open_job({"a": 1})
print(job.sp)
job.update_statepoint({"b": 2}, overwrite=True)   # Here the overwrite behavior doesn't matter because the key does not exist yet
print(job.sp)
job.update_statepoint({"b": 3})
print(job.sp)
```

Output:
```python
{'a': 1}
{'a': 1, 'b': 2}
  File "problem.py", line 7, in <module>
    job.update_statepoint({"b": 3})
  File "/home/bdice/code/signac/signac/contrib/job.py", line 414, in update_statepoint
    raise KeyError(key)
KeyError: 'b'
```

Notice that when `overwrite=True`, the `Job.update_statepoint` method acts like `dict.update` in its behavior of overwriting values: https://docs.python.org/3/library/stdtypes.html#dict.update

I suggest rewriting the docstring:

```python
        Parameters
        ----------
        update : dict
            A mapping used for the state point update.
        overwrite : bool, optional
            If False, an error will be raised if the update modifies the
            values of existing keys in the state point. If True, any existing
            keys will be overwritten in the same way as :meth:`dict.update`.
            Use with caution! (Default value = False).
```

**This also affects `Project.update_statepoint`.**

bdice:
## Description
See #506 for issue description. This alters a docstring whose described behavior doesn't match the implementation.

## Motivation and Context
Resolves #506.

## Types of Changes
- [x] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -26,6 +26,11 @@ Deprecated
 
  - ``doc_filter`` arguments, which are replaced by namespaced filters. Due to their long history, ``doc_filter`` arguments will still be accepted in signac 2.0 and will only be removed in 3.0.
 
+Fixed
++++++
+
+ - Corrected docstrings for ``Job.update_statepoint`` and ``Project.update_statepoint`` (#506, #563).
+
 
 [1.6.0] -- 2020-01-24
 ---------------------
--- a/signac/contrib/job.py
+++ b/signac/contrib/job.py
@@ -410,25 +410,25 @@ class Job:
 
         .. warning::
 
-            While appending to a job's state point is generally safe,
-            modifying existing parameters may lead to data
-            inconsistency. Use the overwrite argument with caution!
+            While appending to a job's state point is generally safe, modifying
+            existing parameters may lead to data inconsistency. Use the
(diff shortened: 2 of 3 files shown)


csadorf (approved):
Thx for fixing!

cbkerr (approved):
Thanks 🥺

PR #582, opened 2021-07-29
Fix warning in synced collections tests with complex value.

bdice:
## Description
This is a minor fix for an issue in a test of synced collections with data types that cannot be serialized.

The current test uses `numpy.complex`, which is an alias for the built-in type `complex`. This test should instead use `numpy.complex128`, which is the scalar type that was intended for this test of unsupported NumPy types.

## Motivation and Context
Removes a warning from the output of `pytest`.

## Types of Changes
- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/tests/test_synced_collections/test_validators.py
+++ b/tests/test_synced_collections/test_validators.py
@@ -105,7 +105,7 @@ class TestJSONFormatValidator:
     @pytest.mark.skipif(not NUMPY, reason="test requires the numpy package")
     def test_numpy_invalid_data(self):
         # complex data
-        data = numpy.complex(1 + 2j)
+        data = numpy.complex128(1 + 2j)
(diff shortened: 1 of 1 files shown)


csadorf (approved):
Does that mean the built-in complex type is also not supported? Should we maybe duplicate this test to capture that behavior as well?

bdice:
> 
> 
> Does that mean the built-in complex type is also not supported? Should we maybe duplicate this test to capture that behavior as well?

The built-in type `complex` is tested here: https://github.com/glotzerlab/signac/blob/1f3809e4a23c192d9ad19a4b4377b1d7e02b5c86/tests/test_synced_collections/test_validators.py#L93

PR #586, opened 2021-08-02
Remove doc_filter argument

bdice:
## Description
Currently this PR only hides the deprecation warning for `doc_filter` in pytest output.

This PR needs to be discussed in a future developer meeting. I don't fully understand the intention of this comment in the source code:

```python
# TODO: In signac 2.0, remove all docstrings for doc_filter parameters. The
# doc_filter parameters will only be preserved for backwards compatibility but
# not advertised as part of the API in signac 2.0.
```

## Motivation and Context
signac 2.0 cleanup.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:


vyasr:
My recommendation would be to remove `doc_filter` arguments in all methods and then attempt pop them from `**kwargs` using `doc_filter = kwargs.pop('doc_filter', None)`. You can then follow [the idiomatic approach I wrote in `_find_job_ids`](https://github.com/glotzerlab/signac/blob/c6e059e5c6f90c72e8317c70f4b6025a5d4dd12f/signac/contrib/project.py#L870):
```
if doc_filter is not None:
    warnings.warn(DOC_FILTER_WARNING, DeprecationWarning)
    filter.update(parse_filter(_add_prefix("doc.", doc_filter)))
```
For functions that don't currently accept arbitrary kwargs you may want to raise a TypeError to maintain current behavior for other unexpected arguments (e.g. `doc_filter = kwargs.pop('doc_filter', None); if len(kwargs): raise TypeError(...)`)

The goal is to not break existing code paths that use `doc_filter` (which I know is quite prevalent in even relatively recent signac scripts) but to completely hide it from new users.

bdice:
@vyasr Using `**kwargs` will still break a lot of code. In signac-flow, we call `project.find_jobs(filter, doc_filter)` using positional arguments. I don't see a way to remove the argument name from the documentation/signature and still preserve the desired functionality. We can't use `*args` and `**kwargs` either, because of the default value for `filter=None`.

vyasr:
The decision to not deprecate until 3.0 is based on our knowledge of common usage patterns. `Project._find_job_ids` is internal since `find_job_ids` is marked for removal in 2.0, and we have no reason to think that users create `JobsCursor` instances manually, so `find_jobs` is the only place that I'm aware of that maintaining backwards compatibility is important.  Given how prevalent the doc filter is in our earlier examples and in many examples of user scripts that I've seen, I think we should be willing to do some slightly ugly things internally to keep that backwards compatibility. I don't follow why `filter` having a default argument makes it impossible to use `*args` and `**kwargs`, although I agree that it makes processing (slightly) uglier. Is there a combination that I'm missing here?

```
# test.py
class TestDocFilter:
    def find_jobs(self, filter=None, *args, **kwargs):
        doc_filter = next(iter(args), None) or kwargs.pop('doc_filter', None)
        print(filter)
        print(doc_filter)
        print()


tester = TestDocFilter()

f = {'foo': 'a'}
df = {'bar': 'b'}

tester.find_jobs()
tester.find_jobs(f)
tester.find_jobs(f, df)
tester.find_jobs(filter=f, doc_filter=df)
tester.find_jobs(doc_filter=df)
```

```
(main) vyasr@vyasr-mlt ~ % python test.py
None
None

{'foo': 'a'}
None

{'foo': 'a'}
{'bar': 'b'}

{'foo': 'a'}
{'bar': 'b'}

None
{'bar': 'b'}
```

bdice:
@vyasr I made a mistake in my recollection of how `*args` worked. I thought it would cause a `SyntaxError: positional argument follows keyword argument` but that's only an issue in _calling_ the function in an invalid way, not an issue in defining it as we want. You're correct, this is possible. Sorry for the mixup.

bdice:
@vyasr What did you want to do with the CLI options? There are also `--doc-filter` arguments in signac's CLI and signac-flow's CLI. I was able to remove `doc_filter` from `Project.find_jobs` but it gets a little more complicated if we want to remove it from `JobsCursor` because of the interactions with aggregation and the CLI in signac-flow. We would probably need to adapt the same `*args, **kwargs` approach for `JobsCursor` and bump the minimum required version to signac 1.7 and then change signac-flow to convert all filters into unified queries.

csadorf:
This is just a quick note to say that I have acknowledged the request for review. However, I will need a little bit more time with this PR to perform it with due diligence, so I'd like to request to be given time until Friday to complete it.

bdice:
@csadorf Thanks for the heads up! Your review will be very helpful. This PR is important for finalizing the move towards unified queries and has implications for the signac CLI, signac-flow, and testing infrastructure. If this PR is too large in scope, let me know. It may be possible for me to backport some of the testing improvements and minor fixes onto `next` and create separate PRs but they will have some conflicts with this PR since there are a lot of deprecation-related changes in this PR. (Sometimes I start pulling on the “loose strings” of stuff to remove/fix and end up getting more than I expected.)

.circleci/config.yml
```diff
@@ -75,7 +75,7 @@ jobs:
           name: Run tests
           command: |
             . venv/bin/activate
-            ${PYTHON} -m pytest --cov signac --cov-config=setup.cfg --cov-report=xml tests/ -v
+            ${PYTHON} -m pytest --cov signac --cov-config=setup.cfg --cov-report=xml tests/ -v -Werror
```
bdice, line 78:
```suggestion
            ${PYTHON} -m pytest --cov signac --cov-config=setup.cfg --cov-report=xml tests/ -v
```

signac/__main__.py
```diff
@@ -153,9 +153,12 @@ def find_with_filter(args):
             return args.job_id
 
     project = get_project()
-    f = parse_filter_arg(args.filter)
-    df = parse_filter_arg(args.doc_filter)
-    return project._find_job_ids(filter=f, doc_filter=df)
+    filter_ = parse_filter_arg(args.filter) or {}
+    if args.doc_filter:
+        filter_.update(_add_prefix("doc.", parse_filter_arg(args.doc_filter)))
```
csadorf, line 158: Should we try to handle conflicts here? If yes, then I would suggest to have a canonical helper function for handling this logic and use it everywhere else as well.

signac/contrib/collection.py
```diff
@@ -936,9 +936,9 @@ def _find(self, filter=None, limit=0):
             if not _valid_filter(filter):
                 raise ValueError(filter)
             result = self._find_result(filter)
-            return set(islice(result, limit if limit else None))
         else:
-            return set(islice(self._docs.keys(), limit if limit else None))
+            result = self._docs.keys()
+        return set(islice(result, limit if limit else None))
```
csadorf, line 941: Unrelated change?

signac/contrib/project.py
```diff
@@ -52,7 +52,7 @@
 # doc_filter parameters will only be preserved for backwards compatibility but
 # not advertised as part of the API in signac 2.0.
 DOC_FILTER_WARNING = (
-    "The doc_filter argument was deprecated in version 1.7 and will be removed "
+    "The doc_filter argument is deprecated in version 1.7 and will be removed "
```
csadorf, line 55:
```suggestion
    "The doc_filter argument is deprecated since version 1.7 and will be removed "
```

signac/contrib/project.py
```diff
@@ -1910,26 +1893,19 @@ class JobsCursor:
...
     _use_pandas_for_html_repr = True  # toggle use of pandas for html repr
 
-    def __init__(self, project, filter=None, doc_filter=None):
+    def __init__(self, project, filter=None, *args, **kwargs):
         self._project = project
         self._filter = filter
 
+        doc_filter = next(iter(args), None) or kwargs.pop("doc_filter", None)
+        if len(args) > 1 or len(kwargs):
+            raise ValueError("Unsupported arguments were provided.")
```
csadorf, line 1908: Please add a test for this.

signac/contrib/project.py
```diff
@@ -1910,26 +1893,19 @@ class JobsCursor:
...
     """
 
     _use_pandas_for_html_repr = True  # toggle use of pandas for html repr
 
-    def __init__(self, project, filter=None, doc_filter=None):
+    def __init__(self, project, filter=None, *args, **kwargs):
         self._project = project
         self._filter = filter
 
+        doc_filter = next(iter(args), None) or kwargs.pop("doc_filter", None)
```
csadorf, line 1906: I'd suggest that we add a deprecation warning if the `doc_filter` argument is not None.

tests/test_project.py
```diff
@@ -106,13 +105,21 @@ def test_root_directory(self):
     def test_workspace_directory(self):
         assert self._tmp_wd == self.project.workspace()
 
+    @pytest.mark.filterwarnings(
+        "ignore:Modifying the project configuration after project initialization "
+        "is deprecated"
+    )
```
csadorf, line 111: Please separate all the unrelated filterwarnings into a separate commit.

tests/test_project.py
```diff
@@ -281,13 +297,30 @@ def test_find_jobs(self):
...
+            list(self.project.find_jobs(filter={"a": 0}, doc_filter={"b": 0}))
+        )
+        assert 1 == len(list(self.project.find_jobs({"a": 0}, doc_filter={"b": 0})))
         assert 1 == len(list(self.project.find_jobs({"sp.a": 0, "doc.b": 0})))
+        assert 0 == len(list(self.project.find_jobs({"a": 0, "doc.b": 5})))
+        assert 0 == len(list(self.project.find_jobs({"a": 0}, {"b": 5})))
+        assert 0 == len(
+            list(self.project.find_jobs(filter={"a": 0}, doc_filter={"b": 5}))
+        )
+        assert 0 == len(list(self.project.find_jobs({"a": 0}, doc_filter={"b": 5})))
```
csadorf, line 316: In principle I think we should make sure to cover all combinations of:
`(None, ...)`, `({ .. }, ...)`, `(filter=None, ...)`, `(filter={ .. }, ...)`

and

`(..., None)`, `(..., { .. })`, `(..., doc_filter=None)`, `(..., doc_filter={ .. })`

Based on this, I believe you are missing: `(None, doc_filter={...})` and `(.., doc_filter=None)`.

tests/test_job.py
```diff
@@ -435,6 +432,9 @@ def test_interface_multiple_changes(self):
             assert job.sp == job2.sp
             assert job.id == job2.id
 
+    @pytest.mark.filterwarnings("ignore:Use of int as key is deprecated")
+    @pytest.mark.filterwarnings("ignore:Use of NoneType as key is deprecated")
+    @pytest.mark.filterwarnings("ignore:Use of bool as key is deprecated")
```
csadorf, line 437: Please separate this and many other unrelated changes relating to filterwarnings marks into a separate PR.

tests/test_project.py
```diff
@@ -281,13 +297,30 @@ def test_find_jobs(self):
...
         assert 0 == len(list(self.project.find_jobs({"sp.a": 0, "doc.b": 5})))
         assert 0 == len(list(self.project.find_jobs({"sp.a": 5, "doc.b": 0})))
         assert 0 == len(list(self.project.find_jobs({"sp.a": 5, "doc.b": 5})))
+        with pytest.raises(ValueError):
+            self.project.find_jobs({"a": 0}, {"b": 5}, "invalid positional arg")
+        with pytest.raises(ValueError):
+            self.project.find_jobs({"a": 0}, invalid_kwarg="invalid argument")
```
csadorf, line 323: Please add the tests for too many arguments including with key-word arguments (`(..., ..., doc_filter=...)`).

tests/test_shell.py
```diff
@@ -255,15 +255,15 @@ def test_find(self):
                 self.call(
                     "python -m signac find ".split() + ['{"doc.a": ' + str(i) + "}"]
                 ).strip()
-                == [job.id for job in project.find_jobs(doc_filter={"a": i})][0]
+                == [job.id for job in project.find_jobs({"doc.a": i})][0]
```
csadorf, line 258: Please duplicate the test instead of altering it to ensure that we do not introduce a regression in future releases.

tests/test_project.py
```diff
@@ -304,10 +337,13 @@ def test_find_jobs_JobsCursor_contains(self):
                 assert self.project.open_job(sp) in cursor_first
             else:
                 assert self.project.open_job(sp) not in cursor_first
-        cursor_doc = self.project.find_jobs(doc_filter={"test": True})
+        cursor_doc = self.project.find_jobs({"doc.test": True})
```
csadorf, line 340: Please duplicate the test instead of altering it to ensure that we do not introduce a regression in future releases.

tests/test_shell.py
```diff
@@ -255,15 +255,15 @@ def test_find(self):
...
             )
 
         for i in range(1, 4):
             assert (
                 self.call(
                     "python -m signac find ".split() + ['{"doc.b": ' + str(i) + "}"]
                 ).strip()
-                == [job.id for job in project.find_jobs(doc_filter={"b": i})][0]
+                == [job.id for job in project.find_jobs({"doc.b": i})][0]
```
csadorf, line 266: Please duplicate the test instead of altering it to ensure that we do not introduce a regression in future releases.

csadorf (changes requested):
Nice work! I have a few change requests.

signac/contrib/project.py
bdice: I settled on "The doc_filter argument is deprecated as of version 1.7" because it matches another deprecation in the code base. This will be added in a separate PR that I'm working on, so I'm resolving this conversation.

tests/test_shell.py
bdice: This case is primarily a test of the CLI. The call to `find_jobs` is only used as a comparison value here, and it's not intending to test `find_jobs`. I think we achieve sufficient coverage of `doc_filter` parameters through the extensive testing of `find_jobs`. If you still think test duplication is warranted, then I will do so.

tests/test_shell.py
bdice: See conversation below -- this is a test of the CLI. The `find_jobs` call is only for comparison. I don't think duplication is necessary.

tests/test_project.py
bdice: Resolved by #603.

tests/test_project.py
bdice: If I understand your suggestion correctly, this is prevented by Python's function calling syntax. Did I miss something?

```python
>>> def func(a=1, b=2):
...   print(a, b)
...
>>> func(4, 5, b=7)
TypeError: func() got multiple values for argument 'b'
```

tests/test_project.py
bdice: Do you mean that you want tests that explicitly pass `filter=None` and `doc_filter=None` instead of using the default parameter values of `None`?

signac/contrib/project.py
bdice: I had some reason for not doing that the first time, potentially related to signac-flow, but it might not be an issue after all. I'll add it.

bdice: Ah, I remember why I hesitated. This code path should not be reachable if users follow our guidance not to instantiate the class directly. We don't test `JobsCursor.__init__` explicitly at all in signac, and I would prefer to remove the `doc_filter` constructor argument entirely, because we treat the constructor as private. @vyasr @csadorf Do you have opinions on this?

signac/contrib/collection.py
bdice: I pushed this small change directly to `next` because it makes code more readable. It's covered by tests.

signac/__main__.py
bdice: I think it's fine for users to have `doc_filter` override unified query parameters. I don't think writing more logic / testing provides a benefit to users when this is deprecated anyway.

signac/contrib/project.py
csadorf: I guess it would be fine then. In this case I would potentially even go so far and to just disallow it completely.

tests/test_project.py
csadorf: > Do you mean that you want tests that explicitly pass `filter=None` and `doc_filter=None` instead of using the default parameter values of `None`?

Yes, when it comes to this kind of manual arg and kwargs passing it is better to really test every possible combination. We might have gotten it right this time, but the next change might introduce a regression.

signac/contrib/project.py
```diff
@@ -732,63 +729,49 @@ def _find_job_ids(self, filter=None, doc_filter=None):
...
 
-    def find_jobs(self, filter=None, doc_filter=None):
+    def find_jobs(self, filter=None, *args, **kwargs):
         """Find all jobs in the project's workspace.
 
-        The optional filter arguments must be a Mapping of key-value pairs and
-        JSON serializable. The ``filter`` argument is used to search against job
-        state points, whereas the ``doc_filter`` argument compares against job
-        document keys.
+        The filter argument must be a JSON-serializable Mapping of key-value
```
csadorf, line 745:
```suggestion
        The filter argument must be a JSON-serializable mapping of key-value
```

signac/contrib/project.py
```diff
@@ -732,63 +729,49 @@ def _find_job_ids(self, filter=None, doc_filter=None):
...
+        pairs. The ``filter`` argument can search against both job state points
+        and job documents. See
+        https://docs.signac.io/en/latest/query.html#query-namespaces
+        for a description of supported queries.
 
         See :ref:`signac find <signac-cli-find>` for the command line equivalent.
 
         Parameters
         ----------
         filter : Mapping
```
csadorf, line 755: Is `Mapping` referring here to the ABC? That might be misleading.

tests/test_project.py
csadorf: But `find_jobs({"a": 0}, {"b: 0}", doc_filter={"c": 0})` would be technically valid, but should raise an error.

csadorf (changes requested):

signac/contrib/project.py
bdice: Current versions of signac-flow do not use this constructor, so we should be safe to change this behavior if we treat the constructor as an internal API. The only uses of the `JobsCursor` constructor are in `Project.find_jobs` and `JobsCursor.groupby`, which should be replaced by `Project.find_jobs` for consistency. I will change that in this PR.

tests/test_project.py
bdice: Done in 92ccbffd4f3c65ad850fbe0c87d7be21d0437081.

signac/contrib/project.py
bdice: We do support anything adhering to the abstract base class `Mapping`. It is correct.

signac/contrib/project.py
bdice: I prefer to leave this capitalized because it indicates the connection to `collections.abc.Mapping` as described below.

tests/test_project.py
bdice: Resolved in fc539938c1efaa693f17fb61d17795dddec3c8eb.

bdice: Resolved in 181dca732ab876e4bd430b40c78c07076a6f5637.

signac/contrib/project.py
csadorf: Ok.

signac/contrib/project.py
csadorf: Ok.

csadorf (approved):
Great! 😃

signac/contrib/project.py
```diff
@@ -732,63 +729,49 @@ def _find_job_ids(self, filter=None, doc_filter=None):
         ValueError
             If the filters are invalid.
 
-        Notes
```
vyasr, line 735: I believe this note was added because of some issue in aggregation where passing some other argument was causing severe performance degradation, but since the use case would have to be "not filtering" I'm not sure what that other arg was. Do you remember? I just want to make sure that this is not still relevant.

**EDIT**: See my next comment below, I think this is specific to cases where `JobsCursor` construction occurs but please verify.

signac/contrib/project.py
```diff
@@ -1860,38 +1843,16 @@ class JobsCursor:
     project : :class:`~signac.Project`
         Project handle.
     filter : Mapping
-        A mapping of key-value pairs that all indexed job state points are
-        compared against (Default value = None).
-
-    Notes
```
vyasr, line 1866: Ah this is what I was remembering. Where is `{}` being treated differently from `None`? In the code above, we just check `if not filter`, which will behave identically for `None` and `{}`. It looks like on line 1857 below we added logic to replace `{}` with `None`, should the comment here have been removed when that was added (I don't remember it being there)? It looks like this issue is specific to when the parameter is passed to building a `JobsCursor`, which is why it's not relevant above, is that correct?

tests/test_shell.py
vyasr: I think this is fine.

vyasr (changes requested):
I apologize for the delay here, I thought I had posted a response after your last question on CLI arguments and I'm not sure what happened. 

It looks like your decision regarding the CLI was to parse CLI doc filters into a `doc.` prefix, which looks correct. Is your question how we deprecate those parameters? We probably should be adding deprecation warnings there if we are translating doc_filter->`doc.` filter since otherwise that will hide the deprecations unless I'm missing something.

tests/test_shell.py
csadorf: You are correct and I am sorry that I didn't see this comment. All good.

tests/test_shell.py
csadorf: Agreed and sorry for not responding earlier. Somehow I missed this. All good.

signac/contrib/project.py
bdice: This was changed in #514 and there is no longer a performance difference between `{}` and `None`. Yes, you are correct that the comment about performance should have been removed at that time.

signac/contrib/project.py
bdice: See reply below -- I think this can be resolved. This was changed in #514 and there is no longer a performance difference between `{}` and `None`.

bdice:
> It looks like your decision regarding the CLI was to parse CLI doc filters into a doc. prefix, which looks correct. Is your question how we deprecate those parameters? We probably should be adding deprecation warnings there if we are translating doc_filter->doc. filter since otherwise that will hide the deprecations unless I'm missing something.

@vyasr Correct: I don't know whether we want to deprecate the `--doc-filter` CLI arguments. For now, I am parsing those CLI arguments before it gets to the Python API, so that the CLI would continue to work even if the shim in `project.find_jobs` was removed in 3.0. Outstanding questions:
- Should `signac find --doc-filter ...` be deprecated? (if so, we can alter this PR to show a warning)
- Should `--doc-filter` in signac-flow's CLI (e.g., `project.py status --doc-filter ...`) be deprecated? (if so, we need a PR to signac-flow)

vyasr:
> @vyasr Correct: I don't know whether we want to deprecate the `--doc-filter` CLI arguments. For now, I am parsing those CLI arguments before it gets to the Python API, so that the CLI would continue to work even if the shim in `project.find_jobs` was removed in 3.0. Outstanding questions:
> 
>     * Should `signac find --doc-filter ...` be deprecated and removed in 3.0? (if so, we can alter this PR to show a warning)
> 
>     * Should `--doc-filter` in signac-flow's CLI (e.g., `project.py status --doc-filter ...`) be deprecated and removed? (if so, we need a PR to signac-flow and need to choose when to remove it)
> 
> 
> (I wasn't quite sure if the end of your previous comment "We probably should be adding deprecation warnings there if we are translating doc_filter->doc. filter" was expressing an opinion on whether the CLI _should_ be deprecated and removed in 3.0.)

Yes, I think `--doc-filter` should be removed in all cases. The signac-flow PR is less urgent since it only needs to go in before whichever flow version is the first to support signac 2.0, but adding the warning for `find --doc-filter` in this PR makes sense.

bdice:
@vyasr Awesome, thanks for weighing in. I’m going to merge this and deprecate the CLI in a separate PR. This one is already fairly big and the conversation is getting hard to read.

Issue #588, opened 2021-08-02
Remove indexing methods from public API


bdice:
### Feature description

In PR #580, I remove a large portion of the `indexing.py` module. Below are some follow-up tasks to be done after #580 is merged.

- [x] Investigate making `Project.index` into a private method. This may eliminate the need for most/all of the contents of `indexing.py`. (See continued discussion below.)
- [x] Investigate removing `index` argument from a number of `Project` methods. (See continued discussion below.)
- [x] Review package documentation and remove indexing topics.
- [x] Review signac-docs to remove indexing topics. see glotzerlab/signac-docs#152
- [x] Review signac-examples to remove indexing topics.

bdice:
Discussed in dev meeting. @atravitz and I will work through this checklist and coordinate via Slack.

We agreed that `Project.index` can be deprecated in 1.x and removed in 2.0, which will eliminate the need for most/all of `indexing.py`.

Next steps (all of these can be done in parallel):
- Wait for #580 to be reviewed/merged.
- In 1.x / `master`:
  - [ ] Deprecate `Project.index` in 1.x.
  - [ ] Add deprecation warnings for all Project methods with non-empty `index` arguments in 1.x.
- In 2.0 / `next`: 
  - [ ] Make `Project.index` a private method, remove all unused arguments (only `include_job_document` should remain, based on current usage).
  - [ ] Remove `index` arguments from all Project methods.

bdice:
Next steps:
- Check whether the `Project`'s `_sp_index()` method and `_index_cache` attribute are redundant with `_sp_cache` attribute and `_index` method. There may also be some overlap with the `_build_index` or `_update_in_memory_cache` methods.
- Do we need to support custom `Job` subclasses that do not adhere to the same workspace layout? This looks like it's incompletely supported in the `Project` class and should probably be removed. https://github.com/glotzerlab/signac/blob/378a6be42d0fae70baf638a3bd7adcc97cb4888f/signac/contrib/project.py#L1966

bdice:
@atravitz No need to wait, we can go ahead and modernize and clean up examples!

bdice:
Update since this issue is marked as stale:

> Next steps:
> 
>     * Check whether the `Project`'s `_sp_index()` method and `_index_cache` attribute are redundant with `_sp_cache` attribute and `_index` method. There may also be some overlap with the `_build_index` or `_update_in_memory_cache` methods.

This is under active development in #652. I'll finalize that PR.

[...]

PR #591, opened 2021-08-04
deprecate Project.index

atravitz:
addresses issue #588 

## Motivation and Context
Indexing is deprecated and we'd like to remove `indexing.py` in v 2.0, but to do so we need to remove dependencies in `project.py`.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -14,6 +14,7 @@ Deprecated
 ++++++++++
 
  - ``Project`` methods ``read_statepoints``, ``write_statepoints``, and ``dump_statepoints`` are deprecated (#579, #197).
+ - ``Project.index`` method is deprecated (#591, #588).
 
 [1.7.0] -- 2021-06-08
 ---------------------
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -2072,6 +2072,11 @@ class Project:
             logger.debug(f"Read cache in {delta:.3f} seconds.")
             return cache
 
+    @deprecated(
+        deprecated_in="1.8",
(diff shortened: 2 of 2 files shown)


atravitz:
Looks like Bradley handled dependencies etc. in #593, so I'm just adding the deprecation message here.

cbkerr (approved):
thanks!

bdice (approved):
I added a changelog line. Merging. We will still need to go back and add deprecations to individual functions accepting an `index` argument. See #588 for a list of those methods.

atravitz:
> I added a changelog line. Merging. We will still need to go back and add deprecations to individual functions accepting an `index` argument. See #588 for a list of those methods.

@bdice I'm doing this now. Is there an easy way to only show a deprecation warning when a user passes the `index` argument? Otherwise I'll put it on all of the functions that are not already deprecated.

bdice:
@atravitz We don't want to deprecate the whole method. Try something like this:
https://github.com/glotzerlab/signac/blob/0c5a096a11caed2d47b8bee2e1f550acc850e5e2/signac/contrib/project.py#L169

For example:
```python
# At the top of the `project.py` file:
INDEX_DEPRECATION_WARNING = "The index argument is deprecated and will be removed in signac 2.0."

# In methods with an `index` argument, put this near the top of the method:
if index is not None:
    warnings.warn(INDEX_DEPRECATION_WARNING, DeprecationWarning)

PR #597, opened 2021-08-06
Clean up utility functions.

bdice:
## Description
This PR removes unused utility functions and cleans up test coverage of the `utility.py` module.

Notes:
- The `_query_yes_no` function doesn't need test coverage (it's strictly for interactive use on the terminal).
- The `_split_and_print_progress` function can probably be simplified and replaced with tqdm, if it is needed at all. The function is used only once, to print info-level logging for updates to the in-memory state point cache. This should be investigated further, but I've made the function private for now.
- There was an important edge case of `_nested_dicts_to_dotted_keys` that wasn't being tested, when a value is an empty dict. I added a test that covers that edge case.

## Motivation and Context
Cleanup for signac 2.0.

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -41,7 +41,7 @@ from .hashing import calc_id
 from .indexing import _SignacProjectCrawler
 from .job import Job
 from .schema import ProjectSchema
-from .utility import _mkdir_p, _nested_dicts_to_dotted_keys, split_and_print_progress
+from .utility import _mkdir_p, _nested_dicts_to_dotted_keys, _split_and_print_progress
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


tests/test_project.py
```diff
@@ -805,8 +805,9 @@ def test_schema(self):
         for i in range(10):
             self.project.open_job(
                 {
-                    "const": 0,
+                    "const1": 0,
                     "const2": {"const3": 0},
+                    "const4": {},
```
bdice, line 810: This edge case was being missed until I added this test: https://github.com/glotzerlab/signac/blob/257c93169709039ef4ca54e8359aace31c62b0c8/signac/contrib/utility.py#L339-L340

signac/contrib/project.py
```diff
@@ -41,7 +41,7 @@
 from .indexing import _SignacProjectCrawler
 from .job import Job
 from .schema import ProjectSchema
-from .utility import _mkdir_p, _nested_dicts_to_dotted_keys, split_and_print_progress
+from .utility import _mkdir_p, _nested_dicts_to_dotted_keys, _split_and_print_progress
```
csadorf, line 44: If I've had to guess the function was introduced prior to having the tqdm dependency or because there was some kind of threading issue or so with tqdm. But I agree, we should certainly try to replace it with tqdm if possible.

csadorf (approved):

PR #579, opened 2021-07-26
Deprecate state point backup files.

bdice:
## Description

Following conversation on #197, I am creating this PR to deprecate methods for state point backup files.

https://github.com/glotzerlab/signac/issues/197#issuecomment-786925614

Once this PR is merged into `master`, we should merge `master` into `next`. Then we can tackle changes from this PR (removing `read_statepoints`, `write_statepoints`, `dump_statepoints` and some corresponding changes in behavior) in a new PR to `next`.

## Motivation and Context
Reducing the API to core features for signac 2.0 in alignment with discussion on #197.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature
- [x] Breaking change<sup>1</sup>

<sup>1</sup>The change breaks (or has the potential to break) existing functionality.

## Checklist:
- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
- [x] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [ ] I have updated the API documentation as part of the package doc-strings.
- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
- [x] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added all related issue and pull request numbers for future reference (if applicable). See example below.

PR #599, opened 2021-08-06
Remove index arguments and state point backups

bdice:
## Description
Additional cleanup related to #588. This PR removes `index` arguments from a number of methods of `Project` and the CLI.

I also removed the public methods `read_statepoints`, `write_statepoints`, and `dump_statepoints` from the public API (see #579).

Finally, I removed the deprecated function `get_statepoint` from the public API because it used `FN_STATEPOINTS` which was removed as a part of the previously mentioned `read/write/dump_statepoints` changes. I would have split this into two pull requests but there would have been a couple annoying merge conflicts. Since this is a pure removal, I thought it would be fine to combine the two sets of changes to avoid having to deal with the conflicts.

Deprecations for the `index` argument will need to be added to signac 1.x.

## Motivation and Context
signac 2.0 API cleanup. There is no longer a public method for generating an index, so it no longer makes sense to allow users to supply one as an argument.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -823,17 +823,16 @@ class Project:
         ValueError
             If the filters are invalid.
         RuntimeError
-            If the filters are not supported by the index.
+            If the filters are not supported.
 
         Notes
         -----
-        If all arguments are ``None``, this method skips indexing the data
-        space and instead simply iterates over all job directories. This
-        code path can be much faster for certain use cases since it defers
-        all work that would be required to construct an index, so in
-        performance-critical applications where no filtering of the data space
-        is required, passing no arguments to this method (as opposed to empty
-        dict filters) is recommended.
+        If all arguments are ``None``, this method simply returns a list of all
+        job directories. This code path can be much faster for certain use cases
+        since it defers all work that would be required to construct an index,
+        so in performance-critical applications where no filtering of the data
+        space is required, passing no arguments to this method (as opposed to
+        empty dict filters) is recommended.
 
         """
         if not filter and not doc_filter:
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -817,8 +817,6 @@ class Project:
             If the filters are not JSON serializable.
         ValueError
             If the filters are invalid.
-        RuntimeError
-            If the filters are not supported.
 
         Notes
         -----
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -213,11 +213,6 @@ class Project:
     KEY_DATA = "signac_data"
     "The project's datastore key."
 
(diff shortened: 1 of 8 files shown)


signac/contrib/project.py
```diff
@@ -851,50 +818,48 @@ def _find_job_ids(self, filter=None, doc_filter=None, index=None):
         ValueError
             If the filters are invalid.
         RuntimeError
-            If the filters are not supported by the index.
+            If the filters are not supported.
```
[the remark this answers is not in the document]
bdice: I don't think that a `RuntimeError` can be raised by this code.

signac/contrib/project.py
```diff
@@ -850,51 +817,47 @@ def _find_job_ids(self, filter=None, doc_filter=None, index=None):
...
-        all work that would be required to construct an index, so in
-        performance-critical applications where no filtering of the data space
-        is required, passing no arguments to this method (as opposed to empty
-        dict filters) is recommended.
+        If all arguments are ``None``, this method simply returns a list of all
+        job directories. This code path can be much faster for certain use cases
+        since it defers all work that would be required to construct an index,
+        so in performance-critical applications where no filtering of the data
+        space is required, passing no arguments to this method (as opposed to
+        empty dict filters) is recommended.
```
csadorf, line 828: I think empty dict filters should be equivalent to no argument due to the condition tested in line 831.

tests/test_project.py
```diff
@@ -614,18 +598,16 @@ def test_repair_corrupted_workspace(self):
                 for job in self.project:
                     # Accessing the job state point triggers validation of the
                     # state point manifest file
-                    job.statepoint
-
-            with pytest.raises(JobsCorruptedError):
-                self.project.repair()
+                    job.statepoint()
```
csadorf, line 601: I assume the call is needed due to lazy loading?

bdice: Yes, I changed this just to be sure that loading and validation are performed. I think this test might have skipped that check on the second (post-repair) iteration until I made this change to fix it. Iterating over the project is not enough, the state point must be accessed.

csadorf: Might be best to than explicitly check the return value, e.g., with:
```suggestion
                    assert len(job.statepoint()) > 0
```
Or something like that.

bdice: Good idea. Done in d194c3bc.

signac/contrib/project.py
bdice: Done in 0a490d0a.

csadorf (approved):

PR #607, opened 2021-08-18
Fix H5Store.mode.

bdice:
## Description
- Fix and test H5store.mode property.
- Add some missing tests for H5Store and H5Group.

## Motivation and Context
The `H5Store.mode` property wasn't tested and did not work correctly. I added a few more tests for features that were not executed in the test coverage.

## Types of Changes
- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/tests/test_h5store.py
+++ b/tests/test_h5store.py
@@ -118,7 +118,7 @@ class TestH5StoreOpen(TestH5StoreBase):
         with self.open_h5store(mode="w") as h5s_w:
             with self.open_h5store(mode="r") as h5s_r:
                 assert h5s_w.mode == "r+"
-                assert h5s_r.mode == "r"
+                assert h5s_r.mode == "r+"
                 assert "foo" not in h5s_r
                 assert "foo" not in h5s_w
 
--- a/tests/test_h5store.py
+++ b/tests/test_h5store.py
@@ -226,6 +226,24 @@ class TestH5Store(TestH5StoreBase):
             self.assertEqual(child1[key], d)
             self.assertEqual(child2[key], d)
 
+    def test_setdefault_group(self):
+        with self.open_h5store() as h5s:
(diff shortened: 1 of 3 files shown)


tests/test_h5store.py
```diff
@@ -103,18 +103,22 @@ def test_open_read_only(self):
...
+            assert h5s.mode == "r+"
             h5s["foo"] = "bar"
             assert "foo" in h5s
             self.assertEqual(h5s["foo"], "bar")
 
     def test_open_write_and_read_only(self):
         with self.open_h5store(mode="w") as h5s_w:
             with self.open_h5store(mode="r") as h5s_r:
+                assert h5s_w.mode == "r+"
+                assert h5s_r.mode == "r+"
```
bdice, line 121: I expected this mode to be `r` instead of `r+`. I did an experiment with swapping the order of these lines. I found it's not possible (on Linux, at least) to open a writeable handle to the file if the read-only handle is created first. To me, this calls into question whether this test is actually useful: the file handle that we're claiming is read-only is NOT read-only -- it is actually using the same file handle as the writeable store. Maybe this doesn't matter. I don't know. 🤷‍♂️

b-butler: We could potential raise a warning in this case if we know that the store is open currently in a writable mode. However, that doesn't need to be in this PR. I would personally remove the test, as it doesn't test anything (well the behavior it is meant to test doesn't work).

b-butler (approved):
Looks good to me, and the new tests look nice. Feel free to delete the test, if you agree.

tests/test_h5store.py
bdice: I see what you mean. Let's leave the test for now, because if this behavior of the upstream library ever changed or varied by platform, we would like to know.

PR #608, opened 2021-08-18
Disallow Project.config mutation

bdice:
## Description
This PR makes the `Project.config` immutable. Follows up on #246 with the planned changes for signac 2.0.

To-do:
- [x] Update tests to pass and ensure that the configuration is immutable.
- [x] Identify and apply optimizations that are now possible with the immutable configuration. There are some changes to make with caching the workspace / project root directories.
- [x] Revise the docs for the `Project.config` attribute.

## Motivation and Context
See #244 for a discussion where the decision was made to have an immutable Project configuration.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -4,6 +4,17 @@ Changelog
 
 The **signac** package follows `semantic versioning <https://semver.org/>`_.
 
+Version 2
+=========
+
+[2.0.0] -- 2021-xx-xx
+---------------------
+
+Removed
++++++++
+
+ - ``Project.config`` is no longer mutable. Use the command line ``$ signac config`` to modify configuration (#608, #246, #244).
+
 Version 1
 =========
 
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -207,6 +207,12 @@ class Project:
     def config(self):
         """Get project's configuration.
 
+        The configuration is immutable once the Project is constructed. To
+        modify a project configuration, use the command line or edit the
+        configuration file directly.
+
+        See :ref:`signac config <signac-cli-config>` for related command line tools.
+
         Returns
         -------
         :class:`~signac.contrib.project._ProjectConfig`
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -130,6 +130,7 @@ class Project:
             )
 
         # Prepare root directory and workspace paths.
+        # os.path is used instead of pathlib.Path for performance.
         self._root_directory = self.config["project_dir"]
         self._workspace = os.path.expandvars(
             self.config.get("workspace_dir", "workspace")
@@ -625,8 +626,9 @@ class Project:
             True if the job id is initialized for this project.
 
         """
-        # We can rely on the project workspace to be well-formed, so just use
-        # str.join with os.sep instead of os.path.join for speed.
+        # Performance-critical path. We can rely on the project workspace and
+        # job id to be well-formed, so just use str.join with os.sep instead of
+        # os.path.join for speed.
         return os.path.exists(os.sep.join((self.workspace(), job_id)))
 
     def __contains__(self, job):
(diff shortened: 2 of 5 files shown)


changelog.txt
```diff
@@ -4,6 +4,17 @@ Changelog
...
+Version 2
+=========
+
+[2.0.0] -- 2021-xx-xx
+---------------------
+
+Removed
++++++++
+
+ - ``Project.config`` is no longer mutable. Use the command line ``$ signac config`` to modify configuration (#608, #246, #244).
```
csadorf, line 16:
```suggestion
 - The ``Project.config`` property is no longer mutable. Use the command line ``$ signac config`` to modify configuration (#608, #246, #244).
```
Just to make it clear that we are talking about the Python API here, not the project configuration in general.

signac/contrib/project.py
```diff
@@ -135,36 +115,29 @@ class Project:
...
-        # Prepare cached properties derived from the project config.
-        self._id = None
-        self._rd = None
-        self._wd = None
-
         # Ensure that the project id is configured.
-        if self.id is None:
+        try:
+            self._id = str(self.config["project"])
+        except KeyError:
```
csadorf, line 124: Any chance you could add a quick test for this? Should be fairly straight-forward.

signac/contrib/project.py
```diff
@@ -135,36 +115,29 @@ class Project:
...
-        if not _ignore_schema_version:
-            self._check_schema_compatibility()
-
-        # Prepare project document
-        self._document = None
-
-        # Prepare project H5StoreManager
-        self._stores = None
+        # Prepare root directory and workspace paths.
+        self._root_directory = self.config["project_dir"]
```
csadorf, line 133: Do we already have an issue to start using `pathlib.Path` for these objects?

signac/contrib/project.py
```diff
@@ -241,19 +230,15 @@ def root_directory(self):
...
-        The workspace defaults to `project_root/workspace`.
-        Configure this directory with the 'workspace_dir'
-        attribute.
-        If the specified directory is a relative path,
-        the absolute path is relative from the project's
-        root directory.
+        The workspace defaults to `project_root/workspace`. Configure this
+        directory with the ``'workspace_dir'`` configuration option. If the
+        specified directory is a relative path, the absolute path is relative
+        from the project's root directory.
```
csadorf, line 241:
```suggestion
        The workspace defaults to `project_root/workspace`. Configure this
        directory with the ``'workspace_dir'`` configuration option. A relative
        path is assumed to be relative to the project's root directory.
```

csadorf (approved):

signac/contrib/project.py
bdice: Done in 994ab740ab5397f9be79e11f9d2de3a91541578d.

signac/contrib/project.py
bdice: I did some benchmarking at one point. Switching to `Path` objects affects performance somewhat significantly. The construction and manipulation of `Path` objects involve a lot of system calls that do not happen if using the `os` library functions like `os.path.join`, etc. I would advocate keeping the current "string path" approach.

bdice: See also: https://github.com/glotzerlab/signac/pull/511#issuecomment-782292032

csadorf: Ok, good to know. 👌

csadorf: @bdice Considering that this question might come up again. Would you mind adding a comment into the code so that future developers are aware?

bdice: @csadorf Where should the comment go?

csadorf: Since a future change would likely need to inspect these variables:
https://github.com/glotzerlab/signac/blob/a5f6c3048c1cd9e1defc1c5521ba2e49bf2e3e43/signac/contrib/project.py#L266-L267
Either right there or above the `import os` statement:
https://github.com/glotzerlab/signac/blob/a5f6c3048c1cd9e1defc1c5521ba2e49bf2e3e43/signac/contrib/project.py#L10
We could also add an explicit `import os.path` import statement and add the comment right there.

csadorf: It is generally a good idea to be pretty explicit in the code about what are critical paths, e.g., like here:
https://github.com/glotzerlab/signac/blob/a5f6c3048c1cd9e1defc1c5521ba2e49bf2e3e43/signac/contrib/project.py#L2587-L2588

bdice: @csadorf Thanks for the suggestion, you're absolutely right that this is important. I improved this in 0e1be15c8ae0fae660eb341ca4accddbfaf68859 and gave greater explanation for a few optimizations that are not obvious.

PR #612, opened 2021-08-22
Avoid AssertionError.

bdice:
## Description
There are a few places where signac raises `AssertionError` as a way to validate program state via `assert` statements. This is not a good practice, because "optimized" execution of Python will ignore these assertions, potentially leading to unwanted or invalid program states. See also: https://stackoverflow.com/questions/9929456/keeping-assert-despite-the-o-optimize-flag

## Motivation and Context
Makes some internal logic more resilient if executing Python with optimization flags.

## Types of Changes
- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/synced_collections/_caching.py
+++ b/signac/synced_collections/_caching.py
@@ -37,11 +37,12 @@ def get_cache():
             cache = redis.Redis()
             test_key = str(uuid.uuid4())
             cache.set(test_key, 0)
-            assert cache.get(test_key) == b"0"  # Redis stores data as bytes
+            if cache.get(test_key) == b"0":  # Redis stores data as bytes
+                raise RuntimeError("Cache access check failed.")
             cache.delete(test_key)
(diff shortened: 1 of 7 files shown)


kidrahahjo (approved):
Changes looks good to me

signac/synced_collections/_caching.py
```diff
@@ -37,11 +37,12 @@ def get_cache():
             cache = redis.Redis()
             test_key = str(uuid.uuid4())
             cache.set(test_key, 0)
-            assert cache.get(test_key) == b"0"  # Redis stores data as bytes
+            if cache.get(test_key) == b"0":  # Redis stores data as bytes
```
bdice, line 40:
```suggestion
            if cache.get(test_key) != b"0":  # Redis stores data as bytes
```

PR #622, opened 2021-09-16
Fix bug for when CLI signac find command used with lists

### LINKED ISSUE #615: Using signac find on the CLI for tuples
### Description

Finding jobs for which a given statepoint parameter, x, is equal to a particular list cannot be as easily done with the `signac find` CLI option as it can with other statepoint parameters. For example, `signac find x '[0, 1, 0]'` is currently interpreted as `'{"x": "[0, 1, 0]"}'` . The user has to explicitly specify `signac find '{"x": [0, 1, 0]}'` to get around this.

### To reproduce

`signac find x '[0, 1, 0]'` will return no jobs, even if x = [0, 1, 0] for some of them

### Error output

No error is thrown. The command is only misinterpreted.

### System configuration

 - Operating System [e.g. macOS]: macOS
 - Version of Python [e.g. 3.7]: 3.9.6
 - Version of signac [e.g. 1.0]: 1.7.0

AlainKadar:
## Description

Signac interprets `signac find x '[0, 1, 0]'` as `'{"x": "[0, 1, 0]"}'`. The PR ensures signac interprets `signac find x '[0, 1, 0]'` as `'{"x": [0, 1, 0]}'`.
## Motivation and Context
Fixes #615 

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/filterparse.py
+++ b/signac/contrib/filterparse.py
@@ -62,23 +62,6 @@ def _read_index(project, fn_index=None):
         return (json.loads(line) for line in fd)
 
 
-def _is_json(q):
-    """Check if q is JSON.
-
-    Parameters
-    ----------
-    q : str
-        Query string.
-
-    Returns
-    -------
-    bool
-        True if q starts with "{" and ends with "}".
-
-    """
-    return (q[0] == "{" and q[-1] == "}") or (q[0] == "[" and q[-1] == "]")
-
-
 def _is_regex(q):
     """Check if q is a regular expression.
 
@@ -181,19 +164,16 @@ def _parse_single(key, value=None):
         If filter arguments have an invalid key.
 
     """
-    if _is_json(key):
-        raise ValueError(
-            "Please check your filter arguments. "
-            "Using a JSON expression as a key is not allowed: '{}'.".format(key)
-        )
-    elif value is None or value == "!":
+    
+    if value is None or value == "!":
         return key, {"$exists": True}
-    elif _is_json(value):
-        return key, _parse_json(value)
     elif _is_regex(value):
         return key, {"$regex": value[1:-1]}
     else:
-        return key, _cast(value)
+        try:
+            return key, _parse_json(str(value))
+        except json.JSONDecodeError:
+            return key, _cast(value)
 
 
 def parse_simple(tokens):
@@ -238,14 +218,13 @@ def parse_filter_arg(args, file=sys.stderr):
     if args is None or len(args) == 0:
         return None
     elif len(args) == 1:
-        if _is_json(args[0]):
+        try:
             return _parse_json(args[0])
-        else:
+        except json.JSONDecodeError:
             key, value = _parse_single(args[0])
             return _with_message({key: value}, file)
     else:
         q = dict(parse_simple(args))
-
         return _with_message(q, file)
 
 
--- a/signac/contrib/filterparse.py
+++ b/signac/contrib/filterparse.py
@@ -96,6 +96,23 @@ def _is_regex(q):
     return q.startswith("/") and q.endswith("/")
 
 
+def _is_list(q):
+    """Check if q is a list.
+
+    Parameters
+    ----------
+    q : str
+        Query string.
+
+    Returns
+    -------
+    bool
+        True if q starts with "[" and ends with "]".
+
+    """
+    return q.startswith("[") and q.endswith("]")
+
+
 def _parse_json(q):
     """Parse a query argument as JSON.
 
@@ -183,7 +200,7 @@ def _parse_single(key, value=None):
     """
     if value is None or value == "!":
         return key, {"$exists": True}
-    elif _is_json(value):
+    elif _is_json(value) or _is_list(value):
         return key, _parse_json(value)
     elif _is_regex(value):
         return key, {"$regex": value[1:-1]}
--- a/signac/contrib/filterparse.py
+++ b/signac/contrib/filterparse.py
@@ -76,7 +76,7 @@ def _is_json(q):
         True if q starts with "{" and ends with "}".
 
     """
-    return q.strip().startswith("{") and q.strip().endswith("}")
+    return (q[0] == "{" and q[-1] == "}") or (q[0] == "[" and q[-1] == "]")
 
 
 def _is_regex(q):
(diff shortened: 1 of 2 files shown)


signac/contrib/filterparse.py
```diff
@@ -183,7 +200,7 @@ def _parse_single(key, value=None):
     """
     if value is None or value == "!":
         return key, {"$exists": True}
-    elif _is_json(value):
+    elif _is_json(value) or _is_list(value):
```
bdice, line 203: I wonder if `_is_json` should be implemented as:
```python
    return (q[0] == "{" and q[-1] == "}") or (q[0] == "[" and q[-1] == "]")
```

This would have several advantages:
- It's more accurate to the meaning of "valid JSON" because `[0, 1, 2]` [is valid JSON](https://www.json.org/json-en.html) per the specification
- It prevents the invalid use of list-like keys (see `elif _is_json(key)` and the `ValueError` it raises).
- It is faster to check the individual character at the start and end rather than `str.startswith`. I find it just as readable.
https://github.com/bdice/python_microbenchmarks/blob/main/microbenchmarks/str_startswith_index.py

Then the `_is_list` method can be removed.

```suggestion
    elif _is_json(value):
```

The only other thing this needs is a test in [test_shell.py](https://github.com/glotzerlab/signac/blob/master/tests/test_shell.py).

AlainKadar: Ok, yes, I think the change to `_is_json` is a lot cleaner. I also think the `ValueError` message should be raised at the beginning of the method. I'll make these changes and add the test. Thanks!

tests/test_shell.py
```diff
@@ -269,8 +270,17 @@ def test_find(self):
         assert '{"a": 0}' in out
         assert '{"a": 2}' in out
 
+        job = project.open_job({"a": [0, 1, 0]})
+        msg = "python -m signac find a".split()
+        msg.append('[0, 1, 0]')
+        msg.append('--sp')
```
AlainKadar: Adding list after calling `split()` because this method splits up list element into strings, even when the whole list is in quotations. (Is there more elegant solution?)

signac/contrib/filterparse.py
vyasr: One important question here is whether or not parsing a filter is ever a performance bottleneck. @bdice can you think of a performance-critical code path you've benchmarked where parsing is even a blip? I can't, but I'm also not in the right mindset at the moment to really think through this.

If not, I would advocate for correctness over performance. @AlainKadar Instead of checking `_is_json`, I would get rid of that method entirely and simply use the `_parse_json` method, which is a try-except that will tell you whether a string is valid JSON. It will be slower, but it will be more accurate and bypass any concerns about whether we are correctly identifying JSON strings.

bdice: I agree that would be better. I generously assumed that the current code structure was designed this way on purpose for some reason, but I think it’s probably a good idea to just parse the whole input to ensure its validity as Vyas suggested.

tests/test_shell.py
vyasr: No, there's not really a more elegant solution without doing more work than you want to. The solutions that I can think of would involve using `re.split` or `csv.reader`, both of which seem pretty heavyweight for this purpose.

signac/contrib/filterparse.py
AlainKadar: Ok makes sense. But I should still keep the `_is_json` for testing whether the key is a JSON, at the beginning of the method?

AlainKadar:
Can I remove the `test_invalid_json` test in test_find_command_line_interface.py? The CLI works fine with the changes; passing an invalid JSON still prints "Failed to parse query argument. Ensure that '{"x": True}' is valid JSON!" but pytest reports the error is not raised. This is because I no longer allow invalid JSONs to be passed to `_parse_json()` ?

bdice:
@AlainKadar Changing the tests should be avoided. In the changes @vyasr suggested, we lose one feature that I hadn't anticipated: we have no way to tell the user when they have provided invalid JSON, because we just treat it as a string if it cannot be parsed as JSON. The previous code attempted to recognize JSON-like data, which would catch invalid JSON like `{"x": True}` -- note that JSON uses `true` instead of `True`.

@vyasr Would you be open to reverting the eager parsing and returning to the previous approach where we attempt to detect JSON-like inputs? I can certainly imagine that our invalid JSON detection has saved a lot of users who may not be familiar with how the format differs from Python literals -- it has caught my mistakes several times.

vyasr:
> @vyasr Would you be open to reverting the eager parsing and returning to the previous approach where we attempt to detect JSON-like inputs? I can certainly imagine that our invalid JSON detection has saved a lot of users who may not be familiar with how the format differs from Python literals -- it has caught my mistakes several times.

That's fair, I hadn't thought that would be too important but I'm happy to preserve if if it's helpful to people. In that case, we should rename the function something like `_is_json_like` or `_may_be_json` to indicate that returning true here is a necessary but not sufficient condition for the input to be JSON.

signac/contrib/filterparse.py
```diff
@@ -181,17 +181,18 @@ def _parse_single(key, value=None):
...
-    elif _is_json(value):
-        return key, _parse_json(value)
-    elif _is_regex(value):
-        return key, {"$regex": value[1:-1]}
-    elif _is_json(key):
+
+    if _is_json_like(key):
         raise ValueError(
             "Please check your filter arguments. "
             "Using a JSON expression as a key is not allowed: '{}'.".format(key)
```
bdice, line 188:
```suggestion
            f"Using a JSON expression as a key is not allowed: '{key}'."
```

tests/test_shell.py
```diff
@@ -269,8 +270,17 @@ def test_find(self):
         assert '{"a": 0}' in out
         assert '{"a": 2}' in out
 
+        job = project.open_job({"a": [0, 1, 0]})
+        msg = "python -m signac find a".split()
+        msg.append("[0, 1, 0]")
+        msg.append("--sp")
```
bdice, line 276: Optional suggestion. Shorter but may be slightly less readable.
```suggestion
        msg = [*"python -m signac find a".split(), "[0, 1, 0]", "--sp"]
```

bdice (approved):
@AlainKadar Nice work. I think this is good. I have two minor suggestions that you can apply if you wish.

edit: also fix the style issue caught by pre-commit for flake8.

AlainKadar:
Style changes made, thanks for the advice

vyasr (approved):

PR #627, opened 2021-10-03
Update filelock pinning to >= instead of ~=.

bdice:
## Description
Currently our pinning for `filelock` is `~=`, which differs from how we handle all other dependencies. I have changed it to `>=` to avoid Dependabot PRs like #623, which is not aligned with our intentions for the pinning.

## Motivation and Context
Maintenance.

## Checklist:

If necessary:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 deprecation>=2
-filelock~=3.0
+filelock>=3.0
 packaging>=15.0
 tqdm>=4.10.0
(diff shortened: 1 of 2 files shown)


csadorf (approved):
It's still odd to me that using the compatible-version operator is discouraged, since a filelock 4.x release could potentially be breaking, which means all releases that depend on it are broken, but since we do it everywhere else, fine.

PR #629, opened 2021-10-10
Feature: Measuring the airspeed velocity of an unladen signac

bdice:
## Description
This PR uses [asv (airspeed velocity)](https://asv.readthedocs.io/en/stable/writing_benchmarks.html#setup-and-teardown) to measure performance of the signac package. It could replace our existing `benchmark.py` script. I haven't yet tried all the features of asv, but it does a lot of things we care about and is a very powerful / helpful tool:
- isolated testing of each commit in a virtual environment (which it handles for you)
- automated testing from git ranges: running `asv run v1.3.0..master` tests every commit from v1.3.0 to master (with the current commit's set of benchmarks, so it doesn't require any git trickery!)
- publishing results into a static HTML site that can be hosted/shared: `asv publish`
- local preview of the HTML site: `asv preview`
- checking that the benchmark suite runs correctly during development: `asv dev`
- profiling+visualization with snakeviz: e.g. `asv profile 'benchmarks.ProjectBench.time_iterate_load_sp(.*)' --gui=snakeviz`
- direct comparison of two commits with `asv compare`
- automatic detection of performance regressions 👀

Preview of asv results from v1.3.0 to master (this link may be removed in the future): https://glotzerlab.github.io/signac/

To-do:
- [x] Decide what to do with `benchmark.py` (see below)
- [x] Update changelog
- [ ] Update docs
- [ ] Port this feature over to signac-flow and close https://github.com/glotzerlab/signac-flow/pull/399

## Motivation and Context
I have always wanted to try asv and I think this is a great tool for understanding the project's longitudinal improvements. From a reporting/visualization/profiling perspective, asv seems like a massive step forward. Especially since releases 1.6.0 and 1.7.0 were very carefully tested for performance, it would be nice to add this tool to ensure that we keep high performance into signac 2.0 and beyond.

We can separately discuss whether to remove `benchmark.py`. For now, I'm happy to keep both. To get rid of `benchmark.py` we'll want to decide how to incorporate all the flags from `benchmark.py`, like those controlling document sizes, number of state point keys, etc. I think I would recommend that we just add those as parameters, and then let users alter the settings of the benchmark file if they need to test particular configurations (e.g. large documents).

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -4,6 +4,7 @@ import random
 import string
 from itertools import islice
 from multiprocessing import Pool
+from tempfile import TemporaryDirectory
 
 from tqdm import tqdm
 
@@ -59,9 +60,10 @@ def setup_random_project(
     if not isinstance(N, int):
         raise TypeError("N must be an integer!")
 
-    project = signac.init_project(f"benchmark-N={N}")
+    temp_dir = TemporaryDirectory()
+    project = signac.init_project(f"benchmark-N={N}", root=temp_dir.name)
     generate_random_data(project, N, num_keys, num_doc_keys, data_size, data_std)
-    return project
+    return project, temp_dir
 
 
 PARAMETERS = {"N": [100, 1_000]}
@@ -73,7 +75,10 @@ class _ProjectBenchBase:
 
     def setup(self, *params):
         (N,) = params
-        self.project = setup_random_project(N)
+        self.project, self.temp_dir = setup_random_project(N)
+
+    def teardown(self, *params):
+        self.temp_dir.cleanup()
 
 
 class ProjectBench(_ProjectBenchBase):
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -1,0 +1,109 @@
+# Copyright 2021 The Regents of the University of Michigan
+
+import random
+import string
+from itertools import islice
+from multiprocessing import Pool
+from tempfile import TemporaryDirectory
+
+from tqdm import tqdm
+
+import signac
+
+
+def _random_str(size):
+    return "".join(random.choice(string.ascii_lowercase) for _ in range(size))
+
+
+def _make_doc(i, num_keys=1, data_size=0):
+    assert num_keys >= 1
+    assert data_size >= 0
+
+    doc = {f"b_{j}": _random_str(data_size) for j in range(num_keys - 1)}
+    doc["a"] = f"{i}{_random_str(max(0, data_size - len(str(i))))}"
+    return doc
+
+
+def _make_job(project, num_keys, num_doc_keys, data_size, data_std, i):
+    size = max(0, int(random.gauss(data_size, data_std)))
+    job = project.open_job(_make_doc(i, num_keys, size))
+    if num_doc_keys > 0:
+        size = max(0, int(random.gauss(data_size, data_std)))
+        job.document.update(_make_doc(i, num_doc_keys, size))
+    else:
+        job.init()
+
+
+def generate_random_data(
+    project, N_sp, num_keys=1, num_doc_keys=0, data_size=0, data_std=0, parallel=True
+):
+    assert len(project) == 0
+
+    if parallel:
+        with Pool() as pool:
+            p = [
+                (project, num_keys, num_doc_keys, data_size, data_std, i)
+                for i in range(N_sp)
+            ]
+            list(pool.starmap(_make_job, tqdm(p, desc="init random project data")))
+    else:
+        from functools import partial
+
+        make = partial(_make_job, project, num_keys, num_doc_keys, data_size, data_std)
+        list(map(make, tqdm(range(N_sp), desc="init random project data")))
+
+
+def setup_random_project(
+    N, num_keys=1, num_doc_keys=0, data_size=0, data_std=0, seed=0, root=None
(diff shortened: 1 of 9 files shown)


csadorf:
@SchoeniPhlippsn Do you think you will be able to have a look at this soon?
@bdice Regardless of the missing review, this might be ready to move forward. Anyway I can help?

bdice:
> @SchoeniPhlippsn Do you think you will be able to have a look at this soon?
> @bdice Regardless of the missing review, this might be ready to move forward. Any way I can help?

Yes! The next action item is to move some additional options like document size, statepoint key/value sizes, etc. into the new benchmark script. I think we can ignore the ability to set those via command line and just edit the script (and rerun asv) if those values need to be set for a particular benchmark.

SchoeniPhlippsn (commented):
Sorry for the late response. I missed this request for review. 
I don't have any more comments. So as soon as Brandon's review has been addressed, I think this is good to go.

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,115 @@
...
+    def time_iterate_single_pass(self, *params):
+        list(self.project)
+
+    def time_iterate(self, *params):
+        for _ in range(10):
+            list(self.project)
+
+    def time_iterate_load_sp(self, *params):
+        for _ in range(10):
+            [job.sp() for job in self.project]
```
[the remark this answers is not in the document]
bdice: I'd like to leave the scope of _what_ we benchmark the same in this PR as we have currently implemented in `benchmark.py`. I am unsure how lazy loading / caching would work if we added another test that loads statepoints here. Ideally we want to have an empty cache when each test executes, but I can't remember if the setup/teardown are run for each test method or each test class.

bdice:
@b-butler @csadorf What do you think we should do with `benchmark.py`? Currently we use `benchmark.py` for CI testing to ensure that we haven't introduced serious regressions, so I don't want to delete it. I would vote to keep the existing `benchmark.py` and its corresponding CI tests. The only downside is that we will have two ways to perform benchmarks, but otherwise we have to re-implement all the CI tooling for asv in this PR.

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,137 @@
...
+
+def setup_random_project(
+    N, num_keys=1, num_doc_keys=0, data_size_mean=0, data_size_std=0, seed=0, root=None
+):
+    random.seed(seed)
+    if not isinstance(N, int):
+        raise TypeError("N must be an integer!")
+
+    temp_dir = TemporaryDirectory()
+    project = signac.init_project(f"benchmark-N={N}", root=temp_dir.name)
```
csadorf, line 72: Any reason to not use (a potentially extended/adapted variant of) https://github.com/glotzerlab/signac/blob/e60bfbc579371698c2690605d10f88b3a12bc27c/signac/contrib/project.py#L2412 and https://github.com/glotzerlab/signac/blob/8a804d0e9b513e576ae622b9044914ff16d0ac9a/signac/testing.py#L8 here?

csadorf:
> @b-butler @csadorf What do you think we should do with `benchmark.py`? Currently we use `benchmark.py` for CI testing to ensure that we haven't introduced serious regressions, so I don't want to delete it. I would vote to keep the existing `benchmark.py` and its corresponding CI tests. The only downside is that we will have two ways to perform benchmarks, but otherwise we have to re-implement all the CI tooling for performance tests with asv in this PR.

I think it's fine to leave it for now, but add an in-code comment where appropriate about the purpose of each method and about the motivation to keep both.

b-butler (approved):
I am fine with this as is.

benchmarks/benchmarks.py
bdice: The `TemporaryProject` yields a project when used as a context manager. The benchmark script requires something we can store (like `self.project` and `self.temp_dir`) in `setup` and then destroy during the `teardown` method. I'm not aware of a clean way to use that context manager here.

bdice:
@csadorf Feel free to give this a final pass of review if you'd like. Otherwise it's ready to merge. I added docs here: https://signac--629.org.readthedocs.build/projects/core/en/629/support.html#benchmarking

benchmarks/benchmarks.py
csadorf: Well, even if you don't use `TemporaryProject`, you could still use the `testing.init_jobs()` function?

csadorf (approved):
I have one more question that I would like to have considered, but it does not block approval.

Not sure how all of these http->https fixes made it into this PR, but I will allow it. 😆

benchmarks/benchmarks.py
bdice: Hmm, I had forgotten about that function. It looks like it is currently designed for tests of projects with complex schemas, not tests of performance with varying data sizes. We would probably have to change that function to support the arguments like `num_keys`, `num_doc_keys`, `data_size`, `data_std`, etc., and that feels out of scope for this PR. I'll finalize and merge as-is, since you indicated this is a non-blocking issue.

bdice:
> Not sure how all of these http->https fixes made it into this PR, but I will allow it. 😆

My mistake, I meant to push that fix directly to master and didn't notice that it was on this branch.

PR #642, opened 2021-11-03
Refactor path function handling

cbkerr:
This PR switched from a bug fix to a refactor. See #666 for the bug fix only.

## Original Title: Don't generate views with underspecified path provided by user
Signac has unexpected behavior when generating a view if the user specifies a custom path that doesn't uniquely specifiy jobs.

I would expect signac to link to all jobs matching the description in the view folder.

What @Nipuli-Gunaratne and I found was that signac just picks one job.

## Description

We check that the mapping of source-> link is 1-1 in other parts of `import_export.py` but don't check user's provided path function.

Two posssible solutions I see are:
1. Generate an error and exit. Suggest the fix in the error message.
2. Try to fix the problem by adding jobs ids to the path.

In this draft PR, I find the places in the code for adding the error checking code in two places it might fit: `import_export.py::_make_path_function` and in `linked_view.py::create_linked_view`.

I think it works best in `_make_path_function`.

## Motivation and Context

If you make this test project
```python
#init.py
import signac

project = signac.init_project('Test-view')

jobs = [dict(a=1,b=1),
        dict(a=1,b=2),
        dict(a=2,b=1),
        dict(a=2,b=2)
        ]


for j in jobs:
    job = project.open_job(j)
    job.init()
    print(j, job.id)
```

and generate a view with a user-specified custom path that does not uniquely identify jobs
```shell
signac view test_error "a/{a}"
```

Signac just picks one of the jobs to link.
a=1 gets job 8aacdb17187e6acf2b175d4aa08d7213 (b=2) and not 386b19932c82f3f9749dd6611e846293 (b=1)                                                                      
a=2 gets job 5e4d14d82c320bafb2f1286fe486d1f8 (b=1) and not d48f81ad571306570e2eb9fe7920cd3c (b=2)

### Fix that we'd suggest to users OR try to do automatically:
Remake the path specification as "a/{a}/id/{id}"

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:


signac/contrib/import_export.py
```diff
@@ -240,12 +240,17 @@ def path_function(job):
                 raise _SchemaPathEvaluationError(f"Unknown key: {error}")
             except Exception as error:
                 raise _SchemaPathEvaluationError(error)
+        # Check that the user-specified path generates a unique mapping
+        link_check = {j.workspace(): path_function(j) for j in jobs}
+        if len(set(link_check.values())) != len(link_check):
+            print("Error caught inside _make_path_function")
```
cbkerr, line 247: Still printing errors while thinking about what to do

signac/contrib/linked_view.py
```diff
@@ -99,19 +99,26 @@ def create_linked_view(project, prefix=None, job_ids=None, index=None, path=None
         raise RuntimeError(err_msg)
 
     path_function = _make_path_function(jobs, path)
-
-    links = {}
+    incomplete_links = {}
```
cbkerr, line 102: The old code generated the link mapping in a way that we could not identify.

The old code put the user specified path as the key in the dictionary. Everywhere else we put the actual job path as the key of the dictionary.

I changed the 1 place in the code that was affected by swapping this dictionary around.

I renamed `links` to `incomplete_links` to see both ways of doing things

b-butler (commented):
I think we should raise an error here generally, and also remove `incomplete_links` in favor of `links`

cbkerr:
> I think we should raise an error here generally, and also remove `incomplete_links` in favor of `links`

I added the error in `import_export.py` but as I was trying to use the inverted dictionary for links, I realized that the existing `_update_view` function needs the link as the key.

I think it will be least disruptive to leave the old code as is in `create_linked_view`

cbkerr:
How shall I test this?

signac/contrib/import_export.py
```diff
@@ -241,6 +241,14 @@ def path_function(job):
             except Exception as error:
                 raise _SchemaPathEvaluationError(error)
 
+        # Check that the user-specified path generates a 1-1 mapping
+        link_check = {j.workspace(): path_function(j) for j in jobs}
+        if len(set(link_check.values())) != len(link_check):
+            raise RuntimeError(
+                "Paths generated with given path function are not 1-1. The easiest "
```
cbkerr, line 248:
```suggestion
                "Links generated with the given path function are not 1-1. The easiest "
```

signac/contrib/import_export.py
```diff
@@ -241,6 +241,14 @@ def path_function(job):
             except Exception as error:
                 raise _SchemaPathEvaluationError(error)
 
+        # Check that the user-specified path generates a 1-1 mapping
+        link_check = {j.workspace(): path_function(j) for j in jobs}
+        if len(set(link_check.values())) != len(link_check):
+            raise RuntimeError(
+                "Links generated with the given path function are not 1-1. The easiest "
+                "way to fix this is to add the job id to the path specification."
+            )
```
bdice, line 250: I suggest the following changes:
- Raise the error for the _first_ failure case rather than evaluating the paths for all jobs before erroring out
- Show the path causing a failure in the error message.
- Use a `ValueError` since it's the user-provided path value (optional, feel free to keep `RuntimeError`)
```suggestion
        links = set()
        for job in jobs:
            job_path = path_function(job)
            if job_path in links:
                raise ValueError(
                    f"Multiple jobs resolve to the same path '{job_path}'. "
                    "The easiest way to fix this is to add the job id to the "
                    "path specification."
                )
            links.add(job_path)
```

vyasr: +1 for using a set, it doesn't look like the dict key matters here. I prefer constructing the entire set at once though (via something like `links = set(path_function(job) for job in jobs)`) rather than erroring on the first duplicate. The main reason is that the error message in the suggestion is potentially misleading: it indicates that there is only a single conflict, when in fact there may (most likely there _will_) be multiple. Even worse, if you have a heterogeneous schema the paths may be underspecified for different reasons. If we really want to provide the most useful diagnostic (but it's a bit more work and I'd understand if you wanted to skip), we should use the for loop, use a flag to indicate that duplicates were found, then log _all_ duplicates using the logger so that the user can inspect the problem. Barring that, I think that constructing the set in one shot and not showing a specific path is probably more appropriate here.

As a bonus, the single-pass approach will be faster in the normal case (no failure). The generator-based construction will be faster than a for loop, you won't need to do a containment check on every add, and the final test is a simple length check

cbkerr: Thanks @bdice @vyasr !
I combined both approaches. Thanks for the advice to use `set()`.

I like being able to see colliding paths, but not by default, in case there are 1000s of them. I put them in the debug log.

I am wondering if it would be faster for most cases as @vyasr points out to first do the simple length check and then only iterate through all jobs if that fails? This would result in duplicated code but could be faster for huge workspaces?

cbkerr:
I changed back to RuntimeError mostly so I don't have to change the existing tests that were looking for that error.

Should I add a test that can check for where the error is caught?

bdice:
It looks like there is also a duplicate check for export: https://github.com/glotzerlab/signac/blob/60646fdf29e5aa28680a6462337ab5a0488bf6c6/signac/contrib/import_export.py#L328-L333

@cbkerr Can you unify these so that we only have one place in views/exports that is used for duplicate checking? The logic should be only implemented once and called in both export/view as a helper function.

signac/contrib/import_export.py
bdice: @vyasr I'm a little unsure about what preference you're expressing in your statement: 
> Barring that, I think that constructing the set in one shot and not showing a specific path is probably more appropriate here.

Now that the "most useful diagnostic" is implemented with a for loop, would you prefer that to the simpler error that does not show a specific path? I think you make good points here and it aligns with what is done for exports: https://github.com/glotzerlab/signac/blob/60646fdf29e5aa28680a6462337ab5a0488bf6c6/signac/contrib/import_export.py#L328-L333

signac/contrib/import_export.py
```diff
@@ -241,6 +241,26 @@ def path_function(job):
...
+                links.add(job_path)
+        num_dups = len(duplicate_links)
+        if num_dups > 0:
+            path_correction = os.path.join("id", "{job.id}")
+            raise RuntimeError(
+                f"The path specification '{path}' would result in {num_dups} duplicate "
+                "links. See the debug log for the list. The easiest way to fix "
+                f"this is to append the job id to the path specification like '{path_correction}'."
+            )
+        logger.info(f"Path specification {path} uniquely maps workspaces.")
```
bdice, line 262: In this implementation, I have a few comments. I am not sure if this is my favored implementation after reading Vyas's comment. See https://github.com/glotzerlab/signac/pull/642/#discussion_r752313976
1. I don't think `duplicate_links` is necessary to track. I would report all duplicates regardless of how many times they occur -- that feels more like canonical behavior for UNIX command-line software.
2. It is probably not necessary to report the number of duplicates. The user can't do much with that information.
3. I don't think it is necessary to log the info message on success.
```suggestion
        # Check that the user-specified path generates a 1-1 mapping
        links = set()
        for job in jobs:
            job_path = path_function(job)
            if job_path in links:
                logger.debug(f"Generated path '{job_path}' is not unique.")
            else:
                links.add(job_path)
        if len(links) != len(jobs):
            path_correction = 
            raise RuntimeError(
                f"The path specification '{path}' would result in duplicate "
                "links. See the debug log for the list. The easiest "
                "way to fix this is to append the job id to the path "
                f"specification like '{os.path.join("id", "{job.id}")}'."
            )
```

bdice:
> Should I add a test that can check for where the error is caught?

Yes, we add tests for all error cases that we can catch. You can model it on this test: https://github.com/glotzerlab/signac/blob/4a4bfde3bcdfb31dc370116190092582e0ee7248/tests/test_project.py#L1131-L1138

cbkerr:
> It looks like there is also a duplicate check for export:
> 
> https://github.com/glotzerlab/signac/blob/60646fdf29e5aa28680a6462337ab5a0488bf6c6/signac/contrib/import_export.py#L328-L333
> 
> @cbkerr Can you unify these so that we only have one place in views/exports that is used for duplicate checking? The logic should be only implemented once and called in both export/view as a helper function.

Yes, I structured my initial check based on this one. However, I don't think we can remove the RunTime error from `_export_jobs` in case it is passed a callable path function that would result in the error.

Do you mean making a new function `_check_path_function(jobs, path_function)` kind of like we have for the existing function `_check_directory_structure_validity`? (and calling this function in both places)

signac/contrib/import_export.py
vyasr: @bdice sorry for the confusion, let me clarify. My ordered prefs were:
1. Check if lengths mismatch, report all unmatched paths in the debug log via `logger.debug(...)` and then raise an exception
2. Check if lengths mismatch and raise an error without specifying paths
3. Loop through and find all mismatches, then raise an error with all of them if any exist

As far as I can tell, what is currently implemented is option (2), which is the simpler case that you found also in `import_export.py`. @cbkerr I'm confused by your last message:

> I like being able to see colliding paths, but not by default, in case there are 1000s of them. I put them in the debug log.
What did you put in the debug log? I don't see any logging right now.

> I am wondering if it would be faster for most cases as @vyasr points out to first do the simple length check and then only iterate through all jobs if that fails? This would result in duplicated code but could be faster for huge workspaces?
The length check is already what you're doing, right? You just aren't doing the iteration at all?

cbkerr: Agree that what's released is option 2. The code I added is option 1 except that it does not do the simple length check first.

> I do think that the best option is to start with a length check first.

I think you're agreeing with what I was asking about here:
> to first do the simple length check and then only iterate through all jobs if that fails

vyasr: Yes, I am agreeing with that.

vyasr:
@cbkerr just making sure, this PR is just waiting on you for edits right? Or do you need more feedback from me right now? Feel free to ping me if I'm leaving you hanging.

cbkerr:
I implemented the helper function, but to provide the user with the specified path in the error message, it seems hacky to also pass the "path" argument into the helper function. I do think it would be nice to have that info.

cbkerr:
Would appreciate your look again @vyasr. I use your suggestion of the simple check first.

vyasr:
> I implemented the helper function, but to provide the user with the specified path in the error message, it seems hacky to also pass the "path" argument into the helper function. I do think it would be nice to have that info.

I don't think it's too hacky, but I see your point. I think it's slightly awkward because we're calling a validation function that raises an exception rather than returning a boolean flag that can be used to raise an exception at the call site. However, since you're reusing this function twice I don't think you want to reproduce the exception in both places, so I don't mind it the way it is. Since you are providing the path spec, though, one thing that might be an improvement on the current error message is to actually show the full path spec that you're recommending as an alternative. Specifically, replace `{os.path.join('id', '{job.id}')}` with `{os.path.join(path_spec, 'id', '{job.id}')}` or so. What do you think?

> Would appreciate your look again @vyasr. I use your suggestion of the simple check first.

Yup the check looks good to me now.

vyasr (commented):
I wonder if `_make_path_function` and `_check_path_function` should really be moved into contrib/utility.py instead of living in `import_export.py` and being imported into `linked_view.py`. The dependency that implies seems off to me, what do you think?

cbkerr:
I added a test in `test_create_linked_view_homogeneous_schema_nested_provide_partial_path` and confirmed that without the changes in this PR, the test fails, _i.e._ the code fails to generate the error.

I'm having trouble adding a test to `test_shell` even following the same format as the others that are looking for generated error messages.
My problem is that the output of `err = self.call ...` is an empty string. Any pointers on this?

tests/test_shell.py
```diff
@@ -235,6 +235,19 @@ def test_view(self):
...
+    def test_view_incomplete_path_spec(self):
+        self.call("python -m signac init my_project".split())
+        project = signac.Project()
+        sps = [{"a": i} for i in range(3)]
+        for sp in sps:
+            project.open_job(sp).init()
+        os.mkdir("view")
+
+        # Should error if user-provided path doesn't make 1-1 mapping
+        err = self.call("python -m signac view non_unique".split(), error=True)
```
cbkerr, line 248: This is the code that is not working as expected

csadorf: The `error` parameters make it so that the STDERR instead of the STDOUT stream is returned. You probably meant to use `raise_error`, see line 68.

cbkerr: I left a breakpoint in the call function and `out` and `err` from `p.communicate()` are both empty strings and `p.returncode` is 0. Does that mean the process doesn't think there is an error? When I run the command on the same terminal I'm testing from I get the error as expected.

bdice:
```
$ python -m signac view --help
usage: __main__.py view [-h] [-f FILTER [FILTER ...]]
                        [-d DOC_FILTER [DOC_FILTER ...]]
                        [-j JOB_ID [JOB_ID ...]] [-i INDEX]
                        [prefix] [path]

Generate a human readable set of paths representing state points in the
workspace, e.g.
view/param_name_1/param_value_1/param_name_2/param_value_2/job. The leaf nodes
of this directory structure are symlinks (named "job") into the workspace
directory for that parameter combination. Note that all positional arguments
must be provided before any keyword arguments. In particular, the prefix and
path must be specified before arguments such as the filters, e.g. signac view
$PREFIX $VIEW_PATH -f FILTERS -d DOC_FILTERS.

positional arguments:
  prefix                The path where the view is to be created. Defaults to
                        view.
  path                  The path used for the generation of the linked view
                        hierarchy, defaults to '{{auto}}' (see
                        Project.export_to for information on how this is
                        expanded).

optional arguments:
  -h, --help            show this help message and exit

select:
  -f FILTER [FILTER ...], --filter FILTER [FILTER ...]
                        Limit the view to jobs matching this state point
                        filter.
  -d DOC_FILTER [DOC_FILTER ...], --doc-filter DOC_FILTER [DOC_FILTER ...]
                        Limit the view to jobs matching this document filter.
  -j JOB_ID [JOB_ID ...], --job-id JOB_ID [JOB_ID ...]
                        Limit the view to jobs with these job ids.
  -i INDEX, --index INDEX
                        The filename of an index file.
```

This test is using `non_unique` as the _prefix_, not the _path_. You need to give `view` as the prefix and `non_unique` as the path:
```suggestion
        err = self.call("python -m signac view view non_unique".split(), error=True)
```

bdice: @csadorf is also correct. My suggestion needs the `raise_error` flag:
```suggestion
        err = self.call(
            "python -m signac view view non_unique".split(),
            error=True,
            raise_error=False,
        )
```

I'm going to commit this change so we can get CI passing.

signac/contrib/utility.py
```diff
@@ -479,3 +481,255 @@ def _nested_dicts_to_dotted_keys(t, encode=_encode_tree, key=None):
...
+            except Exception as error:
+                raise _SchemaPathEvaluationError(error)
+
+        # Check that the user-specified path generates a 1-1 mapping
+        _check_path_function(jobs, path_spec=path, path_function=path_function)
+
+    else:
+        raise ValueError(
+            "The path argument must either be `None`, `False`, or of type `str`."
+        )
```
bdice, line 733: This disagrees with the docstring that says it accepts a `callable`. Based on my suggestion above, we should probably let this function pass through a callable and validate all possible path function values with `_check_path_function` at the end before returning.

signac/contrib/import_export.py
```diff
@@ -293,25 +84,18 @@ def _export_jobs(jobs, path, copytree):
...
-        If paths generated with given path function are not unique.
-
     """
     # Transform the path argument into a callable if necessary.
     if callable(path):
         path_function = path
+        _check_path_function(jobs, path_spec=path, path_function=path_function)
     else:
         path_function = _make_path_function(jobs, path)
+        # path_function is checked inside _make_path_function
```
bdice, line 94: Seems like `_make_path_function` can be used in place of the logic here. We should be able to avoid having two paths (`if callable(path)` in the snippet above) that independently call `_check_path_function`, since `_make_path_function` can take a callable. Might need to refactor `_make_path_function` to validate at the end. See other comment.
```suggestion
    path_function = _make_path_function(jobs, path)
```

signac/contrib/import_export.py
```diff
@@ -293,25 +84,18 @@ def _export_jobs(jobs, path, copytree):
...
 
     # Determine export path for each job.
     paths = {job.workspace(): path_function(job) for job in jobs}
```
bdice, line 97: We're already computing these paths when validating for uniqueness. Should that function return the generated paths to avoid doing the same computation twice?

signac/contrib/utility.py
```diff
@@ -479,3 +481,255 @@ def _nested_dicts_to_dotted_keys(t, encode=_encode_tree, key=None):
...
+    Returns
+    -------
+    callable
+        Function that returns a normalized path.
+
+    """
+    from .schema import _build_job_statepoint_index
+
+    if len(jobs) <= 1:
+        # The lambda must (optionally) take a format spec argument to match the
```
bdice, line 507: "must (optionally)" seems inconsistent. It's either required or it's not.

signac/contrib/utility.py
```diff
@@ -479,3 +481,255 @@ def _nested_dicts_to_dotted_keys(t, encode=_encode_tree, key=None):
...
+        if isinstance(value, Job):
+            return self.paths(value, format_spec)
+        else:
+            return super().format_field(value, format_spec)
+
+
+class _SchemaPathEvaluationError(RuntimeError):
+    """Raised for errors in schema path evaluation."""
+
+    pass
```
bdice, line 595: This needs to be a public class (no underscore) defined in `signac/contrib/errors.py`, because it can be raised to the user.

signac/contrib/utility.py
```diff
@@ -479,3 +481,255 @@ def _nested_dicts_to_dotted_keys(t, encode=_encode_tree, key=None):
...
+    if len(links) != len(jobs):
+        # report all mismatches
+        links = set()
+        for job in jobs:
+            job_path = path_function(job)
+            if job_path in links:
+                logger.debug(f"Generated path '{job_path}' is not unique.")
+            else:
+                links.add(job_path)
+        raise RuntimeError(
```
bdice, line 627: Should this be a `SchemaPathEvaluationError`?

bdice (changes requested):
I made a brief pass of review and have some high-level suggestions.

tests/test_shell.py
cbkerr: Argh, I knew it was something obvious! Thank you

bdice: Not obvious! I think this is an annoying CLI design choice to require the prefix before specifying the path. I always make this mistake and forget to write `signac view view path` (because that sounds silly). Could be fixed in signac 2.0 if we want. Filing an issue.

bdice: I filed #653 for this. I think it's a flaw in the CLI design.

csadorf: > I think this is an annoying CLI design choice to require the prefix before specifying the path.

Sorry. 😁

b-butler:
Given the large number of reviews, I am removing myself.

lyrivera:
I'm following @b-butler and removing myself.

vyasr:
@cbkerr what's the plan here?

cbkerr:
I've extracted enough to be a bug fix into #666. It should be easier to review without moving lots of code around.
I'll rename this PR or make an issue to focus on the refactor.

vyasr:
@cbkerr awesome. If you want to merge master into this branch and fix conflicts, I would be fine with repurposing this PR for the refactor. The conversation here is illuminating and worth preserving next to the final set of changes IMO.

vyasr:
@cbkerr do you need help with next steps on this PR?

b-butler:
@cbkerr do you still plan on getting back to this?

vyasr:
@cbkerr checking in here again, do you still plan to tackle this? Since the bug fix was merged in #666 and this is just a refactor, can the changes be retargeted onto `next` for version 2.0 instead of 1.8?

vyasr:
@cbkerr do you still think this is work worth doing, and just not something that you'll get around to, or is this something we can skip altogether? If you think it's worthwhile, could you write up an issue describing what should be done in case someone wants to pick this up in the future?

PR #643, opened 2021-11-09
Remove project id

### LINKED ISSUE #541: Removing signac project id
I've been meaning to raise this question for a while, and in anticipation of resolving #527 I think now is a good time to consider it: should we remove the project name in signac schema version 2? Currently we use it sparingly, pretty much just to provide a string representation of a project and in associated functions like naming a temporary project. signac-flow uses this to generate the unique submission id, but the fact that a project name is not guaranteed to be unique anyway forced us to add the project directory to the submission id hash anyway. We can always introduce alternative methods in signac-flow's FlowProject to give more meaningfully named submissions, if the main value is that a submitted cluster job can be associated with a project based on the name. Is there any other meaningful upside to keeping this, rather than just falling back to the name of the current directory? I think it's worth considering as part of the shift, especially since we're removing workspace configurability.

vyasr:
## Description

Going forward projects will be entirely identified by their root directory.

**EDIT**
The associated changes have become significantly broader, so I'm going to keep a running list here that can be cleaned up before merge so that the git log is informative. Things that are modified or will be modified by this PR:
- [x] Projects no longer have a name and are identified solely by their root directory.
- [x] `signac.init_project` and `signac.Project.init_project` no longer accept a `name` parameter. Additionally, all parameters are now keyword-only. Calling `init_project` with only a single positional argument is assumed to be a holdover from a version <2.0 and raises a `FutureWarning` indicating that names are no longer supported. 
- [x] Project configs are now stored in `.signac/config`
- [x] Project discovery now only depends on the presence of the `.signac` directory. As a result, there is no longer a merging of multiple signac configs, only the user's global `.signacrc` and the project-local `.signac/config`.
- [x] The Project constructor now just requires the root directory. It does not initialize an uninitialized Project, however, so `init_project` must still be called first or an error will be raised. Furthermore, no discovery is performed, so `get_project` is still the preferred method for acquiring a project handle by traversing up the folder hierarchy.
- [x] Config APIs have been simplified substantially. `get_config` is removed in favor of always using `read_config_file`, various module-level variables have been removed, and most config file logic outside the config module is now calling the config APIs rather than duplicating logic.

## Motivation and Context

Resolves #541

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:


tommy-waltmann:
Is this working towards a full v2 release for the entire project, or is the schema version just changing from a v1 to a v2?

vyasr:
> Is this working towards a full v2 release for the entire project, or is the schema version just changing from a v1 to a v2?

This PR is merging into a temporary branch `schema2` where I'll be accumulating all changes associated with a v1->v2 schema version bump. Eventually `schema2` will get merged into `next`, and `next` will only get merged into `master` when we're ready for the signac 2.0.0 release. So bumping to schema version 2 will coincide with the 2.0.0 release in this case, although we don't require that a schema version bump should coincide with a major release in general.

tommy-waltmann (approved):

vyasr:
This PR is not ready for merge yet. There's on outstanding issue that I could patch here, but I think it is symptomatic of some incorrect issues in our migration code and I plan to make a PR to master to resolve that first.

tommy-waltmann:
I approved because I didn't have any comments on the diff in this PR. Of course, fix the failing tests before merging.

vyasr:
> I approved because I didn't have any comments on the diff in this PR. Of course, fix the failing tests before merging.

Yep no worries, I just wanted to give you some context in case you came back a few days later and wondered why there hadn't been any attempt to fix the failing tests yet.

signac/contrib/project.py
```diff
@@ -2193,17 +2158,15 @@ def _repr_html_(self):
         return repr(self) + self._repr_html_jobs()
 
 
-def init_project(name, root=None, workspace=None, make_dir=True):
-    """Initialize a project with the given name.
+def init_project(root=None, workspace=None, make_dir=True):
```
bdice, line 2161: Since we're already breaking this API, I might suggest requiring keyword arguments here. We're going to have lots of signac 1.x users get confused when they try to run `signac.init_project(name)` and `name` is interpreted as a root directory. Another option would be to catch positional arguments as `*args` and raise a user-friendly error if any are provided, which could help guide users towards 2.x syntax.
```suggestion
def init_project(*, root=None, workspace=None, make_dir=True):
```

bdice (changes requested):
I have attached a first round of comments. I did not read the changes to tests very thoroughly, since I suggested some behavior changes that could affect tests.

signac/contrib/migration/__init__.py
```diff
@@ -31,11 +33,13 @@ def _reload_project_config(project):
...
-        config = get_config(project.fn(fn))
-        if "project" in config:
-            break
-    else:
+        project_fn = project.fn(fn)
+        if os.path.isfile(project_fn):
+            config = get_config(project_fn)
+            found = True
+    if not found:
         raise RuntimeError("Unable to determine project configuration file.")
```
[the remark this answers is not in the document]
vyasr: Good call, I always forget that `for/else` exists.

signac/contrib/migration/v1_to_v2.py
```diff
@@ -0,0 +1,20 @@
...
+"""
+
+import os
+
+from ...common.config import get_config
+
+
+def migrate_v1_to_v2(project):
+    """Migrate from schema version 1 to version 2."""
+    config = get_config(os.path.join(project.root_directory(), "signac.rc"))
```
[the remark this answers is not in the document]
vyasr: It is handled in the calling code, yes. Is there any reason for the migration methods to be user-facing? IMO everything other than the main entry point `apply_migrations` should be internal, but I could be missing some use case.

signac/contrib/project.py
```diff
@@ -194,7 +192,6 @@ def _repr_html_(self):
         """
         return (
             "<p>"
-            + f"<strong>Project:</strong> {self.id}<br>"
             + f"<strong>Root:</strong> {self.root_directory()}<br>"
```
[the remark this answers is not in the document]
vyasr: Sure I'd be fine with that.

signac/contrib/project.py
```diff
@@ -176,7 +173,8 @@ def __init__(self, config=None, _ignore_schema_version=False):
 
     def __str__(self):
         """Return the project's id."""
-        return str(self.id)
+        # TODO: Is this the string representation that we want?
+        return os.path.split(self._root_directory)[-1]
```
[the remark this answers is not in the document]
vyasr: I think maybe `Project(self.root_directory())` would be good. The `__repr__` for the class uses `get_project`, which I find a bit strange. While we don't tell users to instantiate projects directly, I'm not sure that it makes sense to look like that in the repr. I'd like the repr and the str to match up relatively closely here. Thoughts?

signac/contrib/project.py
```diff
@@ -255,18 +252,6 @@ def workspace(self):
         """
         return self._workspace
 
-    @property
-    def id(self):
-        """Get the project identifier.
```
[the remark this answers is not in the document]
vyasr: Personally I'd like to avoid the redundancy, especially because depending on how users make use of the property it could lead to subtle bugs if we change the meaning of the property without removing it. For instance, it wouldn't surprise me if a user was making use of the id to customize filenames or something. Obviously not the intended use case, but anything is possible. That said, I agree that removing it could be quite disruptive. I'd be fine with following the precedent we set for `doc_filter` and mark this as a deprecation to be removed in signac 3.0, redefining it as an alias of `root_directory` during 2.0. Thoughts?

signac/contrib/project.py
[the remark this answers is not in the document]
bdice: The invariant I sometimes advocate for is `eval(repr(obj)) == obj`. We use this in freud, for instance, to ensure that constructors align with the object representation. I recognize it is more awkward for signac because projects are constructed via factory methods. I think the repr is important and would like other opinions on this.

signac/contrib/project.py
vyasr: Yeah this is going to be a problem. Unlike with the `doc_filter` parameter I mentioned above, we can't provide backwards compatibility layers here because `name` is a positional parameter and it comes first in the list. I cannot think of any solution to that. Making the other parameters keyword-only is a reasonable alternative since I suspect that very few people bother to customize any of the other parameters. We could always remove that behavior in 3.0 after users have had sufficient time to accustom themselves to the new behavior.

signac/contrib/project.py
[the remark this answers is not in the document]
bdice: I favor making a decision in 2.0 rather than deferring to 3.0, but which approach to take (removing id vs. redefining it) is not clear to me. I’d like others’ thoughts here.

signac/contrib/project.py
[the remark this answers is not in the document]
vyasr: Exactly, the use of factories make that invariant seem a little awkward here.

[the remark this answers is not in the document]
csadorf: I explicitly implemented the `__repr__()` to use the factory function to ensure the aforementioned invariant. I would advocate to change the constructor signature with all the other breaking changes for 2.0 and construct a `Project` directly from the root directory. That should be sufficient, considering that it is **the** unique identifier. User should still use `get_project()` for project discovery.

With that in mind I would suggest the following:
```python
>>>> str(project)
'<signac.Project at "." with 42 jobs>'
>>> repr(project)
'Project("/absolute/path/to/project")'
```
What do you think?

signac/contrib/project.py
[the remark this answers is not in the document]
csadorf: If the uniquely identifying feature is the path (similar to a SQlite database), then that should be it, no redundancy. In so far I would advocate for removing `id` with 2.0.

signac/contrib/project.py
csadorf: Agreed, this one will be tricky, especially because this function is used [everywhere](https://github.com/search?p=2&q=signac.init_project&type=Code).

Maybe, we could be even more gentle and just raise a deprecation warning when the function is called with only one argument and ignore it?

tests/test_project.py
```diff
@@ -720,8 +708,8 @@ def test_JobsCursor_contains(self):
 
     def test_job_move(self):
         root = self._tmp_dir.name
-        project_a = signac.init_project("ProjectA", os.path.join(root, "a"))
-        project_b = signac.init_project("ProjectB", os.path.join(root, "b"))
+        project_a = signac.init_project(os.path.join(root, "a"))
+        project_b = signac.init_project(os.path.join(root, "b"))
```
csadorf, line 712: I would recommend to keep at least a few tests that test the legacy behavior until we have made a decision on how to deal with backwards-compatibility for this particular function. It is too widely used to just break. See discussion above.

csadorf (commented):
I needed to review the PR to have enough context to be able to contribute to the discussions in a meaningful manner.

signac/contrib/project.py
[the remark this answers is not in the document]
vyasr: Yes I like that idea. @bdice any objections?

signac/contrib/project.py
vyasr: So this:
```
def init_project(*args, root=None, workspace=None, make_dir=True):
    if args:
        if len(args) == 1:
            warnings.warn(f"Project names were removed in signac 2.0. If you intended to call `init_project` with a root directory, please provided it as a keyword argument: `init_project(root={args[0]})`")
        else:
            # Match the usual error from misusing keyword-only args
            raise TypeError(f"init_project() takes 0 positional arguments but {len(args)} was given")

File: signac/common/config.py [hunk 8]
```diff
@@ -84,22 +83,24 @@ def load_config(root=None, local=False):
...
-        for fn in _search_tree(root):
-            tmp = read_config_file(fn)
-            config.merge(tmp)
-            if "project" in tmp:
-                config["project_dir"] = os.path.dirname(fn)
-                break
+        search_func = _search_tree
+    else:
+        # Local searches cannot proceed up the tree.
+        search_func = _search_local
```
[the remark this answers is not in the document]
vyasr: No reason. I wrote the new conditional separately since the body of the old conditional is now just the for loop below, so I didn't register the switch. Will switch back.

signac/contrib/migration/__init__.py
```diff
@@ -31,11 +33,13 @@ def _reload_project_config(project):
...
+    found = False
     for fn in ("signac.rc", ".signacrc"):
-        config = get_config(project.fn(fn))
-        if "project" in config:
-            break
-    else:
+        project_fn = project.fn(fn)
+        if os.path.isfile(project_fn):
+            config = get_config(project_fn)
+            found = True
```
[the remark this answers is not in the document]
vyasr: Yes, this is the definition that is now present. I am open to alternate suggestions. What sort of use cases would be broken by this definition? I agree that it is weaker than the current definition, but is that necessarily a problem. This is closer to how git (our usual 0th order inspiration) behaves. What configuration information might it be important to include for different subdirectories of a project? If we're thinking about e.g. different FlowProjects, we have previously discussed namespacing flow more precisely around FlowProjects, for example.

tests/test_project.py
vyasr: I changed all the tests to make sure that I didn't miss anything in my first round of changes because I wanted reviewers to be able to see the full scope of the change. I won't bother modifying any more until we've decided on the correct semantics.

signac/contrib/project.py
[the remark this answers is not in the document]
mikemhenry: I also think we should just remove `id` in 2.0. We could always post in our #general channel and see if there are any users that really need `id` but it might be better to start the notice that it will be removed now and see if anyone asks about it, we could in the message even direct people to file and issue if they need it and then we can discuss it.

signac/contrib/project.py
csadorf: Yes, I think so. We might want to consider the warning type.

signac/contrib/migration/__init__.py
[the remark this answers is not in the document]
csadorf: This makes me very uneasy and I think we need to have a discussion about this.

signac/contrib/migration/v1_to_v2.py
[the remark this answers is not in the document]
vyasr: I'll be making the change to move this to an internal method in a separate PR. With that context, no need to add an extra check here.

signac/contrib/project.py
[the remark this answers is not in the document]
vyasr: Taking these comments as a vote to move forward with removal. We can gauge community opinions like Mike suggested, and we'll also introduce a deprecation warning in the 1.x line in #644 that should warn users loudly.

signac/contrib/project.py
vyasr: That seems like a good solution to me. The warning type should probably be a FutureWarning since this is effectively a deprecation that we're pushing to 3.0.

signac/contrib/migration/__init__.py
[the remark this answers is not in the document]
vyasr: Sure, I'll ping you offline.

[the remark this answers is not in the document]
vyasr: Closing the loop: @csadorf and I discussed this offline and concluded that the switch planned in #197 will address the problem. Addressing that will move the definition from the existence of a config file to the existence of a `.signac` directory, which is a bit stronger because it avoids edge cases such as treating any directory for centralized config file storage as a signac project. That change may or may not occur in this PR, but it will happen on the `schema2` branch before we merge it into `next`.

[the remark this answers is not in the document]
bdice: By "centralized config file storage", do you mean something like `$HOME/.signacrc`? I don't think we've discussed centralized configuration like that in #197 yet, which feels like a loose end we should clarify in that issue. I would guess that we don't want to use a `$HOME/.signac` directory for centralized configuration because it would make the home directory into a signac project, from my understanding of the new definition proposed in #197. I'm happy with pretty much any solution here but would like to hear your thoughts, @vyasr @csadorf.

signac/contrib/project.py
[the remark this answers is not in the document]
vyasr: One related question here that we should think about is whether the project constructor would have the semantics of `get_project` or `init_project`. Will the construct create a project if one does not already exist in a directory?

signac/contrib/migration/__init__.py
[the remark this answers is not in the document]
csadorf: @bdice That's exactly the point. I don't think we plan on abolishing centralized configuration, so with the definition introduced here, the home directory (or any other suitable configuration directly like `~/.config/io.signac/` or `~/Library/Application Support/io.signac/`) would be a signac project which just feels wrong to me. So instead, following the git-approach, any directory that contains a `.signac` sub-directory would be a signac project root directory.

signac/contrib/project.py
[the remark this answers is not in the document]
csadorf: Strongly against including discovery with the constructor, that just leads to messy code and also breaks the invariant from above. Is `Project('/path/to/project/sub-dir', discover=True) == Project('/path/to/project')`?

I'd also caution against side-effects on object construction. We had this for jobs initially and thankfully moved away from that. I see two viable approaches:
1. We require the Project to "exist" before we can obtain the corresponding object for it. So, `Project(/path/to/project')` would raise an exception in case the `/paht/to/project/.signac` does not exist.
2. We don't require the project to exist, and create it on the first write-operation. That's similar to how jobs work: `Project('/path/to/project').init()`.
I lean toward approach 2.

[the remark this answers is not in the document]
vyasr: I agree with no discovery. At first glance I like approach 2 as well, although I'd have to think about the other implications. Since our plan is to tighten the job/project linkage, we'd probably also want things like job creation to imply project creation and so on. There would likely be additional such implied behaviors.

signac/__main__.py
```diff
@@ -909,7 +904,7 @@ def write_history_file():
                 banner=SHELL_BANNER.format(
```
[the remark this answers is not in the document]
vyasr: For now we're only doing this in two places so I'm comfortable leaving this as is. We can revisit if we add more specialized shells/banners in the future.

signac/__main__.py
```diff
@@ -718,16 +714,12 @@ def main_config_show(args):
...
-                else:
-                    cfg.merge(config.read_config_file(fn))
+        if os.path.isfile(config.PROJECT_CONFIG_FN):
+            cfg = config.read_config_file(config.PROJECT_CONFIG_FN)
     elif args.globalcfg:
-        cfg = config.read_config_file(config.FN_CONFIG)
+        cfg = config.read_config_file(config.USER_CONFIG_FN)
     else:
-        cfg = config.load_config()
+        _, cfg = config.load_config()
```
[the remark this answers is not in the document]
bdice: From a directory with no signac project config (and no global config), I execute `signac config verify`.
- On `master` and `next`, I get the output `project : Possibly invalid or missing.`
- On this branch, I get the output `Passed.` That's a regression.

I caught this because I saw the codecov was missing tests for `signac config verify`. Thanks codecov!

signac/__main__.py
```diff
@@ -757,16 +749,12 @@ def main_config_verify(args):
...
         raise ValueError("You can specify either -l/--local or -g/--global, not both.")
     elif args.local:
-        for fn in config.CONFIG_FILENAMES:
-            if os.path.isfile(fn):
-                if cfg is None:
-                    cfg = config.read_config_file(fn)
-                else:
-                    cfg.merge(config.read_config_file(fn))
+        if os.path.isfile(config.PROJECT_CONFIG_FN):
+            cfg = config.read_config_file(config.PROJECT_CONFIG_FN)
```
bdice, line 753: We need to add tests for `signac config verify` - see bug mentioned above.

signac/common/config.py
```diff
@@ -72,35 +93,52 @@ def read_config_file(filename):
...
-    return Config(infile, configspec=configspec, *args, **kwargs)
-
-
 def load_config(root=None, local=False):
-    """Load configuration, searching upward from a root path."""
+    """Load configuration, searching upward from a root path if desired.
+
+    Parameters
+    ----------
+    root : str
```
bdice, line 101:
```suggestion
    root : str, optional
```

signac/common/config.py
```diff
@@ -72,35 +93,52 @@ def read_config_file(filename):
...
+        tmp = read_config_file(fn)
+        config.merge(tmp)
+        # Once a valid config file is found, we cease looking any further, i.e.
+        # we assume that the first directory with a valid config file is the
+        # project root.
+        # TODO: Is this a sufficiently portable solution? We could use pathlib's
+        # `.paths` attribute, but that has significant performance
+        # implications: this takes ~200 ns, whereas
+        # os.path.join(*(pathlib.PosixPath(fn).parts[:-2]))
+        # takes ~6 us, almost a 20x slowdown.
```
bdice, line 138: I'm fine with this. Just rewrite the TODO as an explanation of behavior and design choice, rather than as a question.

signac/contrib/migration/v1_to_v2.py
```diff
@@ -0,0 +1,18 @@
...
+"""
+
+from ...common.config import _get_project_config_fn, read_config_file
+
+
+def migrate_v1_to_v2(project):
+    """Migrate from schema version 1 to version 2."""
+    config = read_config_file(_get_project_config_fn(project.root_directory()))
+    config["schema_version"] = 2
+    del config["project"]
```
bdice, line 17: Is there a benefit to deleting the `"project"` key? Extraneous keys in the config are just ignored if I recall correctly, so I'd think we could leave it without deleting user data. We still need a schema version bump to indicate the change in how the config is interpreted.

signac/contrib/project.py
```diff
@@ -85,15 +89,16 @@ def __setitem__(self, key, value):
 class Project:
     """The handle on a signac project.
 
-    Application developers should usually not need to
-    directly instantiate this class, but use
-    :meth:`~signac.get_project` instead.
+    A :class:`Project` may only be constructed in a directory that is already a
+    signac project, i.e. a directory in which :meth:`~signac.init_project` has
```
bdice, line 93: This might need `:func:` instead of `:meth:`?
```suggestion
    signac project, i.e. a directory in which :func:`~signac.init_project` has
```

signac/contrib/project.py
```diff
@@ -85,15 +89,16 @@ def __setitem__(self, key, value):
...
+    already been run. If project discovery (searching upwards in the folder
+    hierarchy until a project root is discovery) is desired, users should
```
bdice, line 95:
```suggestion
    hierarchy until a project root is discovered) is desired, users should
```

signac/contrib/project.py
```diff
@@ -85,15 +89,16 @@ def __setitem__(self, key, value):
...
+    instead invoke :func:`~signac.get_project` or :meth:`Project.get_project`.
 
     Parameters
     ----------
-    config :
-        The project configuration to use. By default, it loads the first signac
-        project configuration found while searching upward from the current
+    root : str
```
bdice, line 100:
```suggestion
    root : str, optional
```

signac/contrib/project.py
```diff
@@ -113,26 +118,26 @@ class Project:
...
 
-        # Ensure that the project id is configured.
-        try:
-            self._id = str(self.config["project"])
-        except KeyError:
+        # Ensure that the project is configured.
+        if root_dir is None:
             raise LookupError(
-                "Unable to determine project id. "
+                "Specified project root does not contain a configuration. "
```
bdice, line 132:
```suggestion
                "Specified root does not contain a project configuration. "
```

signac/contrib/project.py
```diff
@@ -176,12 +181,10 @@ def __init__(self, config=None, _ignore_schema_version=False):
...
     def __str__(self):
         """Return the project's id."""
-        return str(self.id)
+        return repr(self)
 
     def __repr__(self):
-        return "{type}.get_project({root})".format(
-            type=self.__class__.__name__, root=repr(self.root_directory())
-        )
+        return f'{self.__class__.__name__}("{self.root_directory()}")'
```
bdice, line 187: Woop, this is a regression. This must call `repr(self.root_directory())` to add its own quotes and get the escaping correct for arbitrary paths. Then double-quotes are permitted around the outside f-string, required by style fixers when double quotes are not contained in the string.
```suggestion
        return f"{self.__class__.__name__}({repr(self.root_directory())})"
```

tests/test_shell.py
```diff
@@ -49,6 +49,7 @@ def setUp(self, request):
             pythonpath = [os.getcwd()] + pythonpath.split(":")
         os.environ["PYTHONPATH"] = ":".join(pythonpath)
         self.tmpdir = TemporaryDirectory(prefix="signac_")
+        self.project_name = f'Project("{os.path.realpath(self.tmpdir.name)}")'
```
bdice, line 52:
```suggestion
        self.project_name = f"Project({repr(os.path.realpath(self.tmpdir.name))})"
```

bdice (changes requested):
Another round of feedback. I identified a problem/regression in `signac/common/config.py`'s `load_config` that needs to be fixed and tested. I also added some bug fixes via suggestions.

signac/common/config.py
```diff
@@ -12,18 +12,19 @@
...
 logger = logging.getLogger(__name__)
 
-DEFAULT_FILENAME = ".signacrc"
-CONFIG_FILENAMES = [DEFAULT_FILENAME, "signac.rc"]
-HOME = os.path.expanduser("~")
-CONFIG_PATH = [HOME]
-FN_CONFIG = os.path.expanduser("~/.signacrc")
+# TODO: Should these variables be internal or do we want them exposed?
+PROJECT_CONFIG_FN = os.path.join(".signac", "config")
+USER_CONFIG_FN = os.path.expanduser("~/.signacrc")
```
csadorf, line 17: Pendantic:
```suggestion
USER_CONFIG_FN = os.path.expanduser(os.path.join("~", ".signacrc"))
```

signac/common/config.py
```diff
@@ -12,18 +12,19 @@
 
 logger = logging.getLogger(__name__)
 
-DEFAULT_FILENAME = ".signacrc"
-CONFIG_FILENAMES = [DEFAULT_FILENAME, "signac.rc"]
-HOME = os.path.expanduser("~")
-CONFIG_PATH = [HOME]
-FN_CONFIG = os.path.expanduser("~/.signacrc")
+# TODO: Should these variables be internal or do we want them exposed?
```
csadorf, line 15: Since this should be documented behavior anyways I see no issue with exposing them.

signac/contrib/migration/v1_to_v2.py
csadorf: Why leave it? Wouldn't that be confusing?

signac/common/config.py
bdice: I can confirm this is needed on Windows:
```python
>>> os.path.expanduser("~/.signacrc")
'C:\\Users\\bdice/.signacrc'
>>> os.path.expanduser(os.path.join("~", ".signacrc"))
'C:\\Users\\bdice\\.signacrc'
```

As a note, these lines will execute any time the package is imported. We want to minimize the syscall requirements to keep fast import times (all the `os` functions here must perform syscalls, to my knowledge). It may be more appropriate to define anything calling `os.path.join` or `os.path.expanduser` in a function that is called when needed, rather than as a module constant evaluated at import time. It looks like these module constants are used O(1) times and not O(N) times, so it's not paying off the cost incurred at import time, only shifting a cost from usage-time to import-time.

signac/contrib/migration/v1_to_v2.py
bdice: 🤷‍♂️ I don't have a strong opinion on this. We can remove that key for clarity - I just wanted to check because we wouldn't strictly _need_ to delete this key of the "user data" as a part of a migration.

signac/common/config.py
csadorf: > I can confirm this is needed on Windows:

I had a hunch that this might be an issue on Windows, but figured that it wasn't an actual bug, because it had not been an issue until now. 😕

signac/contrib/migration/v1_to_v2.py
vyasr: I think we should remove it. Yes, extraneous keys would be ignored, but why would you want to leave the clutter of something from a previous schema version? And in any case, part of the logic of the version 2 schema is to separate the config completely from user data (which is why we're moving signac.rc to the hidden .signac/config).

vyasr:
I'm waiting to address these comments until after #654 is merged and this can be updated with that.

signac/contrib/migration/v1_to_v2.py
bdice: Cool. Let's remove it.

csadorf: Ok, sorry to piss on the parade, but thinking about it again, the only reservation I have is that some users might have put crucial (provenance) metadata into the value of this key. 😟  Thoughts?

bdice: That was my primary concern: users might have named one project `SelfAssembly-Shape1-N=4000` and another project `CoolingRuns-Shape42-N=8000`. This key might have meaning for some users.

In practice, I know of very few users that store anything here besides a vague project name that usually matches the directory name that it is stored in. An alternative would be to move this configuration information into a project document key/value.

“Project names have been removed from signac’s database schema. Do you wish to create a project document key `project_name` with the value `{project.id}` to preserve this metadata?”

csadorf: @bdice I like that idea.

vyasr: Yes I'd support an optional move to the project document. I feel strongly that we should remove it from the config to clearly separate user data from internal metadata, so offering an optional move to the project document would also be a clear indication to the user of how such data _should_ be stored by users in the future.

vyasr:
In the interest of what we discussed at the last meeting I'm going to attempt to split this PR into multiple separate changes since it's ballooned substantially from the original.

vyasr:
All the relevant changes have been completed in separate PRs, with only #711 left to be merged.

PR #645, opened 2021-11-14
Add indentation when dumping json.

vyasr:
## Description

This PR adds pretty indented formatting to the JSON files dumped by signac.

## Motivation and Context

The introduction of synced collections in #484 led to a change in the formatting of the JSON files used for job documents and statepoints. While we do not promise any specific formatting (so this is not technically a bug), these files are much more readable if the original formatting is maintained and users are accustomed to the old formatting.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,6 +23,7 @@ Fixed
 +++++
 
  - ``H5Store.mode`` returns the file mode (#607).
+ - Formatting of JSON files now matches the changes from synced collections (#645).
 
 [1.7.0] -- 2021-06-08
 ---------------------
--- a/signac/synced_collections/backends/collection_json.py
+++ b/signac/synced_collections/backends/collection_json.py
@@ -251,7 +251,7 @@ class JSONCollection(SyncedCollection):
     def _save_to_resource(self):
         """Write the data to JSON file."""
         # Serialize data
-        blob = json.dumps(self, cls=SyncedCollectionJSONEncoder).encode()
+        blob = json.dumps(self, cls=SyncedCollectionJSONEncoder, indent=2).encode()
         # When write_concern flag is set, we write the data into dummy file and then
         # replace that file with original file. We also enable this mode
         # irrespective of the write_concern flag if we're running in


syjlee (approved):
Nice to add pretty printing.

bdice (changes requested):
Before synced collections, state point files were indented but job documents were not indented. This PR adds indentation to both files, to which I am opposed.

I generally oppose whitespace in either file and prefer the new behavior of synced collections (no indentation). However, state point indentation was previously a behavior of the package that some users implicitly relied on (e.g. committing state points in a git repository) so I can accept a reversion to that behavior if we believe it is necessary for continuity with past versions. The "pretty" aspect is not important to me, only the fact that synced collections introduces a diff with respect to the previous formatting. Many tools exist for pretty-printing JSON, including one built into Python: `python -m json.tool signac_statepoint.json`.

I do not support indenting job documents because that was not the behavior of job documents before synced collections, and the overhead in file size is significant for job documents. Many users have large amounts of data in job documents, and indenting that data will result in an unnecessarily large output file. For a sample job document from my research projects, adding `indent=2` changes the size of the file from 2092 characters to 2690 characters, a 28% increase in disk space and a corresponding loss of some measurable amount of performance in reading/parsing/writing.

I would prefer to close this PR without merging and keep the current behavior of the package rather than revert to our implicit behavior pre-synced collections.

mikemhenry:
@bdice I agree that we should keep both flat, but I could be convinced to have state points follow their old behavior (with a path to making them flat)

How hard would it be to use asv to measure the difference (or do none of the tests create enough docs/state points that would have an impact?

I don't need convincing but it would be nice to point it to a user who wants the old behavior back.

vyasr:
I had this change sitting around locally for a couple of months and I found it when I started working on other signac projects. It was specifically requested by a user at some point. That said, synced collections have been out for months and we've only had one user request this change, and he was OK with the new version once he was able to resolve discrepancies between older and newer versions of signac. @bdice I'd be remiss if I didn't remind you that when that issue was originally raised you _did_ care about the pretty-printing for state points :) so I'm not sure what has changed since then. Personally I do not think it's important for us to revert to that old behavior and would be fine closing this PR. Just having it here is sufficient to socialize the knowledge of this change

bdice:
> I'm not sure what has changed since then.

I was young and naïve at the time but little has changed. 😉

vyasr:
Closing this for now. If we hear many future complaints from users we can revisit the question.

PR #652, opened 2021-12-05
Simplify internal indexing

bdice:
## Description
This PR simplifies the internal indexing behavior by eliminating some functions (`Project._index`, `Project._sp_index`) and internal caches (`Project._index_cache`) with behavior that is nearly-equivalent to other functions (`Project._build_index`) and internal caches (`Project._sp_cache`). See comments below. I'm opening this for review but need to make a final pass for changelog, etc.

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/collection.py
+++ b/signac/contrib/collection.py
@@ -494,7 +494,9 @@ class Collection:
         if self._dirty:
             for _id in self._dirty:
                 self._remove_from_indexes(_id)
-            docs = [self[_id] for _id in self._dirty]
+            # For performance, we read directly from the internal _docs dict to
+            # avoid copying the data like in __getitem__.
+            docs = [self._docs[_id] for _id in self._dirty]
             for key, index in self._indexes.items():
                 tmp = _build_index(docs, key, self._primary_key)
                 for v, group in tmp.items():
--- a/signac/contrib/collection.py
+++ b/signac/contrib/collection.py
@@ -494,9 +494,7 @@ class Collection:
         if self._dirty:
             for _id in self._dirty:
                 self._remove_from_indexes(_id)
-            # For performance, we read directly from the internal _docs dict to
-            # avoid copying the data like in __getitem__.
-            docs = [self._docs[_id] for _id in self._dirty]
+            docs = [self[_id] for _id in self._dirty]
             for key, index in self._indexes.items():
                 tmp = _build_index(docs, key, self._primary_key)
                 for v, group in tmp.items():
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -709,8 +709,9 @@ class Project:
         if not filter:
             return list(self._job_dirs())
         filter = dict(parse_filter(_add_prefix("sp.", filter)))
-        include_job_document = "doc" in _root_keys(filter)
-        index = list(self._build_index(include_job_document=include_job_document))
+        index = list(
+            self._build_index(include_job_document="doc" in _root_keys(filter))
+        )
         return list(Collection(index, _trust=True)._find(filter))
 
     def find_jobs(self, filter=None, *args, **kwargs):
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -669,7 +669,7 @@ class Project:
         """
         from .schema import _build_job_statepoint_index
 
-        index = self._index(include_job_document=False)
+        index = self._build_index(include_job_document=False)
         if subset is not None:
             subset = {str(s) for s in subset}
             index = [doc for doc in index if doc["_id"] in subset]
@@ -710,7 +710,7 @@ class Project:
             return list(self._job_dirs())
         filter = dict(parse_filter(_add_prefix("sp.", filter)))
         include_job_document = "doc" in _root_keys(filter)
-        index = self._index(include_job_document=include_job_document)
+        index = list(self._build_index(include_job_document=include_job_document))
         return list(Collection(index, _trust=True)._find(filter))
 
     def find_jobs(self, filter=None, *args, **kwargs):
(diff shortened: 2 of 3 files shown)


bdice:
Interesting observation while I was experimenting with this PR. Compare the following snippets:

- `list(project.find_jobs(...))` (3.24 seconds for commit 7f448c5 with 100k jobs)
- `list((job for job in project.find_jobs(...)))` (2.09 seconds for commit 7f448c5 with 100k jobs)

The first one is slower because the `JobsCursor.__len__` method is [called as an optimization for pre-allocating lists](https://stackoverflow.com/questions/41474829/why-does-list-ask-about-len). However, in our case, calling `JobsCursor.__len__` and then calling `JobsCursor.__iter__` is expensive compared to calling `JobsCursor.__iter__` alone. Both `__len__` and `__iter__` methods call `_find_job_ids`. Wrapping the `JobsCursor` returned by `project.find_jobs(...)` in the generator expression `(job for job in cursor)` hides the `__len__` method while having a pass-through `__iter__` implementation.

While there is a way to provide a [length hint](https://docs.python.org/3/reference/datamodel.html#object.__length_hint__) to CPython for iterables without a `__len__`, there is no corresponding way to _disable_ this optimization when calling `__len__` is more expensive than the benefit of pre-allocation. It is sensible for `JobsCursor` to have a length defined, so I don't recommend any changes here, but wanted to note this peculiarity.

bdice:
Benchmarks look fine between fbb86df5 (current head of `next`) and 7f448c5d. I saw a tiny bit of regression (slower by maybe 10%) but I couldn't definitively identify what piece got slower. However, I found an optimization for Collection in 7f448c5 while profiling that made up the difference in performance.

       before           after         ratio
     [fbb86df5]       [7f448c5d]
     <next>           <feature/simplify-internal-indexing>
         83.9±3μs       82.6±0.8μs     0.98  benchmarks.ProjectBench.time_determine_len(100, 10, 0, 100, 0)
          861±8μs         855±10μs     0.99  benchmarks.ProjectBench.time_determine_len(1000, 10, 0, 100, 0)
      5.82±0.09ms      5.73±0.05ms     0.99  benchmarks.ProjectBench.time_iterate(100, 10, 0, 100, 0)
       56.7±0.7ms       57.5±0.9ms     1.01  benchmarks.ProjectBench.time_iterate(1000, 10, 0, 100, 0)
       36.9±0.4ms       36.9±0.6ms     1.00  benchmarks.ProjectBench.time_iterate_load_sp(100, 10, 0, 100, 0)
          365±9ms          361±6ms     0.99  benchmarks.ProjectBench.time_iterate_load_sp(1000, 10, 0, 100, 0)
         587±20μs          576±4μs     0.98  benchmarks.ProjectBench.time_iterate_single_pass(100, 10, 0, 100, 0)
      6.16±0.08ms      6.14±0.06ms     1.00  benchmarks.ProjectBench.time_iterate_single_pass(1000, 10, 0, 100, 0)
         373±10μs          385±6μs     1.03  benchmarks.ProjectRandomJobBench.time_search_lean_filter(100, 10, 0, 100, 0)
       6.33±0.2ms       6.37±0.3ms     1.01  benchmarks.ProjectRandomJobBench.time_search_lean_filter(1000, 10, 0, 100, 0)
      1.60±0.03ms      1.62±0.02ms     1.01  benchmarks.ProjectRandomJobBench.time_search_rich_filter(100, 10, 0, 100, 0)
       27.1±0.6ms         26.3±1ms     0.97  benchmarks.ProjectRandomJobBench.time_search_rich_filter(1000, 10, 0, 100, 0)
       27.2±0.4μs       26.5±0.2μs     0.97  benchmarks.ProjectRandomJobBench.time_select_by_id(100, 10, 0, 100, 0)
         25.9±1μs       26.2±0.5μs     1.01  benchmarks.ProjectRandomJobBench.time_select_by_id(1000, 10, 0, 100, 0)

vyasr:
@bdice can you rebase this on `next`/merge in `next` again? I'm not sure if it's just because of my most recent rebase of `next` a couple of days ago or if you originally branched this off earlier, but the diff is very noisy and it's hard to find the relevant changes. I can review once you do. Apologies if the conflicts are caused by my last rebase.

bdice:
@vyasr Just rebased. Take a peek now and see what you think.

bdice:
For posterity: another broad push for optimization would be to reduce the amount of code we use in the `Collection` class when calling `Project.find_jobs`. We copy a large amount of data used from the `Project`'s internal caches during the construction of a `Collection` and its `find` method. At the very least, we never use the `Collection`'s ability to interact with data on disk, so we could eliminate some logic there in a cut-down class.

signac/contrib/project.py
```diff
@@ -710,10 +709,8 @@ def _find_job_ids(self, filter=None):
         if not filter:
             return list(self._job_dirs())
         filter = dict(parse_filter(_add_prefix("sp.", filter)))
-        if "doc" in _root_keys(filter):
-            index = self._index(include_job_document=True)
-        else:
-            index = self._sp_index()
+        include_job_document = "doc" in _root_keys(filter)
+        index = list(self._build_index(include_job_document=include_job_document))
```
vyasr, line 713: Nit: could just inline the single-use flag:
```suggestion
        index = list(self._build_index("doc" in _root_keys(filter)))
```

signac/contrib/collection.py
```diff
@@ -494,7 +494,9 @@ def _update_indexes(self):
         if self._dirty:
             for _id in self._dirty:
                 self._remove_from_indexes(_id)
-            docs = [self[_id] for _id in self._dirty]
+            # For performance, we read directly from the internal _docs dict to
+            # avoid copying the data like in __getitem__.
+            docs = [self._docs[_id] for _id in self._dirty]
```
vyasr, line 499: I haven't looked at Collection code in a very long time so I can't be 100% sure without doing a more thorough re-review of the codebase, but changing this to a non-copying operation will cascade that reference through to `self._indexes`. Are we sure this change is safe? It seems plausible, just covering our bases.

vyasr (commented):
Overall these changes look good, the removals in the Project class all seem sensible to me. One important question about the Collection change, let me know how confident you feel about it and I can re-review in more depth accordingly.

b-butler (approved):
Changes look clean and good.

signac/contrib/collection.py
bdice: This is a good question. I know it's safe for signac's internal use of `Collection` (because we create the `Collection`, use it, and then throw it away). However, I will investigate this a little further to make sure. If it turns out to be unsafe in the general case, I think that this is an indication that we should consider a minimized form of `Collection` for signac's internal use with searching. There's a lot of extraneous logic and copying that could be eliminated.

signac/contrib/project.py
bdice: My rationale for the existing snippet was that giving a name `include_job_document` to the condition helped with self-documenting code. I also liked the use of a keyword argument in the `_build_index` call, and the use of a variable gave it slightly nicer formatting than the longer inlined statement which wraps (iirc) when used with a keyword argument.

signac/contrib/collection.py
bdice: @vyasr If it's to your preference, I might just revert the changes to `Collection` and then make a follow-up PR that implements a minimal `Collection` class that we can use internally. It seems like that would be a better plan in the long term, anyway.

bdice: I reverted the changes to `Collection` in a4b08dc42dea35cbfb3f636c93a4194ad39ecb90. I will create a separate issue to document the potential optimizations in that class (or a minimized version of the class).

bdice: Issue #664 addresses this.

signac/contrib/project.py
bdice: Resolved in 9457215f481648e39bf010cc3dd9a9001a197983.

signac/contrib/collection.py
vyasr: Sounds good, I think this is the safer choice for now. Let's restrict this PR to just those changes you have now that remove largely redundant functions from project.py. `Collection` is complicated and something none of us have touched in a while, so I'd prefer to make a focused effort on improving that if we think it's worthwhile.

signac/contrib/project.py
vyasr: I agree with using the kwarg, so I'm in favor of the change you made. I prefer a linter's line-wrapping to an extraneous variable definition since the latter could change expectations for someone reading the code while formatting is clearly just formatting.

vyasr (approved):

PR #644, opened 2021-11-09
Deprecate project id

vyasr:
## Description

Projects now have a default name, and setting any custom name triggers a warning.

## Motivation and Context

Related to #541. This change provides a softer off-ramp for current users with scripts where they provide the name parameter for project initialization. Rather than removing it entirely, this PR provides a warning that this parameter will be removed in 2.0.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change<sup>1</sup>

<sup>1</sup>The change breaks (or has the potential to break) existing functionality.

## Checklist:

- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
- [x] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [x] I have updated the API documentation as part of the package doc-strings.
- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
- [x] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added all related issue and pull request numbers for future reference (if applicable). See example below.

Issue #527, opened 2021-03-09
Update signac's internal schema


vyasr:
This issue is a meta-issue meant to encompass a few different points relating to changes to the internal schema that we're hoping to have in place for signac 2.0. Ideally, we'd like to synchronize updating from schema version 1 to version 2 with the 2.0 release of the package to simplify the migration for users. Resolution of this issue should imply resolution of at least #423, #390, #541, and #197. If possible, we may also want to resolve #283, but that decision can be made independently. Additionally, we will need to make a final decision on whether our config file should be `signac.rc` or `.signac/config`, for which I would like to have a vote of all committers; this can be handled at the end once most of the other issues have been tackled. By necessity, this issue involves changes to both `signac` and `signac-flow`, but hopefully it will move us a long way towards decoupling any dependencies.

The discussion on [this thread](https://github.com/glotzerlab/signac/pull/517#discussion_r583992472) indicated the need to implement the changes proposed in #197 via a proper schema migration using the tools implemented in #253. Additionally, it also highlighted the problems with how `signac-flow` uses the `signac` configuration files internally, which makes it difficult to independently make changes to either. As a result, I believe that we will need to make a number of changes before we can update our schemata in a consistent manner. I'm going to make a list of action items here, but anyone should feel free to edit it if they think I have any of the items wrong or in the wrong order. There are a few places where I haven't thought through this thoroughly yet and there will definitely be room for improvement.

**Note: This checklist has been edited multiple times as plans have evolved.**
- [x] Remove signac-flow's dependence on signac's internal file schema and configuration. The principal connection that needs to be severed is flow's monkey-patching of the signac config for verifying flow config keys. Once that link is broken, signac-flow may rely on signac's config API alone without being tied to its internals. **Completed in glotzerlab/signac-flow#573**
- [x] Make a new release of `signac-flow` that introduces this migration.
- [x] Introduce a migration of `signac` to `schema_version=2`. Based on the list [here](https://github.com/glotzerlab/signac/issues/197#issuecomment-727297710) as well as discussions on #517 and [here](https://github.com/glotzerlab/signac/issues/197#issuecomment-786932462), this process involves moving `signac_shell_history` and `signac_sp_cache.json.gz`, requiring the workspace to be an immediate subdirectory of the project called `workspace` (i.e. making `workspace_dir` no longer a mutable config variable), and possibly moving `signac.rc` to `.signac/config`.
[...]

PR #654, opened 2021-12-11
Migrate using project directories instead of project instances

vyasr:
## Description
This PR changes the schema migration code to operate in terms of a project's root directory rather than a project class instance. Additionally, loading and checking the schema version of a project is done using a cascade of multiple schema loaders that support different schemas.

## Motivation and Context

The current schema migration logic makes some easily invalidated assumptions about what operations are valid in the course of a migration. The two most problematic assumptions are that 1) `config.load_config` can load schemas of any schema version, and 2) a valid `Project` can be created from any schema version by simply skipping the schema compatibility check. The former is invalidated by, for instance, a change in the format of the config file like the one planned for schema version 2. The latter would be invalidated by any change to the `Project` constructor that assumes the existence of certain keys in the config. The migration to schema version 2 does not immediately break this logic since we are _removing_ project ids, not adding them (see #643 and #644), but there is no reason to assume that we will not make changes in the opposite direction in future schema versions. This PR protects us against such changes.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


vyasr:
@csadorf @jennyfothergill apologies in advance if the quality isn’t up to my usual standard. It’s been hard to dedicate a significant block of time, so I’ve been working on this in very brief fits and starts and I figured at this point it would be better to just get feedback on what I had rather than wait longer.

signac/contrib/migration/__init__.py
```diff
@@ -5,102 +5,135 @@
...
-            except Exception as e:
-                raise RuntimeError(f"Failed to apply migration {destination}.") from e
-            else:
-                _update_project_config(project, schema_version=destination)
-                print("OK", file=sys.stderr)
-                yield origin, destination
+    from ..project import Project
+
+    if isinstance(root_directory, Project):
+        warnings.warn(
```
bdice, line 104: Since we don't have any migrations yet, I would favor just changing the behavior rather than throwing a warning. We don't need any backwards compatibility yet because we haven't released a new schema version.

signac/contrib/migration/__init__.py
```diff
@@ -5,102 +5,135 @@
...
+                    _get_config_schema_version(root_directory, guess),
+                    __version__,
+                    schema_version,
                 )
             )
 
 
-def apply_migrations(project):
+def apply_migrations(root_directory):
     """Apply migrations to a project."""
```
bdice, line 100: Is this method supposed to be public or private? If public, it needs more docstrings (might not hurt to expand the docstring even if it's private). If private, it needs to be renamed to `_apply_migrations`.

signac/contrib/migration/__init__.py
```diff
@@ -5,102 +5,135 @@
...
+    from ..project import Project
+
+    if isinstance(root_directory, Project):
+        warnings.warn(
+            "Migrations should be applied to a directory containing a signac project, "
+            "not a project object.",
+            FutureWarning,
+        )
+        root_directory = root_directory.root_directory()
+    lock = FileLock(os.path.join(root_directory, FN_MIGRATION_LOCKFILE))
```
bdice, line 110: Should this go in the `try`? (move down one line?)

signac/contrib/migration/__init__.py
```diff
@@ -5,102 +5,135 @@
...
+                        end="",
+                        file=sys.stderr,
+                    )
+                    migrate(root_directory)
+                except Exception as e:
+                    raise RuntimeError(
+                        f"Failed to apply migration {destination}."
+                    ) from e
+                else:
+                    config = get_config(os.path.join(root_directory, "signac.rc"))
```
bdice, line 126: Wondering how this would be affected by suggestions in #527 to move to `.signac/config`.

signac/contrib/migration/v0_to_v1.py
```diff
@@ -6,8 +6,61 @@
...
+def _load_config_v1(root_directory):
+    try:
+        import configobj
+    except ModuleNotFoundError:
+        try:
+            from signac.common import configobj
+        except ModuleNotFoundError:
+            raise RuntimeError(
+                "signac schema version 1 can only be read with configobj. "
+                "Please install configobj and try again."
```
bdice, line 31: configobj hasn't been released since 2014... 😬 Should we mention "configobj or signac 1.x" since signac 1.x may be installable for longer than configobj itself? Not sure.

bdice (changes requested):
I have a handful of comments. I think this looks pretty good overall. It's unfortunate that the logic for migrations is so complex, but I can't envision a simpler design.

signac/contrib/migration/__init__.py
vyasr: That's fine with me too. Technically this function is already public API, so I introduced the compatibility in case others preferred it, but in practice it's a no-op and there's no reason for a user to be calling it yet.

signac/contrib/migration/__init__.py
vyasr: It is public. This is the public Python API for migrating, the analog to the CLI `signac migrate` (which is currently disabled but should be exposed soon). I can more documentation.

signac/contrib/migration/__init__.py
vyasr: I don't know the ways in which `FileLock` can fail, and I don't think this `try-finally` was designed to handle that failure specifically, but there's no harm in moving it down and allowing the lock file cleanup to happen just in case `FileLock` fails.

signac/contrib/migration/__init__.py
vyasr: Good catch, I missed this change. This line should be using the `_CONFIG_LOADERS` dict just like the other changes in this PR.

signac/contrib/migration/v0_to_v1.py
vyasr: Well we can't use signac 1.x because even though that could load the config file it wouldn't be able to perform the migration. If we're really worried about this one option would be to not delete the configobj embedded in signac until signac >2.0 so that a user could use signac 2.0 for the migration even if configobj was no longer installable as a standalone. I don't think we really need to worry about this though; in the worst case scenario where configobj completely vanishes (both source and distributions) we can always add a section to the docs on how to extract the source of configobj from signac 1.x and make it available for the purpose of migrations.

csadorf: I think it would be better to not change the import logic and to simply continue to include configobj with the package since it is _required_ for the migration and thus a real dependency of signac 2.0.

signac/contrib/migration/__init__.py
```diff
@@ -5,102 +5,135 @@
...
 FN_MIGRATION_LOCKFILE = ".SIGNAC_PROJECT_MIGRATION_LOCK"
 
 
-MIGRATIONS = {
-    ("0", "1"): migrate_v0_to_v1,
+# Config loaders must be functions whose first argument is a directory from
+# which to read configuration information. If the logic for loading a schema
+# has not changed from previous versions, it need not be added to this dict.
+# This dictionary only needs to contains all unique loaders in signac's history
+# to ensure that any prior config may be loaded for migration.
```
csadorf, line 27: I do not think we need to support all loaders, but only those that are needed for supported migration paths.
```suggestion
# This dictionary only needs to contain all unique loaders for schema
# versions that are supported as starting points for migration.
```

signac/contrib/migration/__init__.py
```diff
@@ -5,102 +5,135 @@
...
+_CONFIG_LOADERS = {
+    # The following line should be uncommented when schema version 2 is
+    # introduced, making load_config fail for v1 schemas.
+    # "1": _load_config_v1,
```
csadorf, line 31: Same comment as above, maybe this could be automated?

csadorf (changes requested):
Codecov is complaining a lot about missing tests. Maybe we can restructure the existing tests to more systematically cover those additional cases?

signac/contrib/migration/__init__.py
```diff
@@ -5,102 +5,135 @@
...
 from filelock import FileLock
 from packaging import version
 
-from ...common.config import get_config
+from ...common.config import get_config, load_config
 from ...version import SCHEMA_VERSION, __version__
-from .v0_to_v1 import migrate_v0_to_v1
+
+# To be uncommented when switching to version 2.
+# from .v0_to_v1 import _load_config_v1, _migrate_v0_to_v1
```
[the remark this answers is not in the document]
vyasr: This question is now moot because I've made this change now and structured the code so that we'll always make the change whenever the schema version is updated. There's no reason to continue using the primary config API here, we can just always use the migration-specific config APIs.

signac/contrib/migration/v0_to_v1.py
vyasr: I'm fine including it for a while, although I think it would be annoying to include indefinitely. If not in 2.0, I'd at least consider removing it in 3.0. I don't anticipate keeping signac _quite_ this dependency free forever. In particular, when we switch to TOML do we want to embed a TOML parsing package still, or just make one a dependency?

signac/contrib/migration/__init__.py
vyasr: Same response as above.

signac/contrib/migration/v0_to_v1.py
csadorf: I would suggest to include it as long as we support the related migration. I think it is fairly typical to expect users to install intermediate major versions for longer migration paths. This means signac v3 would only support a migration from v2, but not from v0 or v1.

vyasr: That sounds reasonable to me.

vyasr:
> Codecov is complaining a lot about missing tests. Maybe we can restructure the existing tests to more systematically cover those additional cases?

Thanks for pointing this out. I've gone through the untested cases and improved test coverage where appropriate. The remaining untested cases fall into two buckets:
1. Various lines in `_get_config_schema_version` that will start getting triggered as soon as we have more than two versions and more than a single migration. I'm comfortable leaving these untested for now with the expectation that additional tests in  #643 will start covering these.
2. Two exception case in `apply_migrations` itself. The `FileNotFoundError` on the lock file I think is something we don't have to do anything about. We frequently leave those types of cases untested in signac since they're there to case subtle filesystem errors rather than logic errors. The other exception case is there to handle the scenario when a migration fails. This code is also IMO not worth testing until #643. If either of you can think of a reasonable scenario where we wouldn't fail in `_collect_migrations` (i.e. an invalid schema version) but where the v1->v2 migration could fail, I'm happy to add that to #643.

In general the lack of coverage just reflects that we've added migration logic before we had a need for it in anticipation of future use cases. I made these changes on master because they would have cluttered the PR actually introducing the v1->v2 migration and made them hard to review, and the lack of testability is the tradeoff. We should see that improve with #643 (and I'm sure we'll find additional bugs then, but at least the core logic in place now seems sound).

There is one other change of note that I introduced. While attempting to improve coverage, I realized that because `apply_migrations` was defined as a generator it was a no-op unless the result was iterated over, i.e. `apply_migrations(root)` does nothing unless it's wrapped with something like `list(apply_migrations(root))` or `for (origin, destination) in apply_migrations(root): ...`. That API doesn't make any sense to me, and being able to iterate over the intermediate states of migration is IMO only really useful for testing purposes, so I've removed the yield statement in `apply_migrations` so that it simply performs all migrations at once. Let me know if either of you foresee any issues with that or would prefer a different resolution.

signac/contrib/migration/__init__.py
```diff
@@ -5,102 +5,141 @@
...
+        root_directory, _PARSED_SCHEMA_VERSION
+    )
+    if current_schema_version > schema_version:
         # Project config schema version is newer and therefore not supported.
         raise RuntimeError(
-            "The signac schema version used by this project is {}, but signac {} "
-            "only supports up to schema version {}. Try updating signac.".format(
-                get_config_schema_version(), __version__, SCHEMA_VERSION
-            )
+            f"The signac schema version used by this project is "
```
bdice, line 77: This might be considered lint by some tools. It's a f-string with no fields.
```suggestion
            "The signac schema version used by this project is "
```

bdice (approved):
I'm happy with this PR. As noted before, it's a little hard to review these changes because we don't actually have a migration defined yet. The code looks stable and general enough, and there is a significant benefit to having this merged to help unblock other work. Let's find the remaining sharp edges when we introduce a real migration!

vyasr:
@csadorf let me know if you'd like to re-review this. It would be fine IMO to merge this, then work out any further kinks in the context of #643 since that will give us our first real migration to stress test this with anyway, but if you see other issues that you think are worth addressing in the context of this PR we can deal with them first.

csadorf:
> @csadorf let me know if you'd like to re-review this. It would be fine IMO to merge this, then work out any further kinks in the context of #643 since that will give us our first real migration to stress test this with anyway, but if you see other issues that you think are worth addressing in the context of this PR we can deal with them first.

Thanks, I will review it right now.

signac/contrib/migration/__init__.py
```diff
@@ -5,102 +5,141 @@
...
-    for fn in ("signac.rc", ".signacrc"):
-        config = get_config(project.fn(fn))
-        if "project" in config:
-            break
-    else:
-        raise RuntimeError("Unable to determine project configuration file.")
-    config.update(kwargs)
-    config.write()
-    _reload_project_config(project)
+_PARSED_SCHEMA_VERSION = version.parse(SCHEMA_VERSION)
```
csadorf, line 36: I'm afraid that the term "parsed schema version" originally threw me off, because I thought it was parsed from the directory or so. Maybe just rename it to `_THIS_SCHEMA_VERSION` or so?

signac/contrib/migration/__init__.py
```diff
@@ -5,102 +5,141 @@
...
 
+_PARSED_VERSION_LIST = list(
```
csadorf, line 38: Same as above, I don't think the word `_PARSED` is adding much context here. It's still just a list of versions.

signac/contrib/migration/__init__.py
```diff
@@ -5,102 +5,141 @@
...
+                    )
+                    config["schema_version"] = destination
+                    config.write()
+
+                    print("OK", file=sys.stderr)
+    finally:
+        try:
+            os.unlink(lock.lock_file)
+        except FileNotFoundError:
+            pass
```
csadorf, line 142: Is this not the responsibility of the `FileLock` context manager protocol?

signac/contrib/migration/v0_to_v1.py
```diff
@@ -6,8 +6,30 @@
...
+
+def _load_config_v1(root_directory):
+    cfg = configobj.ConfigObj(
+        os.path.join(root_directory, "signac.rc"), configspec=_cfg.split("\n")
+    )
+    validator = configobj.validate.Validator()
+    if not cfg.validate(validator):
+        raise RuntimeError(
+            "This project's config file is not compatible with signac's v1 schema."
+        )
```
csadorf, line 29: I believe that additional keys will throw here (see https://github.com/glotzerlab/signac/blob/a53a098e8366a785f4572e0872a5788c20c00665/signac/common/config.py#L103-L109). Is that intended?

csadorf (commented):
Looks very good overall, just a few minor questions.

signac/contrib/migration/v0_to_v1.py
bdice: This is a good question that I meant to ask about and forgot!

signac/contrib/migration/__init__.py
vyasr: In this case I agree that simply `_VERSION_LIST` is sufficient and have made that change.

signac/contrib/migration/__init__.py
vyasr: The intent here and below was to communicate the difference between version strings and `packaging.version.Version` objects. I don't know if `_THIS_SCHEMA_VERSION` really conveys that. Do you have an alternate name that might help with that? Otherwise I can go with your suggestion, but it seems too ambiguous to meaningful differentiate from `SCHEMA_VERSION`.

signac/contrib/migration/__init__.py
vyasr: I would think so, but I assumed there was a reason this code was added in the first place. If not, I'm happy to remove this whole try-except block and simplify the code.

signac/contrib/migration/v0_to_v1.py
vyasr: `ConfigObj.validate` will not throw when provided extra keys. It only validates keys that are actually present in the spec. For example:
```
In [1]: >>> from signac.common import configobj

In [2]: >>> _cfg = """
   ...:    ...: schema_version = string()
   ...:    ...: project = string()
   ...:    ...: workspace_dir = string(default='workspace')
   ...:    ...: """

In [3]: >>> cfg = configobj.ConfigObj(configspec=_cfg.split("\n"))

In [4]: >>> validator = configobj.validate.Validator()

In [5]: >>> cfg.validate(validator)
Out[5]: {'schema_version': False, 'project': False, 'workspace_dir': True}

In [6]: >>> cfg.update({'schema_version': '1', 'project': 'foo'})

In [7]: >>> cfg.validate(validator)
Out[7]: True

In [8]: >>> cfg.update({'foo': 'bar'})

In [9]: >>> cfg.validate(validator)  # This returns true despite the extraneous key.
Out[9]: True
```
However, there is a bug here in that I'm checking the truthiness of the validate output rather than checking that it's exactly `True`. I just fixed that.

signac/contrib/migration/__init__.py
csadorf: 🤷

vyasr: @bdice do you know?

signac/contrib/migration/__init__.py
csadorf: > The intent here and below was to communicate the difference between version strings and packaging.version.Version objects.

I got that, I am just letting you know that I was thrown off by the prefix "parsed_" until I saw that. If you don't like "this_", just leave it _as is_.

signac/contrib/migration/__init__.py
csadorf: I have since realized that the context manager does in fact not remove the lock file. The lock file is "unlocked", but not necessarily removed.

bdice: Correct. I tried it locally. The `FileLock` context manager does not remove the lock file -- it creates it and uses it. The `SoftFileLock`, which only checks file existence and does not rely on the deeper OS primitives for file locking, performs cleanup and removes the file when exiting the context manager. https://py-filelock.readthedocs.io/en/latest/index.html#filelock-vs-softfilelock

vyasr:
@csadorf I'm treating our discussion at the meeting as approval and I'm going to go ahead and merge this PR now.

PR #666, opened 2022-01-21
Raise error on underspecified path for import/export and view

cbkerr:
## Description

Just the bug fix.

There is some duplicate code, but that will be refactored in the rest of #642.

## Motivation and Context

see #642 for a description

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -29,6 +29,7 @@ Fixed
 +++++
 
  - ``H5Store.mode`` returns the file mode (#607).
+ - User-provided path functions now raise an error if not unique (#642).
 
 [1.7.0] -- 2021-06-08
 ---------------------
--- a/changelog.txt
+++ b/changelog.txt
@@ -29,7 +29,7 @@ Fixed
 +++++
 
  - ``H5Store.mode`` returns the file mode (#607).
- - User-provided path functions now raise an error if not unique (#642).
+ - User-provided path functions now raise an error if not unique (#666).
 
 [1.7.0] -- 2021-06-08
 ---------------------
--- a/signac/contrib/import_export.py
+++ b/signac/contrib/import_export.py
@@ -169,15 +169,16 @@ def _check_path_function_unique(jobs, path_spec, path_function):
 
     """
     job_paths = Counter(path_function(job) for job in jobs)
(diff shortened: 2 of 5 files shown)


changelog.txt
```diff
@@ -29,6 +29,7 @@ Fixed
 +++++
 
  - ``H5Store.mode`` returns the file mode (#607).
+ - User-provided path functions now raise an error if not unique (#642).
```
cbkerr, line 32:
```suggestion
 - User-provided path functions now raise an error if not unique (#666).
```

signac/contrib/import_export.py
```diff
@@ -149,6 +149,43 @@ class _SchemaPathEvaluationError(RuntimeError):
...
+        A callable path generating function.
+
+    Raises
+    -----
+    RuntimeError
+        If paths generated with given path function are not unique.
+
+    """
+    # quick check first
+    links = {path_function(job) for job in jobs}
```
bdice, line 171: One way to avoid re-computing the `path_function` for all jobs if the "quick check" fails would be to use a `Counter`: https://docs.python.org/3/library/collections.html#collections.Counter

signac/contrib/import_export.py
```diff
@@ -293,25 +333,18 @@ def _export_jobs(jobs, path, copytree):
     dst : str
         Destination path.
 
-    Raises
-    ------
-    RuntimeError
-        If paths generated with given path function are not unique.
```
bdice, line 299: This can still raise a RuntimeError, so we shouldn't delete this. (The errors listed in a function's docstring may be raised by the code it calls, like in this case with `_check_path_function`.)

bdice (changes requested):
I have a couple small suggestions. Otherwise this looks good.

signac/contrib/import_export.py
cbkerr: I added this but didn't investigate the impact on the performance.

bdice (approved):
Thanks @cbkerr. I pushed some changes and will set this to automerge once CI passes.

PR #674, opened 2022-02-02
Remove unnecessary config APIs

vyasr:
## Description

This PR removes `config.get_config`, which can be replaced by using `config.read_config_file` directly, and makes `search_standard_dirs` an internal function. If this PR is approved I will make a corresponding deprecation PR for master.

## Motivation and Context

Reducing cruft for 2.0. This change was extracted from #643.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature
- [x] Breaking change<sup>1</sup>

<sup>1</sup>The change breaks (or has the potential to break) existing functionality.

## Checklist:

- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
- [x] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [x] I have updated the API documentation as part of the package doc-strings.
- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
- [ ] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added all related issue and pull request numbers for future reference (if applicable). See example below.

PR #675, opened 2022-02-03
Deprecate unnecessary config APIs

vyasr:
## Description

This PR deprecates the config APIs removed on next in #674. I went with the simplest possible deprecation method of implementing an internal method called by the existing methods since all of these have already been removed on next with the appropriate changes, so limiting this PR's changeset minimizes conflicts when it is merged down.

## Motivation and Context

Prep for 2.0

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/common/config.py
+++ b/signac/common/config.py
@@ -7,6 +7,9 @@ import logging
 import os
 import stat
 
+from deprecation import deprecated
+
+from ..version import __version__
 from .configobj import ConfigObj, ConfigObjError
 from .errors import ConfigError
 from .validate import cfg, get_validator
@@ -55,10 +58,20 @@ def search_tree(root=None):
             root = up
 
 
-def search_standard_dirs():
+def _search_standard_dirs():
     """Locates signac configuration files in standard directories."""
     for path in CONFIG_PATH:
         yield from _search_local(path)
+
+
+@deprecated(
+    deprecated_in="1.8",
+    removed_in="2.0",
+    current_version=__version__,
+    details="The search_standard_dirs method is deprecated.",
+)
+def search_standard_dirs():  # noqa: D103
+    yield from _search_standard_dirs()
 
 
 def check_permissions(filename):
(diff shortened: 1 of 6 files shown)


tests/test_job.py
```diff
@@ -55,7 +55,7 @@ def builtins_dict():
 
 def config_from_cfg(cfg):
     cfile = io.StringIO("\n".join(cfg))
-    return signac.common.config.get_config(cfile)
+    return signac.common.config._get_config(cfile)
```
bdice, line 58: Rewrite this helper to use functions from the public API for 2.0 if possible.

tests/test_project.py
```diff
@@ -2485,7 +2485,7 @@ def test_project_schema_versions(self):
         assert version.parse(self.project.config["schema_version"]) < version.parse(
             impossibly_high_schema_version
         )
-        config = get_config(self.project.fn("signac.rc"))
+        config = _get_config(self.project.fn("signac.rc"))
```
bdice, line 2488: Can this test (and others below) be rewritten to use the functions that will be in the public API in 2.0?

signac/common/config.py
```diff
@@ -55,12 +58,22 @@ def search_tree(root=None):
             root = up
 
 
-def search_standard_dirs():
+def _search_standard_dirs():
```
bdice, line 61: Do we need to deprecate `search_tree` and the other permissions-related functions in this file? Can’t recall if we’ve also removed those in 2.0 or if they should be removed.

tests/test_job.py
vyasr: The reason I didn't do this is listed in the description: those changes are already made in next so this will likely just lead to conflicts that I don't want to address again :)

signac/common/config.py
vyasr: Those were already changed in next, but whoever made those changes originally didn't bother to deprecate and I wasn't sure whether to go ahead and add those deprecations here. I can do that now if you'd like.

tests/test_job.py
bdice: Sorry! I misunderstood which conflicts you were describing. This is fine as-is, then.

signac/common/config.py
bdice: I don't think deprecations are necessary -- let's just remove them.

bdice (approved):

PR #676, opened 2022-02-04
Improve test coverage of Collection class's find methods.

bdice:
## Description
While working on removing internal usage of the `Collection` class in #667 for signac 2.0, I wanted to improve the test coverage for the `Collection` class's find methods to ensure that the new code in signac 2.0 does not introduce bugs/regressions.

I fixed one bug when searching by the primary key (an `AssertionError` was raised unexpectedly when searching for an id that did not exist, when the result should just be an empty set). Additionally, I expanded test coverage of several other features relating to search operators like `$and` and `$type`. I also fixed some mistakes/omissions in the class documentation.

## Motivation and Context
These tests will be used to improve #667 and ensure that we have greater coverage of our `find` API in the smaller Collection-like class in that PR.

## Types of Changes
- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -35,6 +35,7 @@ Fixed
 
  - ``H5Store.mode`` returns the file mode (#607).
  - User-provided path functions now raise an error if not unique (#666).
+ - ``Collection`` class no longer raises an error when searching by a primary key that does not exist (#676).
 
 [1.7.0] -- 2021-06-08
 ---------------------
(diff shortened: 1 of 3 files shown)


csadorf (approved):
If I was Linus T. I would argue that this "bug" must remain. While maybe not intended, it is well-defined behavior that some end-users might be relying on. But since I am not and this is signac and not the Linux kernel, I approve of this fix and I am honestly a bit surprised that this has never come up before. Thanks a lot!

PR #677, opened 2022-02-04
Remove user-provided project name from API

vyasr:
## Description

This PR removes the ability for users to set a project name. The name still exists, but it is always just "project". These changes were split off from #643. Currently this changeset only includes the minimal set of changes to ensure that tests pass, but a large number of tests will now trigger warnings. Once reviewers indicate their approval of the new API I can go through and clean up tests.

Specifically, the changes here ensure that in 2.0 a user who accidentally passes a project name as either a positional or keyword argument will only see a warning, not an error. This choice is intended to provide a softer off-ramp for project names since they have been an integral part of signac since its inception. The only cases that were previously valid that will now fail are if the user provides multiple positional arguments in order to set the root directory, e.g. `init_project("project_name", "/root/directory")`. I think that this is an acceptable level of breakage since the vast majority of users do not set the root explicitly.

I haven't removed the `Project.id` property here since the name still exists and users can still access it until the 2.0 migration actually removes it. I will remove the id property at that time.

## Motivation and Context

Prep for 2.0. Part of #541.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -1545,7 +1545,7 @@ class Project:
                     f"init_project() got an unexpected keyword argument '{next(iter(kwargs))}'"
                 )
             if name is not None:
-                assert version.parse(__version__).major < 3
+                assert version.parse(__version__).release[0] < 3
                 warnings.warn(
                     "Project names were removed in signac 2.0. If your project name contains "
                     "important information, consider storing it in the project document instead.",
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -1518,8 +1518,11 @@ class Project:
             configuration.
 
         """
+        # TODO: Remove both the `if args` and `if kwargs` blocks in version 3.0
+        # when we remove backwards compatibility for project name APIs.
         if args:
-            if len(args) == 1:
+            num_args = len(args)
+            if num_args == 1:
                 warnings.warn(
                     "Project names were removed in signac 2.0. If you intended to call "
                     "`init_project` with a root directory, please provided it as a keyword "
@@ -1530,22 +1533,22 @@ class Project:
                 )
             else:
                 # Match the usual error from misusing keyword-only args.
-                num_args = len(args)
                 raise TypeError(
                     f"init_project() takes 0 positional arguments but {num_args} "
-                    f"{'was' if num_args == 1 else 'were'} given"
+                    f"were given"
                 )
         if kwargs:
-            if len(kwargs) == 1 and "name" in kwargs:
+            name = kwargs.pop("name", None)
+            if kwargs:
+                # Match the usual error from extra keyword args.
+                raise TypeError(
+                    f"init_project() got an unexpected keyword argument '{next(iter(kwargs))}'"
+                )
+            if name is not None:
                 warnings.warn(
                     "Project names were removed in signac 2.0. If your project name contains "
                     "important information, consider storing it in the project document instead.",
                     FutureWarning,
-                )
-            else:
-                # Match the usual error from extra keyword args.
-                raise TypeError(
-                    f"init_project() got an unexpected keyword argument '{next(iter(kwargs))}'"
                 )
 
         if root is None:
@@ -2167,37 +2170,9 @@ def init_project(*args, root=None, workspace=None, make_dir=True, **kwargs):
         configuration.
 
     """
-    if args:
-        if len(args) == 1:
-            warnings.warn(
-                "Project names were removed in signac 2.0. If you intended to call "
-                "`init_project` with a root directory, please provided it as a keyword "
-                f"argument: `init_project(root={args[0]})`. If your project name "
-                "contains important information, consider storing it in the project document "
-                "instead.",
-                FutureWarning,
-            )
-        else:
-            # Match the usual error from misusing keyword-only args.
-            num_args = len(args)
-            raise TypeError(
-                f"init_project() takes 0 positional arguments but {num_args} "
-                "{'was' if num_args == 1 else 'were'} given"
-            )
-    if kwargs:
-        if len(kwargs) == 1 and "name" in kwargs:
-            warnings.warn(
-                "Project names were removed in signac 2.0. If your project name contains "
-                "important information, consider storing it in the project document instead.",
-                FutureWarning,
-            )
-        else:
-            # Match the usual error from extra keyword args.
-            raise TypeError(
-                f"init_project() got an unexpected keyword argument '{next(iter(kwargs))}'"
-            )
-
-    return Project.init_project(root=root, workspace=workspace, make_dir=make_dir)
+    return Project.init_project(
+        *args, root=root, workspace=workspace, make_dir=make_dir, **kwargs
+    )
 
 
 def get_project(root=None, search=True, **kwargs):
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -1485,7 +1485,7 @@ class Project:
 
     @classmethod
     def init_project(cls, *args, root=None, workspace=None, make_dir=True, **kwargs):
-        """Initialize a project at the provided root directory.
+        """Initialize a project in the provided root directory.
 
         It is safe to call this function multiple times with the same
         arguments. However, a `RuntimeError` is raised if an existing project
@@ -1545,6 +1545,7 @@ class Project:
                     f"init_project() got an unexpected keyword argument '{next(iter(kwargs))}'"
                 )
             if name is not None:
+                assert version.parse(__version__).major < 3
                 warnings.warn(
                     "Project names were removed in signac 2.0. If your project name contains "
                     "important information, consider storing it in the project document instead.",
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -1521,25 +1521,16 @@ class Project:
         # TODO: Remove both the `if args` and `if kwargs` blocks in version 3.0
         # when we remove backwards compatibility for project name APIs.
         name = None
+        # The key used to store project names in the project document.
+        name_key = "signac_project_name"
         if args:
             num_args = len(args)
             if num_args == 1:
                 name = args[0]
-                warnings.warn(
-                    "Project names were removed in signac 2.0. If you intended to call "
-                    "`init_project` with a root directory, please provided it as a keyword "
-                    f"argument: `init_project(root={args[0]})`. If your project name "
-                    "contains important information, consider storing it in the project document "
-                    "instead. The name provided will be stored in the project document with "
-                    "the key `signac_project_name`. Calling `init_project` with a "
-                    "name will become an error in signac 3.0.",
-                    FutureWarning,
-                )
             else:
                 # Match the usual error from misusing keyword-only args.
                 raise TypeError(
-                    f"init_project() takes 0 positional arguments but {num_args} "
-                    f"were given"
+                    f"init_project() takes 0 positional arguments but {num_args} were given"
                 )
         if kwargs:
             name = kwargs.pop("name", None)
@@ -1548,25 +1539,25 @@ class Project:
                 raise TypeError(
                     f"init_project() got an unexpected keyword argument '{next(iter(kwargs))}'"
                 )
-            if name is not None:
-                assert version.parse(__version__) < version.parse("3.0.0")
-                warnings.warn(
-                    "Project names were removed in signac 2.0. If your project name contains "
-                    "important information, consider storing it in the project document instead."
-                    "The name provided will be stored in the project document with "
-                    "the key `signac_project_name`. Calling `init_project` with a "
-                    "name will become an error in signac 3.0.",
-                    FutureWarning,
-                )
 
-        # The key used to store project names in the project document.
-        name_key = "signac_project_name"
+        if name is not None:
+            assert version.parse(__version__) < version.parse("3.0.0")
+            warnings.warn(
+                "Project names were removed in signac 2.0. If you intended to call "
+                "`init_project` with a root directory as the sole positional argument, please "
+                f"provide it as a keyword argument: `init_project(root={name})`. If your "
+                "project name contains important information, consider storing it in the "
+                "project document instead. The name provided will be stored in the project "
+                f"document with the key `{name_key}`. Calling `init_project` with a name will "
+                "become an error in signac 3.0.",
+                FutureWarning,
+            )
 
         if root is None:
             root = os.getcwd()
         try:
             project = cls.get_project(root=root, search=False)
-            existing_name = project.config.get(name_key)
+            existing_name = project.doc.get(name_key)
             if name is not None and name != existing_name:
                 raise ValueError(
                     "The name provided to `init_project` does not match the existing "
@@ -1584,6 +1575,8 @@ class Project:
             config["schema_version"] = SCHEMA_VERSION
             config.write()
             project = cls.get_project(root=root)
+            if name != "project":
+                project.doc[name_key] = name
             assert project.id == str(name)
             return project
         else:
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -1564,7 +1564,6 @@ class Project:
                     f"project document in which {name_key}={existing_name}."
                 )
         except LookupError:
-            name = name or "project"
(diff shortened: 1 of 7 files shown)


signac/common/config.py
```diff
@@ -6,9 +6,6 @@
 import logging
 import os
 
-from deprecation import deprecated
-
-from ..version import __version__
```
bdice, line 11: Can we get this change merged directly onto `next` to satisfy flake8 / pre-commit CI? Did we merge something with CI failing?

signac/contrib/project.py
```diff
@@ -1526,11 +1518,42 @@ def init_project(cls, name, root=None, workspace=None, make_dir=True):
             configuration.
 
         """
+        if args:
+            if len(args) == 1:
+                warnings.warn(
```
bdice, line 1523: If we add this API layer for backwards compatibility, we definitely need to test it well (and mark the tests as `TODO: Remove when...`). It appears that this isn't being tested right now, according to codecov.

signac/contrib/project.py
```diff
@@ -1526,11 +1518,42 @@ def init_project(cls, name, root=None, workspace=None, make_dir=True):
...
+                    "contains important information, consider storing it in the project document "
+                    "instead.",
+                    FutureWarning,
+                )
+            else:
+                # Match the usual error from misusing keyword-only args.
+                num_args = len(args)
+                raise TypeError(
+                    f"init_project() takes 0 positional arguments but {num_args} "
+                    f"{'was' if num_args == 1 else 'were'} given"
```
bdice, line 1536: We've explicitly ruled out `num_args == 1` from the previous `if` block. Since this entire error block should be removed in the future, I would mark it as `# TODO: Remove this error when removing support for single positional argument project name` and always use "were" in this message.

signac/contrib/project.py
```diff
@@ -1526,11 +1518,42 @@ def init_project(cls, name, root=None, workspace=None, make_dir=True):
...
+                )
+        if kwargs:
+            if len(kwargs) == 1 and "name" in kwargs:
```
bdice, line 1539: This should check `"name" in kwargs` by popping it with a default value, and then raise a `TypeError` if other kwargs remain. Otherwise it's possible to pass `name` along other invalid keyword arguments and get an error message about `name`, when `name` would otherwise be supported if the other invalid keyword arguments were removed.
```suggestion
            if kwargs.pop("name", None):
```

signac/contrib/project.py
```diff
@@ -1526,11 +1518,42 @@ def init_project(cls, name, root=None, workspace=None, make_dir=True):
...
+                warnings.warn(
+                    "Project names were removed in signac 2.0. If your project name contains "
+                    "important information, consider storing it in the project document instead.",
+                    FutureWarning,
+                )
+            else:
```
bdice, line 1545: If any kwargs remain after popping `"name"` (see above):
```suggestion
            if kwargs:
```

signac/contrib/project.py
```diff
@@ -1526,11 +1518,42 @@ def init_project(cls, name, root=None, workspace=None, make_dir=True):
...
+                # Match the usual error from extra keyword args.
```
bdice, line 1546: Add a TODO about when this workaround can be removed like I suggested above: `# TODO: Remove this error when removing support for keyword argument name.`

signac/contrib/project.py
```diff
@@ -2154,9 +2167,37 @@ def init_project(name, root=None, workspace=None, make_dir=True):
...
+                "important information, consider storing it in the project document instead.",
+                FutureWarning,
+            )
+        else:
+            # Match the usual error from extra keyword args.
+            raise TypeError(
+                f"init_project() got an unexpected keyword argument '{next(iter(kwargs))}'"
+            )
+
+    return Project.init_project(root=root, workspace=workspace, make_dir=make_dir)
```
bdice, line 2200: Just pass through `*args, **kwargs` here -- don't re-implement the error checking.

bdice (commented):
Suggestions attached.

signac/contrib/project.py
bdice: Or maybe switch this logic around, so it can issue the error _before_ potentially issuing a warning. It would look like:
```python
name = kwargs.pop("name", None)
if kwargs:
    raise TypeError(...)
if name:
    warnings.warn(...)
```

signac/common/config.py
vyasr: No, just an oversight in the last rebase. I fixed it there.

signac/contrib/project.py
vyasr: That's only showing up right now because as you pointed out I was duplicating the error handling in the `init_project` function, which is typically what is called in our tests (`signac.init_project` rather than `Project.init_project`). That being said, currently this behavior is being tested by a large number of tests that are still using the old API. As I mentioned in the description, I'll need to go through and update all the existing tests now that a reviewer has approved this approach, and when I do that I can add dedicated tests for this behavior.

signac/contrib/project.py
```diff
@@ -1526,11 +1518,45 @@ def init_project(cls, name, root=None, workspace=None, make_dir=True):
...
+                    f"were given"
+                )
+        if kwargs:
+            name = kwargs.pop("name", None)
+            if kwargs:
+                # Match the usual error from extra keyword args.
+                raise TypeError(
+                    f"init_project() got an unexpected keyword argument '{next(iter(kwargs))}'"
+                )
+            if name is not None:
```
csadorf, line 1547: Maybe we can fail tests here explicitly?
```suggestion
            if name is not None:
                # 3.x: remove backwards compatible handling of name argument
                assert parse(__version__).major < 3
```

signac/contrib/project.py
```diff
@@ -1544,15 +1570,12 @@ def init_project(cls, name, root=None, workspace=None, make_dir=True):
             assert project.id == str(name)
             return project
         else:
-            if project.id != str(name) or (
-                workspace is not None
-                and os.path.realpath(workspace) != os.path.realpath(project.workspace())
-            ):
+            if workspace is not None and os.path.realpath(
+                workspace
+            ) != os.path.realpath(project.workspace()):
```
csadorf, line 1575: **Not in this PR:** We should aim to consistently use the `pathlib.Path` API for all `os.PathLike` objects.

signac/contrib/project.py
```diff
@@ -1526,11 +1518,45 @@ def init_project(cls, name, root=None, workspace=None, make_dir=True):
...
+        # when we remove backwards compatibility for project name APIs.
+        if args:
+            num_args = len(args)
+            if num_args == 1:
+                warnings.warn(
+                    "Project names were removed in signac 2.0. If you intended to call "
+                    "`init_project` with a root directory, please provided it as a keyword "
+                    f"argument: `init_project(root={args[0]})`. If your project name "
+                    "contains important information, consider storing it in the project document "
+                    "instead.",
```
csadorf, line 1531: Should we store it automatically similar to the proposed migration? I feel like simply dropping this piece of information (even with warning) is risky. I think if we actually just completely ignore this information, we need to fail instead.

signac/contrib/project.py
bdice: In this specific case, I think realpath is the best choice — it is necessary and sufficient for the check we want to perform.

signac/contrib/project.py
```diff
@@ -1481,17 +1477,15 @@ def temporary_project(self, name=None, dir=None):
...
         _mkdir_p(self.workspace())  # ensure workspace exists
-        with TemporaryProject(name=name, cls=type(self), dir=dir) as tmp_project:
+        with TemporaryProject(cls=type(self), dir=dir) as tmp_project:
             yield tmp_project
 
     @classmethod
-    def init_project(cls, name, root=None, workspace=None, make_dir=True):
-        """Initialize a project with the given name.
+    def init_project(cls, *args, root=None, workspace=None, make_dir=True, **kwargs):
+        """Initialize a project at the provided root directory.
```
bdice, line 1488:
```suggestion
        """Initialize a project in the provided root directory.
```

signac/contrib/project.py
```diff
@@ -1526,11 +1518,45 @@ def init_project(cls, name, root=None, workspace=None, make_dir=True):
...
+                    "important information, consider storing it in the project document instead.",
+                    FutureWarning,
+                )
+
         if root is None:
             root = os.getcwd()
         try:
             project = cls.get_project(root=root, search=False)
         except LookupError:
+            name = "project"
```
bdice, line 1559: It seems like this `except` block might be doing things that we no longer want to do? e.g. `config["project"] = name` and `assert project.id == str(name)`?

signac/contrib/project.py
vyasr: This PR removes the ability for a user to call the API with a project name, but it does not change the schema or remove the project name internally. If a user has already created a project with a name other than project (or if they initialize a project and then modify `signac.rc` directly), calling `signac.get_project` here will still provide the name they used. I did this in attempt to keep this PR's scope to a minimum and would prefer to leave it that way since there were questions like this one that I would prefer to address in a separate PR dedicated to logic for schema migration. Note that this PR is being merged into `next`, not `schema2`, again in an attempt to completely separate API changes from schema changes except where the two are inextricably coupled.

signac/contrib/project.py
vyasr: See https://github.com/glotzerlab/signac/pull/677#discussion_r805193191. For now, project names are always "project" upon initialization, but they do still exist and the id is well-defined, so I left those checks intact. I will remove all of that logic in a separate PR removing project names from the config.

bdice (approved):
This looks good to me!

edit: I see Python 3.6 tests are failing. As far as I'm concerned, we can remove Python 3.6 from our testing matrix. The underlying problem is that `version.parse` only has the attribute `major` for packaging >= 20.0. [We pin Python 3.6 to `packaging==15.0`.](https://github.com/glotzerlab/signac/blob/5111cde0d36eb8ee230a5c030e33fc1ae2292859/.circleci/ci-oldest-reqs.txt#L7)

https://packaging.pypa.io/en/latest/changelog.html#id14 (the linked issue is wrong, it should point to https://github.com/pypa/packaging/pull/225)

vyasr:
@bdice thanks for debugging that for me and making the change painless!

vyasr:
Looks like I spoke too soon. For such an old version of `packaging` even the `release` attribute doesn't exist. We have to manually parse the string out of `Version.public`.

bdice:
@vyasr Sorry, my suggestion wasn’t enough. The `release` attribute was added in [packaging 17.0](https://packaging.pypa.io/en/latest/changelog.html#id20).

You may need to construct a `Version` object and compare with `< Version("3.0.0dev0")` or something like that? I’m not sure if there is a better way.

bdice:
@vyasr I think the method I proposed above would be a bit less hacky with version parsing but I approve the fix either way.

signac/contrib/project.py
```diff
@@ -1526,11 +1518,46 @@ def init_project(cls, name, root=None, workspace=None, make_dir=True):
...
+                )
+        if kwargs:
+            name = kwargs.pop("name", None)
+            if kwargs:
+                # Match the usual error from extra keyword args.
+                raise TypeError(
+                    f"init_project() got an unexpected keyword argument '{next(iter(kwargs))}'"
+                )
+            if name is not None:
+                assert int(version.parse(__version__).public.split(".")[0]) < 3
```
csadorf, line 1548: Can we add a comment about why we are taking convoluted approach here? We can probably increase our minimum requirement regarding the `packaging` package at some point in which case we should refactor this.

tests/test_project.py
```diff
@@ -1051,6 +1041,43 @@ def test_temp_project(self):
         assert not os.path.isdir(tmp_root_dir)
 
 
+# Use the major version to fail tests expected to fail in 3.0.
+_MAJOR_VERSION = int(version.parse(signac.__version__).public.split(".")[0])
```
csadorf, line 1045: Same comment as above.

csadorf (changes requested):

signac/contrib/project.py
vyasr: In [this discussion](https://github.com/glotzerlab/signac/pull/643#discussion_r786069407) on #643 we decided that during a migration we would offer to automatically move the project name into the project document. In [this other discussion](https://github.com/glotzerlab/signac/pull/643#discussion_r752469873) we proposed changing the `init_project` API to warn users that the project name will be ignored when provided to the `init_project` API in signac 2.x. The goal would be to avoid destroying preexisting provenance while providing a smooth transition to the new API without breaking existing scripts. Am I interpreting these discussions correctly?

It seems like what you are proposing now is to additionally not just warn users, but to also to e.g. offer to automatically store the project name in the project document (one option for addressing the "mitigation issue" that you describe) if a user of signac 2.x calls `init_project(name=...)`. It's one thing for the migration to offer to preserve preexisting names, but should we really provide that feature on new project creation in signac 2.x? I would rather make our warning more explicit: "If the project name is important to you, consider storing it in the project document." We could offer to store project names in `init_project` as well if you think that it's particularly important, but I suspect that will encourage more users to avoid updating their usage of the API until 3.x. We would also need to continue maintaining machinery to check that the provided project name matches one in the document if it exists and then decide how to act in the case of a mismatch since it's no longer critical provenance (error? warn? overwrite? ignore?).

signac/contrib/project.py
vyasr: I think switching to this would be the most self-documenting approach.

signac/contrib/project.py
csadorf: I understand that some of the decisions we made in those past discussions are slightly contradictory. I simply did not sufficiently consider that the proposed approach with a warning is good to not break existing code, but it would also mean to dump user-provided metadata into `/dev/null`.

I would propose to _either_:

1. Keep the warning, but also store the project name in the project document under a "name" key. The warning should explicitly say that this is what happened. This should fail if the key already exists with a different value.
2. Raise an error and advise users to store the project name in the project document if they consider it important.

In principle, we should never destroy user-provided information IMO.

vyasr: No worries, I didn't mean to imply any issue with changing direction. I just wanted to make sure that we were on the same page as to what has been discussed since at this point discussions are fragmented over multiple PRs and over a reasonably long period of time.

I think option 1 is probably what we should do. My personal preference would be option 2 since it is simpler for us, but realistically I think converting `init_project(name=...)` into an error will break enough scripts that we're better off at least giving people the opportunity to move off more gradually. In practice I think a large number of users will just ignore the warning anyway, so we'll just run into the same issues when people upgrade to 3.x, but it's better than nothing.

signac/contrib/project.py
```diff
@@ -1544,20 +1539,32 @@ def init_project(cls, *args, root=None, workspace=None, make_dir=True, **kwargs)
...
             project = cls.get_project(root=root, search=False)
+            existing_name = project.doc.get(name_key)
+            if name is not None and name != existing_name:
+                raise ValueError(
+                    "The name provided to `init_project` does not match the existing "
+                    f"project document in which {name_key}={existing_name}."
+                )
         except LookupError:
-            name = "project"
+            name = name or "project"
```
csadorf, line 1567: Why don't we use `None` as the default/fallback instead of the literal `"project"`?

signac/contrib/project.py
```diff
@@ -1568,6 +1575,8 @@ def init_project(cls, *args, root=None, workspace=None, make_dir=True, **kwargs)
             config["schema_version"] = SCHEMA_VERSION
             config.write()
             project = cls.get_project(root=root)
+            if name != "project":
+                project.doc[name_key] = name
```
csadorf, line 1579: What is the expected behavior for when the user provides `"project"` as value?

signac/contrib/project.py
vyasr: Good point. No reason. Changed it.

vyasr:
@csadorf feel free to merge this if you think it's ready.

csadorf (approved):

csadorf:
Thanks @vyasr !

PR #678, opened 2022-02-07
Initial migration to schema version 2 including config rename

vyasr:
## Description

This PR contains the first components of a migration to schema version 2. The primary change in this PR is that the config file for a signac project has been moved to ".signac/config" rather than "signac.rc". ".signacrc" is no longer supported except as the global configuration file. Additionally, root discovery via `load_config` no longer checks for the presence of the "project" key in a config file to determine the project root. Any directory containing ".signac/config" file is sufficient to terminate the search and establish a project root directory.

## Motivation and Context

Contributes to #197. Part of schema 2 changes.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/common/config.py
+++ b/signac/common/config.py
@@ -12,12 +12,22 @@ from .validate import cfg, get_validator
 
 logger = logging.getLogger(__name__)
 
-PROJECT_CONFIG_FN = "signac.rc"
+PROJECT_CONFIG_FN = os.path.join(".signac", "config")
 USER_CONFIG_FN = os.path.expanduser(os.path.join("~", ".signacrc"))
 
 
 def _get_project_config_fn(root):
     return os.path.abspath(os.path.join(root, PROJECT_CONFIG_FN))
+
+
+def _get_config_dirname(fn):
+    # TODO: Is the string replace a sufficiently portable solution? We could
+    # use pathlib's `.paths` attribute, but that has significant performance
+    # implications: this takes ~200 ns, whereas
+    # os.path.join(*(pathlib.PosixPath(fn).parts[:-2])) takes ~6 us, almost a
+    # 20x slowdown. On the other hand, loading a project should not be a
+    # frequent operation that leading to bottlenecks.
+    return fn.replace(os.path.sep + PROJECT_CONFIG_FN, "")
 
 
 def _search_local(root):
@@ -127,7 +137,7 @@ def load_config(root=None, local=False):
         # should be returned separately (i.e. the return should become a tuple
         # (root_dir, config). The current approach confuses the discovery with
         # the contents of the config.
-        config["project_dir"] = os.path.dirname(fn)
+        config["project_dir"] = _get_config_dirname(fn)
         break
     return config
 
--- a/signac/common/config.py
+++ b/signac/common/config.py
@@ -21,13 +21,12 @@ def _get_project_config_fn(root):
 
 
 def _get_config_dirname(fn):
-    # TODO: Is the string replace a sufficiently portable solution? We could
-    # use pathlib's `.paths` attribute, but that has significant performance
-    # implications: this takes ~200 ns, whereas
+    # We could use pathlib's `.paths` attribute to remove the trailing project
+    # config filename, but that has significant performance
+    # implications: it takes ~200 ns, whereas
     # os.path.join(*(pathlib.PosixPath(fn).parts[:-2])) takes ~6 us, almost a
-    # 20x slowdown. On the other hand, loading a project should not be a
-    # frequent operation that leading to bottlenecks.
-    return fn.replace(os.path.sep + PROJECT_CONFIG_FN, "")
+    # 20x slowdown.
+    return fn.replace(os.sep + PROJECT_CONFIG_FN, "")
 
 
 def _search_local(root):
--- a/signac/common/config.py
+++ b/signac/common/config.py
@@ -12,18 +12,14 @@ from .validate import cfg, get_validator
 
 logger = logging.getLogger(__name__)
 
-DEFAULT_FILENAME = ".signacrc"
-CONFIG_FILENAMES = [DEFAULT_FILENAME, "signac.rc"]
-HOME = os.path.expanduser("~")
-CONFIG_PATH = [HOME]
-FN_CONFIG = os.path.expanduser("~/.signacrc")
+PROJECT_CONFIG_FN = "signac.rc"
+USER_CONFIG_FN = os.path.expanduser(os.path.join("~", ".signacrc"))
 
 
 def _search_local(root):
-    for fn in CONFIG_FILENAMES:
-        fn_ = os.path.abspath(os.path.join(root, fn))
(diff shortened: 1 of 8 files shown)


vyasr:
Reviewers can ignore the pre-commit failures. Those are fixed on #677 and will be resolved once those changes are merged down to here.

signac/common/config.py
```diff
@@ -15,18 +15,28 @@
...
+
+
+def _get_config_dirname(fn):
+    # TODO: Is the string replace a sufficiently portable solution? We could
+    # use pathlib's `.paths` attribute, but that has significant performance
+    # implications: this takes ~200 ns, whereas
+    # os.path.join(*(pathlib.PosixPath(fn).parts[:-2])) takes ~6 us, almost a
+    # 20x slowdown. On the other hand, loading a project should not be a
+    # frequent operation that leading to bottlenecks.
+    return fn.replace(os.path.sep + PROJECT_CONFIG_FN, "")
```
bdice, line 33: We use `os.sep` elsewhere, so I'd recommend that as a canonical name. It's [imported](https://github.com/python/cpython/blob/9539400390494f4930c245b3f98453592f4a1a8c/Lib/os.py#L96) from `os.path`, though. 🤷
```suggestion
    return fn.replace(os.sep + PROJECT_CONFIG_FN, "")
```

signac/common/config.py
```diff
@@ -15,18 +15,28 @@
...
+PROJECT_CONFIG_FN = os.path.join(".signac", "config")
+USER_CONFIG_FN = os.path.expanduser(os.path.join("~", ".signacrc"))
+
+
+def _get_project_config_fn(root):
+    return os.path.abspath(os.path.join(root, PROJECT_CONFIG_FN))
+
+
+def _get_config_dirname(fn):
+    # TODO: Is the string replace a sufficiently portable solution? We could
```
bdice, line 27: I have no issue with this approach. Just change the TODO to an explanation of the choice for performance reasons.

signac/common/config.py
```diff
@@ -78,10 +90,11 @@ def read_config_file(filename, configspec=None, *args, **kwargs):
     try:
         config = Config(filename, configspec=configspec, *args, **kwargs)
     except (OSError, ConfigObjError) as error:
-        msg = "Failed to read configuration file '{}':\n{}"
-        raise ConfigError(msg.format(filename, error))
+        raise ConfigError(f"Failed to read configuration file '{filename}':\n{error}")
     verification = config.verify()
     if verification is not True:
+        # TODO: In the future this should raise an error, not just a
```
bdice, line 96: Is this a "2.0 requirement" future or some arbitrary future? Is there a problem with raising now?

signac/contrib/migration/v1_to_v2.py
```diff
@@ -0,0 +1,39 @@
+# Copyright (c) 2022 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+"""Migrate from schema version 1 to version 2.
+
+This migration involves the following changes:
```
bdice, line 6: Probably need to fill out some bullet points here.

tests/test_project.py
```diff
@@ -2458,7 +2463,14 @@ def test_project_schema_versions(self):
     def test_project_schema_version_migration(self, implicit_version):
         from signac.contrib.migration import apply_migrations
 
-        config = read_config_file(self.project.fn("signac.rc"))
+        # First need to move the v2 config file to a suitable v0/v1 file.
```
bdice, line 2466: Schema version 0 doesn't exist except for our testing purposes, right? I think we only have schema version v1 in the wild.

bdice: I think I'd rather hard-code a config here and use a specific filename in a temporary directory, without reference to the existing v2 schema or any other project initialization logic. That is, mock the schema v1 project construction with a few lines of Python so that we don't have to engineer a "roll-back" approach for testing future schema versions and we have something constant that lives outside our signac implementation. Roughly like:
```python
with TemporaryDirectory():
    with open("signac.rc", "w") as f:
        f.write(textwrap.dedent("""\
            project = asdf
            workspace_dir = workspace
            schema_version = 1"""
        )
```

bdice (commented):
I have some feedback. Overall this is a good set of changes. Thanks for making smaller PRs, it's much easier to take bites for review. Ping me for a second round when you're ready.

signac/common/config.py
vyasr: Sure

signac/common/config.py
vyasr: This is a "out of scope for this PR" TODO. I don't plan to address this until switching to TOML, no point doing it with the configobj API since TOML is likely to look somewhat different anyway.

signac/contrib/migration/v1_to_v2.py
vyasr: I originally had something here but figured I would update it at the end. No harm in building it up incrementally though.

tests/test_project.py
vyasr: > I think we only have schema version v1 in the wild.
Yeah there's no real version 0. The default version assigned when a project doesn't have a version is 1.

> I think I'd rather hard-code a config here and use a specific filename in a temporary directory
Good call, I agree that would be better.

signac/common/config.py
```diff
@@ -12,18 +12,27 @@
...
+
+
+def _get_project_config_fn(root):
+    return os.path.abspath(os.path.join(root, PROJECT_CONFIG_FN))
+
+
+def _get_config_dirname(fn):
+    # We could use pathlib's `.paths` attribute to remove the trailing project
+    # config filename, but that has significant performance
+    # implications: it takes ~200 ns, whereas
```
bdice, line 26:
```suggestion
    # implications: this takes ~200 ns, whereas
```

tests/test_project.py
```diff
@@ -2490,6 +2474,45 @@ def test_no_migration(self):
...
+                    )
+                )
+
+            config = read_config_file(cfg_fn)
+            print(config)
+            if implicit_version:
+                del config["schema_version"]
+                assert "schema_version" not in config
+            else:
+                config["schema_version"] = "0"
```
bdice, line 2500: Add a comment that this block is referring to schema version 0. It is a little confusing to have the string defined with version 1 only to revert here.

tests/test_project.py
```diff
@@ -2490,6 +2474,45 @@ def test_no_migration(self):
...
+                    textwrap.dedent(
+                        """\
+                        project = project
+                        workspace_dir = workspace
+                        schema_version = 1"""
+                    )
+                )
+
+            config = read_config_file(cfg_fn)
+            print(config)
```
bdice, line 2495: Leftover debug print?

tests/test_project.py
vyasr: I'll just define with version 0 to start with. No reason not to.

PR #680, opened 2022-02-12
Changed code to use id_ instead of _id

ParisaTork:
I changed the codebase to use id_ instead of _id.

## Motivation and Context
This helps to avoid shadowing builtins.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

N.B. this is a refactor.

## Checklist:

If necessary:

### DIFF
--- a/benchmark.py
+++ b/benchmark.py
@@ -361,11 +361,11 @@ def read_benchmark(filename, filter=None, include_metadata=False):
         docs = list(c.find(filter))
 
     df_data = pd.DataFrame(
-        {doc["_id"]: dict(normalize(doc["data"], doc["meta"]["N"])) for doc in docs}
+        {doc["id_"]: dict(normalize(doc["data"], doc["meta"]["N"])) for doc in docs}
     ).T
 
     if include_metadata:
-        df_meta = pd.DataFrame({doc["_id"]: doc["meta"] for doc in docs}).T
+        df_meta = pd.DataFrame({doc["id_"]: doc["meta"] for doc in docs}).T
         return pd.concat([df_meta, df_data], axis=1)
     else:
         return df_data
--- a/changelog.txt
+++ b/changelog.txt
@@ -340,7 +340,7 @@ Added
  - Adds the ``signac.Project.temporary_project()`` context manager which creates a temporary project within the root project workspace.
  - Add ``signac`` to the default namespace when invoking ``signac shell``.
  - Add option to specify a custom view path for the ``signac view``/ ``Project.create_linked_view()`` function.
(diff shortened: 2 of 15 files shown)


bdice:
@ParisaTork I think you might need to pull and rebase this PR on `next`? It appears there are merge conflicts.

ParisaTork:
@bdice Ah I see - is that not already the case: ParisaTork wants to merge 1 commit into [glotzerlab:next] from [ParisaTork:679]? If not, how do I achieve this?

vyasr:
> @bdice Ah I see - is that not already the case: ParisaTork wants to merge 1 commit into [glotzerlab:next] from [ParisaTork:679]? If not, how do I achieve this?

Hi @ParisaTork, what you've pasted above is showing that you have created a PR that adds a commit to next, which is indeed what we want. The problem is that when you created the change you created your branch `ParisaTork:679` by branching off of the `master` branch. I assume that happened because the `master` branch is the default branch, so when you `git clone` the repo you start off on `master` and you just created your new branch from there. In order to make your changes compliant, you have three options. I've listed them in decreasing order of git familiarity that they require since I'm not sure how comfortable that you are with git. The easier options that require less knowledge of git also require more manual work, so I'll leave it to you to decide what the right tradeoff is for you. Note that both options 2 and 3 will require that you create a new PR. Let us know if you have any more questions, and sorry for the confusion here!

1) Try to rebase your changes as @bdice suggested. Depending on your familiarity with git this option may be confusing to you, but it is also the option that takes the most advantage of the tool to minimize the amount of work you have to redo. Depending on your level of confidence in using git, option 2 or 3 may be the easier route, but based on the conflicts above I think that your main issue is that you changed a number of files that exist on `master` but have since been removed on `next`, so if those are the bulk of your issues then the rebase may not be too hard.
2) Check out the `next` branch locally, then create a new branch via `git checkout -b issue_679` and then cherry pick your changes using `git cherry-pick 2be8901b5598a7d10ebdfd34223afbb4b282d980` (where 2be8901b5598a7d10ebdfd34223afbb4b282d980 is the commit that you pushed in this PR). This option will attempt to make the exact same changes that you already made, but as in option 1 you will need to resolve conflicts.
3) Check out the `next` branch locally, then create a new branch via `git checkout -b issue_679` and then manually reapply all the changes that you made in this PR.

ParisaTork:
Thanks for the really handy explanation @vyasr! :) I ran a lot of git commands yesterday to fix this, but thought it would be best to start all over again - delete my fork and local copies and start all over again - I've created a new PR at https://github.com/glotzerlab/signac/pull/681 so hopefully this should sort things out! Thank you so much for your help! :)

PR #681, opened 2022-02-13
New, fixed PR: replaces _id with id_

### LINKED ISSUE #679: Use `id_` instead of `_id`.
@vyasr [noted](https://github.com/glotzerlab/signac/pull/667#discussion_r805197076) that we use `_id` instead of `id_`: "A subsequent underscore is a more canonical way to avoid shadowing builtins, as opposed to a preceding underscore indicating a private variable."

I agree with making this change across our entire codebase in a single PR on the `next` branch before the 2.0 release. We are currently consistent about using `_id` everywhere.

For context, this is to avoid shadowing the builtin [`id`](https://docs.python.org/3/library/functions.html#id) even though `_id` disagrees with [PEP8's naming conventions](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles) which would indicate to use `id_`.

ParisaTork:
I changed the codebase to use id_ instead of _id. N.B. this is the new, fixed PR! Thanks @vyasr :) 

## Motivation and Context
This helps to avoid shadowing builtins.

(edit by @bdice: this resolves #679)

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

N.B. this is a refactor.

## Checklist:

If necessary:


bdice:
I pushed a commit to trigger CI. I am not sure exactly why CI didn't trigger the first time -- I am guessing it might have happened due to something weird with the fork being deleted and recreated, so I disabled and re-enabled building PRs from forks. It seems that pushing the empty commit successfully triggered CI.

I haven't had a chance to review the PR yet but will take a look when I am able.

signac/contrib/job.py
```diff
@@ -90,7 +90,7 @@ def _save(self):
         # All elements of the job list are shallow copies of each other, so any
         # one of them is representative.
         job = next(iter(self._jobs))
-        old_id = job._id
+        old_id = job.id_
```
bdice, line 93: We may need to revert some of the changes in this file, like `job.id_` back to `job._id` and `self.id_` back to `self._id`. That is meant to be a private attribute because of the leading `_`. Also attribute names do not shadow the builtin `id`. cc: @vyasr for thoughts on this.

ParisaTork:
Great! Thank you so much @bdice! :)

signac/contrib/project.py
```diff
@@ -119,7 +119,7 @@ def __init__(self, config=None):
 
         # Ensure that the project id is configured.
         try:
-            self._id = str(self.config["project"])
+            self.id_ = str(self.config["project"])
```
bdice, line 122: Similar to the job, this private attribute may need to be changed back from `self.id_` to `self._id`. I think the cases that we want to rename are mostly local variables or function parameters named `_id`.

signac/contrib/collection.py
```diff
@@ -417,7 +417,7 @@ class Collection:
 
     """
 
-    def __init__(self, docs=None, primary_key="_id", compresslevel=0, _trust=False):
+    def __init__(self, docs=None, primary_key="id_", compresslevel=0, _trust=False):
```
bdice, line 420: I'm not sure if we want to rename the primary key here -- we may want to revert the changes to strings named `"_id"`. I'd like to consult @vyasr on this, since it was his suggestion to rename the variable.

bdice (commented):
Some preliminary thoughts -- @vyasr please take a look when you can, there are a few ambiguous cases that might warrant further consideration of the premise of this renaming. We use the name `_id` in a few unique ways:
- As a local variable representing a job's id string.
  - This is the case that I think @vyasr was targeting, and all of these could be changed to `id_`.
- As a string key in places like `Collection`.
  - This is potentially a lesser issue if `Collection` is removed in signac 2.0 -- but still arises in a few places.
- As a private attribute like `Job._id` and `Project._id`
  - These cases should not be changed, because the attributes are denoted as private by the leading underscore. Also, attribute names cannot shadow the builtin `id`.
- As a function/method parameter name
  - Changing this is an API break and while it could be done in signac 2.0, I don't see a net benefit to breaking these APIs for adhering to what I view as a fairly low priority guideline on variable naming. That said, I would be okay with deferring to others' opinions on this.

signac/contrib/collection.py
vyasr: I agree, in strings like this I think we can leave this as is. I was primarily concerned with variables as @bdice indicated.

signac/contrib/job.py
vyasr: Agreed, in this case the leading underscore is correctly indicating the intent of making a variable internal ("private")

vyasr:
@ParisaTork I generally agree with everything @bdice said. We mainly want to rename local variables or function parameters for internal functions. The class attributes like `Job._id` should not be changed, and I don't think we need to bother changing anything in strings either.

The one place I would differ from @bdice is regarding API breaks, not because I think we should break a lot of APIs but because I don't see any meaningful ones. Almost all of the function signatures where the variable name was changed from `_id` to `id_` are internal functions or dunder methods, so the names of the parameters are either not part of our API or aren't even accessible (e.g. changing __contains__). The only public API that I see changing is the signature of the `Job` constructor, and since our users never actually use that constructor (we always advocate `Project.open_job` instead) I think that's a negligible API break for 2.0.

vyasr:
@ParisaTork I saw that you requested review, did you see my last comment? There are some changes that need to be made to this PR, do you need some clarification on what we were asking for?

ParisaTork:
@vyasr Thank you so much for the review - I will act upon the ones you and @bdice have commented on and get the new PR sorted for you tomorrow :)

ParisaTork:
The post-review changes are up :)

signac/contrib/project.py
```diff
@@ -834,13 +834,13 @@ def _register(self, _id, statepoint):
 
         Parameters
         ----------
-        _id : str
+        id_ : str
```
lyrivera, line 837: the actual id argument of the `_register` method was not changed here

tests/test_project.py
```diff
@@ -2440,7 +2440,7 @@ def test_get_job_symlink_other_project(self):
         job_a.init()
         job_b = project_b.open_job({"b": 1})
         job_b.init()
-        symlink_path = os.path.join(project_b.workspace(), job_a._id)
+        symlink_path = os.path.join(project_b.workspace(), job_a.id_)
         os.symlink(job_a.ws, symlink_path)
```
lyrivera, line 2444: I think this one should be reverted

signac/contrib/job.py
```diff
@@ -224,7 +224,7 @@ class Job:
     Application developers should not directly instantiate this class, but
     use :meth:`~signac.Project.open_job` instead.
 
-    Jobs can be opened by ``statepoint`` or ``_id``. If both values are
+    Jobs can be opened by ``statepoint`` or ``id_``. If both values are
```
lyrivera, line 227: There should probably be a changelog entry to notify of this change.

lyrivera (commented):
Left a few comments based on the discussion.

vyasr:
@ParisaTork how are things going? It looks like you need to merge in the next branch and address the open reviews before we can move this forward.

signac/contrib/project.py
ParisaTork: @lyrivera Just to clarify, would you like _id in the function signature to be changed to id_?

signac/contrib/job.py
ParisaTork: @lyrivera How would you like the changelog entry to be implemented?

ParisaTork:
@vyasr @lyrivera Thank you both for your comments - just wanted to clarify a few things before making any changes and I've left some comments above in the corresponding code above.

signac/contrib/project.py
ParisaTork: Thank you very much - this has been sorted :)

tests/test_project.py
ParisaTork: Thank you very much - this has been sorted :)

ParisaTork:
I've sorted 2 of 3 issues above - How would you like the changelog entry to be implemented?

signac/contrib/job.py
lyrivera: @ParisaTork a comment on `changelog.txt` notifying users that the keyword `_id` of the `Job` constructor has been renamed to `id_` should do

ParisaTork: Thank you so much - this has been sorted

ParisaTork:
I've now implemented all the changes requested in the previous code review :)

tests/test_collection.py
```diff
@@ -120,7 +120,7 @@ def test_insert(self):
         assert self.c["0"] == doc
         assert list(self.c.find()) == [doc]
         with pytest.raises(ValueError):
-            self.c["0"] = dict(_id="1")
+            self.c["0"] = dict(id_="1")
```
bdice, line 123: Let's revert the changes in this file. We are going to remove the `Collection` class and its tests soon, but these tests probably rely on `_id` being the primary key defined here: https://github.com/glotzerlab/signac/blob/c2a481ef5f144bad27b2a856c267a63294bff99b/signac/contrib/collection.py#L420

bdice (changes requested):
Looking good so far. We need to revert changes to `test_collection.py`. We're planning to remove `Collection` in signac 2.0, so we won't need to update that file.

tests/test_collection.py
ParisaTork: Fantastic - that's all been sorted

bdice:
@vyasr It looks like a force-push to `next` did something unexpected to this PR. I can try to sort it out. We should try to finalize and merge this PR before we force-push to `next` again.

bdice:
I didn't meant to close this PR. GitHub automatically closed it when I force pushed to `ParisaTork:next` with a fixed/rebased set of changes...

bdice:
I don't have permissions to force-push to that fork, unfortunately. I am opening a new PR.

mikemhenry:
I run into this sometimes as well, it makes me question my git-foo and the github engineers git-foo

PR #683, opened 2022-02-18
Remove Collection

### LINKED ISSUE #664: Optimize `Collection` for internal use in `Project`
A possible optimization for signac 2.0 (or later) would be to reduce the amount of code we use in the `Collection` class when calling `Project.find_jobs`. See: https://github.com/glotzerlab/signac/blob/b29e0485c7998ba0c6d041e9ec15b533334d9b64/signac/contrib/project.py#L717

We copy a large amount of data used from the `Project`'s internal caches during the construction of a `Collection` and calling its `find` method. At the very least, we never use the `Collection`'s ability to interact with data on disk or its ability to automatically generate ids (primary keys) for new records in the context of a signac `Project`, so we could eliminate some logic there in a cut-down class.

_Originally posted by @bdice in https://github.com/glotzerlab/signac/issues/652#issuecomment-1002303783_

I attempted an optimization in 13f2a8fb205f65d49421f4f4009ee3d78d00f9bf but it was [unclear if copy/reference semantics would be correct in the resulting indices](https://github.com/glotzerlab/signac/pull/652#discussion_r777170033) outside the context of a signac `Project`'s limited usage of `Collection`. A smaller class that is designed for the actual use case of signac's `Project.find_jobs` could act as an internal cache with only the necessary logic (e.g. no file I/O or id generation).

bdice:
## Description
Follow-up from #667 to remove the `Collection` class from the public API of signac.

## Motivation and Context
Resolves #664. A smaller public API is helpful for maintenance.

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


changelog.txt
```diff
@@ -19,6 +19,7 @@ Removed
 +++++++
 
  - The ``Project.config`` property is no longer mutable. Use the command line ``$ signac config`` to modify configuration (#608, #246, #244).
+ - The ``Collection`` class has been removed.
```
bdice, line 22:
```suggestion
 - The ``Collection`` class has been removed (#664, #667, #683).
```

bdice:
We'll need to fix the CI benchmarks to use asv first. I'll do that as a PR to `master` and then merge it up into `next` and this branch.

vyasr:
This PR is blocked by #704.

vyasr:
This needs to have next merged in. But also, the blocker hasn't been resolved yet. What are we doing about benchmarks going forward?

bdice:
> What are we doing about benchmarks going forward?

I propose removing `benchmark.py` and not running benchmarks in CI -- they're too noisy to trust. We can use asv to run comparison benchmarks offline, when desired.

bdice:
@vyasr I removed `benchmark.py` and rebased on `next`.

setup.cfg
```diff
@@ -17,7 +17,7 @@ select = E,F,W
 ignore = E123,E126,E203,E226,E241,E704,W503,W504
 
 [pydocstyle]
-match = ^((?!\.sync-zenodo-metadata|setup|benchmark).)*\.py$
+match = ^((?!\.sync-zenodo-metadata|setup).)*\.py$
```
vyasr, line 20:
```suggestion
match = ^((?!\.sync-zenodo-metadata|setup|benchmark).)*\.py$
```

vyasr: We need to leave this ignore in so that pydocstyle doesn't look at the asv benchmarks.

csadorf (approved):

PR #685, opened 2022-02-21
Additional deprecations

vyasr:
## Description

This PR contains a number of deprecations as well as TODO comments indicated _proposed_ deprecations. I have intentionally been a little bit aggressive with some of the proposals to spark discussion on whether certain APIs should continue to exist in the long term or not.

## Motivation and Context

Aiming to slim down the API as much as possible in 2.0 as well as planning for future developments.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:


signac/contrib/project.py
```diff
@@ -260,9 +260,7 @@ def __init__(self, config=None):
         self._lock = RLock()
 
         # Prepare cached properties derived from the project config.
-        self._id = None
-        self._rd = None
-        self._wd = None
+        _invalidate_config_cache(self)
```
csadorf, line 263: Is this a deprecation?

signac/contrib/schema.py
```diff
@@ -332,6 +330,8 @@ def difference(self, other, ignore_values=False):
             )
         return ret
 
+    # TODO: We don't use this method anywhere that I see. Is there a particular use case that is
+    # facilitated by keeping it, or should we deprecate?
```
csadorf, line 334: It was probably used to implement this functionality at some point:
https://github.com/glotzerlab/signac/blob/c291e1ef1d8cc7bb3b9591ae4e6265f4c27cf7aa/signac/contrib/project.py#L923-L925

signac/contrib/schema.py
```diff
@@ -267,43 +274,34 @@ def _repr_html_(self):
...
         return output
 
+    # TODO: This method can be removed in signac 2.0 once support for list keys
+    # is removed.
     def __contains__(self, key_or_keys):
-        if isinstance(key_or_keys, str):
-            return key_or_keys in self._schema
-        key_or_keys = ".".join(key_or_keys)
+        # NotOverride default __contains__ to support sequence and str inputs.
+        if not isinstance(key_or_keys, str):
```
csadorf, line 281:
```suggestion
        assert version.major < 2
        if not isinstance(key_or_keys, str):
```

signac/contrib/schema.py
```diff
@@ -267,43 +274,34 @@ def _repr_html_(self):
...
+                FutureWarning,
+            )
+            key_or_keys = ".".join(key_or_keys)
         return key_or_keys in self._schema
 
     def __getitem__(self, key_or_keys):
-        if isinstance(key_or_keys, str):
-            return self._schema[key_or_keys]
-        return self._schema[".".join(key_or_keys)]
+        if not isinstance(key_or_keys, str):
```
csadorf, line 292:
```suggestion
        assert version.major < 2
        if not isinstance(key_or_keys, str):
```

signac/testing.py
```diff
@@ -5,6 +5,8 @@
 from itertools import cycle
 
 
+# TODO: This feels like something that should be a test fixture or
+# fixture-adjacent, not code that belongs in the main package.
```
csadorf, line 9: Tend to disagree on this one. There is value in user-facing code to easily generate a sample project. The fixture code would necessarily be shipped with the package.

It's also used here: https://github.com/glotzerlab/signac-flow/blob/4d49e4f72e8d2201fa27efe2fbb3b0940a31ca79/flow/testing.py#L35

signac/contrib/job.py
```diff
@@ -359,6 +359,14 @@ def _statepoint_filename(self):
...
         return os.sep.join((self.workspace(), self.FN_MANIFEST))
 
+    # TODO: Why is ws a property while workspace is a method? Why don't the
+    # Project methods for workspace/root_directory also behave the same? We
+    # should standardize this.
+    # TODO: The discrepancy between Project and Job in what is defined as a
+    # "workspace" is also slightly confusing. In view of moving towards a
+    # unified 3.0 API where both are subclasses of a common Directory or
+    # similar, we should aim to unify these APIs (although probably with an
+    # extremely long deprecation period like the entire lifetime of signac 2.0)
```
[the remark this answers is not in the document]
vyasr: Right, my understanding is that workspace being a function is a historical artifact rather than an indication of how we would want to design in now. That is addressing the first note here, but not the second. Apologies for sticking two separate ideas immediately next to each other like this.

The second question I'm asking here is, rather than converting workspace into a property, should we consider defining a new property `Job.root` (or `Job.path` or similar), mimicking it in `Project`, and then deprecating `Job.workspace`, `Job.ws`, and `Project.root_directory`? The reason I ask this is that we currently have different names for the path to a `Job` and the path to a `Project`, and even worse, `Project.workspace` means something semantically different from `Job.workspace`. The historical reasons for these names make sense, but now would be a good time to try and choose something more canonically appealing like `path`. `Project.workspace` would still exist since that has a specific meaning in the context of signac's (current) data model, but the other properties would then be aligned.

This change would also be in line with our original signac 2.0 proposal to enable different folder layouts via different data models, which would be greatly facilitated by moving towards a common `signac.Directory` class as a common parent for `Job` and `Project` that implements this `path`/`root` interface.

signac/contrib/job.py
```diff
@@ -483,6 +491,11 @@ def statepoint(self):
 
         return self._statepoint
 
+    # TODO: If we really believe what `reset_statepoint`'s documentation says,
+    # I think we should deprecate this functionality. If resetting a statepoint
+    # is a dangerous operation, making it settable via a property is too easy,
+    # and moreover it's impossible to document effectively because property
+    # setters don't show up in docs.
```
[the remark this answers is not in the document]
vyasr: > Just because something is potentially destructive does not mean it shouldn't exist.
Sure, but many APIs are designed to avoid providing users with foot guns unless they have demonstrable value. In this case I agree that setting a statepoint is a potentially useful operation, and I'm not advocating for making it impossible. However, it is inconsistent to mark `reset_statepoint` as "dangerous" while considering it obvious that `job.statepoint = ...` is destructive. I'm advocating for more consistency in the level of user protection between these two, and consequently making it slightly harder for a user to shoot themselves in the foot. By forcing them to call a method that they'd probably have to look up, we increase the chances that they see a documented warning.

I don't feel terribly strongly on this issue and would settle for updating the docstring of the `statepoint` property (it would have to be the getter since setter docstrings don't show up anywhere) to include the warning instead if you would rather leave the setter.

signac/contrib/project.py
```diff
@@ -2625,17 +2619,19 @@ def __iter__(self):
             self._project, self._project._find_job_ids(self._filter)
         )
 
+    @deprecated(
+        deprecated_in="1.8",
```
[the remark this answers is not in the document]
vyasr: I wasn't sure what number to put there since the old warning has apparently been there forever and nobody saw it (including us). But I agree, we can use the old version number even though clearly nobody saw that deprecation as long as we wait to remove until 2.0.

signac/contrib/project.py
vyasr: Ah no sorry this is just making use of an existing API that it probably already should have been. I snuck that in here while I was making these changes but it's not related.

signac/testing.py
vyasr: I agree in principle. In practice, I don't know if anyone has ever used this. I certainly haven't. Perhaps we should be doing a better job publicizing its existence if we think it could help users with prototyping something?

signac/contrib/job.py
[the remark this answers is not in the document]
bdice: I could buy into the `signac.Directory` (abstract?) class providing a `path` property.

Regardless of naming or plans for 3.0, we should definitely change the methods that should be properties in 2.0.

signac/contrib/job.py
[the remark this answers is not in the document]
bdice: I'd be in favor of a solution that reduces the number of equivalent ways to do things.

> [There should be one-- and preferably only one --obvious way to do it.](https://www.python.org/dev/peps/pep-0020/)

That would be one of the following:
- Deprecate (1.x) and remove (2.0) `reset_statepoint`, move the cautionary statements in the docstring to the `statepoint` property. I prefer this approach, even though it may reduce the likelihood of users finding the cautionary statements in the docs. Honestly, I still can't remember what `reset_statepoint` does from the function name alone. It's one of my least favorite API names, and I strictly use the `@statepoint.setter` in practice.
- Deprecate (1.x) and remove (2.0) `@statepoint.setter` and require the use of `reset_statepoint`.

signac/contrib/project.py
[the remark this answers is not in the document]
bdice: This is correct. I'll apply @csadorf's suggestion.

We did not remove `next()` in 1.0.0 because it had been deprecated very recently. Source: https://github.com/glotzerlab/signac/pull/139

signac/testing.py
bdice: I have always wanted to replicate this kind of thing (shipping helpful test fixtures) in more packages I work on, but have never gotten around to it. I believe I tried using this feature in the benchmarking code, but it's not meant to generate large projects suitable for benchmarking -- only small projects with complex schemas. We might want to revisit that at some point, and try to unify those two fixture-like code paths.

tests/test_project.py
```diff
@@ -841,16 +841,19 @@ def test_schema(self):
         assert len(s) == 9
         for k in "const", "const2.const3", "a", "b.b2", "c", "d", "e.e2", "f.f2":
             assert k in s
-            assert k.split(".") in s
+            with pytest.warns(FutureWarning):
+                assert k.split(".") in s
```
bdice, line 845: This should only warn if the key contains a dot, right? Conditional context managers are somewhat awkward. Luckily, we only need to test this for split keys if the combined key contains a dot -- otherwise the code is equivalent to the test above, `assert k in s`.
```suggestion
            if "." in k:
                with pytest.warns(FutureWarning):
                    assert k.split(".") in s
```

tests/test_project.py
```diff
@@ -841,16 +841,19 @@ def test_schema(self):
...
             # The following calls should not error out.
             s[k]
-            s[k.split(".")]
+            with pytest.warns(FutureWarning):
+                s[k.split(".")]
```
bdice, line 849:
```suggestion
            if "." in k:
                with pytest.warns(FutureWarning):
                    s[k.split(".")]
```

bdice (commented):
Comments attached. Thanks so much for helping with this! This takes off a portion of the [`requires-1.x-deprecation` label](https://github.com/glotzerlab/signac/issues?q=label%3Arequires-1.x-deprecation). I'm not sure what's left. Maybe we should add the remaining ones to a milestone, or remove the label for the ones that are complete.

signac/testing.py
bdice: It looks like this function is undocumented.

signac/contrib/job.py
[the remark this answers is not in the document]
csadorf: I've outlined all of these ideas in the 2.0 API I proposed at some point. The unification of Project and Job to Directory, those being basically Paths with some additional properties and so on. Are those concepts considered here? Should I link them again?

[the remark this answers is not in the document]
bdice: I’d appreciate a link. It’s been a long time since I’ve seen that document and I’m not sure where it lives.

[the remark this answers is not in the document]
vyasr: > Are those concepts considered here? Should I link them again?

@csadorf That's what I was referring to with this:

> This change would also be in line with our original signac 2.0 proposal to enable different folder layouts via different data models, which would be greatly facilitated by moving towards a common signac.Directory class as a common parent for Job and Project that implements this path/root interface.

Is that not what you're asking about? The whole point of my proposing these changes now is to make the necessary breaking changes in 2.0 so that we could implement (at least some of) the components of the original 2.0 proposal in minor releases along the 2.x line rather than needing to make a breaking major release.

@bdice here's [the link](https://github.com/glotzerlab/signac-2-prototype/). I intended to bring up long-term roadmap discussions at the next dev meeting anyway, so I can outline some ideas when I do.

signac/contrib/job.py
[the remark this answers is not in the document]
vyasr: Exactly, that's where I'm coming from as well. We don't need both of these, and the way they're documented makes them seem different. I would be fine with keeping the setter instead of `reset_statepoint`, since I agree that it should be fairly obvious that this operation could impact the data space.

signac/testing.py
csadorf: @bdice I'm not sure I fully understand where you fall on this based on your comments. Are you recommending to keep this, but maybe document and improve? Or would you rather remove it?

bdice: I think it’s a nice feature. Keep it, document it, and improve it.

signac/contrib/job.py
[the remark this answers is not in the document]
csadorf: > Is that not what you're asking about? The whole point of my proposing these changes now is to make the necessary breaking changes in 2.0 so that we could implement (at least some of) the components of the original 2.0 proposal in minor releases along the 2.x line rather than needing to make a breaking major release.

@vyasr Yeah, I think I got my wires crossed on this one.

signac/testing.py
csadorf: Ok, ideally we use these exact functions as part of the pytest fixtures.

vyasr: OK I'll drop this in this PR and make an issue to improve this.

signac/contrib/schema.py
vyasr: I should have clarified. Yes, there is a test for it so coverage is green, but it's not used anywhere in the actual code base. I'm going to take this as approval for deprecating though :)

signac/contrib/schema.py
```diff
@@ -42,6 +43,9 @@ def _collect_by_type(values):
     return values_by_type
 
 
+# TODO: This function feels slightly out of place since it's never used here while it is used in
+# import_export.py and project.py, but I see why it fits as "schema". May still make sense to move,
+# though.
```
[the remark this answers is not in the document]
vyasr: Right, good point. I didn't remember that when I wrote this. I think this comment is probably already sufficiently resolved, then.

signac/contrib/job.py
```diff
@@ -359,11 +355,36 @@ def _statepoint_filename(self):
         # use str.join with os.sep instead of os.path.join for speed.
         return os.sep.join((self.workspace(), self.FN_MANIFEST))
 
-    @property
+    # Decorated properties aren't supported: https://github.com/python/mypy/issues/1362
+    @property  # type: ignore
```
bdice, line 359: Can this be decorated in the opposite order? Aside from the mypy issue, it feels like putting `@property` closer to `def ws(self):` would be clearer.

signac/contrib/job.py
```diff
@@ -359,11 +355,36 @@ def _statepoint_filename(self):
...
-        """Alias for :meth:`~Job.workspace`."""
-        return self.workspace()
+        """Alias for :meth:`~Job.path`."""
+        return self.path
 
+    @property
+    def path(self):
+        """str: The path to the job directory.
+
+        See :ref:`signac job -w <signac-cli-job>` for the command line equivalent.
```
bdice, line 374: Do we need to deprecate/update the CLI flag name to `-p/--path`?

signac/contrib/job.py
```diff
@@ -359,11 +355,36 @@ def _statepoint_filename(self):
...
+        """
+        if self._wd is None:
```
bdice, line 376: I'd like to rename `self._wd` to `self._path` to match.

signac/contrib/job.py
```diff
@@ -500,6 +531,7 @@ def sp(self):
         """Alias for :attr:`~Job.statepoint`."""
         return self.statepoint
 
+    # TODO: Same concern as statepoint setter.
```
bdice, line 534: This TODO can be removed now, I think?

signac/contrib/project.py
```diff
@@ -353,15 +354,19 @@ def config(self):
...
+        removed_in="2.0",
+        current_version=__version__,
+        details="Use Project.path instead.",
+    )
     def root_directory(self):
-        """Return the project's root directory.
+        """Alias for :meth:`~Project.path`."""
```
bdice, line 364: Double-check if `:meth:` or `:attr:` is the correct way to refer to properties. I can't remember.

signac/contrib/project.py
```diff
@@ -353,15 +354,19 @@ def config(self):
...
-        Returns
-        -------
-        str
-            Path of project directory.
-
-        """
+    @property
+    def path(self):
+        """str: The path to the project directory."""
         if self._rd is None:
```
bdice, line 370: Rename `self._rd` to `self._path`.

signac/contrib/project.py
```diff
@@ -766,15 +766,16 @@ def num_jobs(self):
             Count of initialized jobs.
 
         """
+        return len(self)
+
+    def __len__(self):
```
bdice, line 771: Thanks for using a `def` here, that's clearer.

signac/contrib/project.py
```diff
@@ -790,7 +791,7 @@ def _contains_job_id(self, job_id):
 
         """
         # We can rely on the project workspace to be well-formed, so just use
-        # str.join with os.sep instead of os.path.join for speed.
+        # os.sep.join instead of os.path.join for speed.
         return os.path.exists(os.sep.join((self.workspace(), job_id)))
```
bdice, line 795: Will `self.workspace()` become a property instead of a method? How are we handling that?

signac/contrib/project.py
```diff
@@ -2638,17 +2641,19 @@ def __iter__(self):
...
+    @deprecated(
+        deprecated_in="0.9.6",
+        removed_in="2.0",
+        current_version=__version__,
+        details="Use next(iter(...)) instead.",
+    )
     def next(self):
         """Return the next element.
 
         This function is deprecated. Users should use ``next(iter(..))`` instead.
```
bdice, line 2653:
```suggestion
        This function is deprecated. Users should use ``next(iter(...))`` instead.
```

signac/contrib/schema.py
```diff
@@ -332,6 +338,11 @@ def difference(self, other, ignore_values=False):
             )
         return ret
 
+    @deprecated(
+        deprecated_in="1.8",
+        removed_in="2.0",
```
bdice, line 343: Are we using `x.y` vs. `x.y.z` version numbers consistently in our deprecation notices?

signac/core/utility.py
```diff
@@ -78,6 +88,12 @@ def __repr__(self):
         return "Version({})".format(",".join((f"{k}={v}" for k, v in self.items())))
 
 
+@deprecated(
+    deprecated_in="1.8",
+    removed_in="2.0",
+    current_version=__version__,
+    details="Use packaging.version.Version instead.",
```
bdice, line 95:
```suggestion
    details="Use packaging.version.parse instead.",
```

signac/testing.py
bdice: Consider linking the issue in the TODO, perhaps.

bdice (changes requested):

signac/contrib/job.py
vyasr: Unfortunately no, I tried this before and it doesn't work. `deprecation.deprecated` only works on a callable.

signac/contrib/job.py
vyasr: Good idea.

signac/contrib/job.py
vyasr: Yeah we should do that, but you may have noticed that I made a separate note elsewhere in this PR that we should explore using `functools.cached_property` instead of manually maintaining these variables. I'd like to try that first since that would remove these variables altogether, if you and other reviewers agree.

signac/contrib/project.py
vyasr: It should be `:attr:`, I just copy-pasted and forgot to change it. Thanks.

signac/contrib/project.py
vyasr: Same question about `cached_property`.

signac/contrib/project.py
vyasr: Also the other method will be gone soon so there won't be anything to alias :)

signac/contrib/project.py
vyasr: Good question. Yes, it should become a property.

signac/contrib/schema.py
vyasr: With the exception of JobsCursor.next being deprecated in 0.9.6 I think so.

signac/core/utility.py
vyasr: Neither `version.Version` nor `version.parse` really has the same signature as this object. I suppose `version.parse` is the more useful alternative to indicate to users, but `version.Version` is probably the more accurate analog to this object.

signac/contrib/job.py
bdice: And require Python 3.8, when that feature was introduced?

vyasr: I think we have to start requiring Python 3.8 anyway as of #658, right? We just need to update the changelog/setup.py/etc to make that official. But yes, that would require more changes. And to be clear, I'm proposing to do that in a separate PR.

bdice: We support older numpy versions that are compatible with Python < 3.8. We just can’t test “latest numpy” with all the Python versions we currently support.

csadorf: Python 3.7 support is dropped according to NEP 29.

signac/contrib/job.py
csadorf, line 359:
```suggestion
    # Adding '# type: ignore' for mypy, because decorated properties aren't supported:
    # https://github.com/python/mypy/issues/1362
    @property  # type: ignore
```

signac/contrib/job.py
```diff
@@ -359,11 +355,36 @@ def _statepoint_filename(self):
...
+    @deprecated(
+        deprecated_in="1.8",
+        removed_in="2.0",
+        current_version=__version__,
+        details="Use Job.path instead.",
+    )
     def ws(self):
-        """Alias for :meth:`~Job.workspace`."""
-        return self.workspace()
+        """Alias for :meth:`~Job.path`."""
```
csadorf, line 367: Shouldn't this be `:attr:` ?

signac/contrib/project.py
```diff
@@ -790,7 +791,7 @@ def _contains_job_id(self, job_id):
 
         """
         # We can rely on the project workspace to be well-formed, so just use
-        # str.join with os.sep instead of os.path.join for speed.
+        # os.sep.join instead of os.path.join for speed.
```
csadorf, line 794:
```suggestion
        # os.sep.join instead of os.path.join for performance.
```

signac/contrib/schema.py
```diff
@@ -36,6 +42,8 @@ def _collect_by_type(values):
         A mapping of types to a set of input values of that type.
 
     """
+    # TODO: This function should be moved to project.py in version 2.0.
```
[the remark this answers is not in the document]
bdice: I find it valuable to include TODO so we can search for these comments and clean them up in version 2.0. If we remove the word, it will be hard to find comments like this one.

Athul0491 (approved):

signac/contrib/job.py
vyasr: Yeah I would think #658 is a reason to require Python 3.8 because that PR updated a dependency (Numpy) to a version that requires 3.8 directly as a consequence of NEP 29. [We document that we follow NEP 29](https://docs.signac.io/projects/core/en/latest/installation.html?highlight=NEP#installation), and I think every time this debate has come up we've agreed that we should follow it. By that we really should have dropped Python 3.6 almost 2 years ago, but our docs still say that they support it. I think starting with signac 2.0 we can make the move to strictly following NEP 29.

vyasr: That said, I'll go ahead and rename these variables for now.

signac/contrib/schema.py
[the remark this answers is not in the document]
vyasr: Yeah I personally still prefer to use TODOs. I would rather have unactionable items show up for a while rather than not have the TODO and therefore forget about the task when it does become actionable.

signac/contrib/job.py
bdice: > I would think https://github.com/glotzerlab/signac/pull/658 is a reason to require Python 3.8 because that PR updated a dependency (Numpy) to a version that requires 3.8 directly as a consequence of NEP 29.

If the next version of some dependency required Python 3.10, we wouldn’t need to require Python 3.10 because we can always use an older version of that dependency for testing with Python versions <3.10. The same applies here — I don’t think #658 is a valid justification to require Python 3.8. The justification is that we want to use Python 3.8 features (and that’s the highest we can go while being NEP 29 compliant).

bdice: Put differently, “strictly following NEP 29” just means we drop support _no faster_ than it allows. It’s an “infimum,” in the sense of a greatest lower bound.

In some cases, it has been easier to leave our CI alone and not increase our requirements immediately when NEP 29 drops support for something.

vyasr: I disagree that strictly following NEP 29 means treating it as a "greatest lower bound". [Having different packages decide to drop support for Python/NumPy versions at different times is explicitly discouraged by the NEP](https://numpy.org/neps/nep-0029-deprecation_policy.html#ad-hoc-version-support). A key goal is to create consistent expectations of what language/library features are available for use across the ecosystem. In practice, most packages (including NEP authors like Matplotlib, scikit-learn, and IPython) are treating the NEP's timeline as an infimum, so I'm OK with doing the same. However, that comes at the cost of creating more inconsistent expectations across the Python ecosystem, which is precisely what the NEP is designed to solve, and we lose clarity on what constitutes a "good enough reason" for a version bump.

If we are OK with that, then the question is how we make that decision. Do we wait until the NEP 29 window has passed and then bump whenever we see a new language/NumPy feature that we think is worthwhile? Do we standardize in any way, or do we make this decision on an ad hoc basis from version to version? Are there other constraints (e.g. the ability to use the Python versions available on supercomputers) that provide a more conservative guide than NEP 29's relatively aggressive policy?

bdice: These statements from NEP 29 are a little conflicting (emphasis added):
> When a project releases a new major or minor version, we recommend that they support _at least_ all minor versions of Python introduced and released […]

> Minimum Python and NumPy version support should be _adjusted upward on every major and minor release_, but never on a patch release.

I was aware of the first statement’s “at least” but was not aware of the second statement’s recommendation to (frequently!) bump Python/NumPy version minimum requirements. That explains part of why we had different interpretations of this piece of NEP 29. 😅

> Do we wait until the NEP 29 window has passed and then bump whenever we see a new language/NumPy feature that we think is worthwhile?

This is my recommendation. Bumping on every major/minor release would be a pain. Minimizing maintenance costs while supporting NEP 29 is my goal.

> Are there other constraints (e.g. the ability to use the Python versions available on supercomputers) that provide a more conservative guide than NEP 29's relatively aggressive policy?

This is not a high priority to me unless we have users willing to help identify and maintain such support. Most users have the ability to use a conda environment, container, or similar means to get beyond the provided system Python version. I don’t want to put that burden entirely on signac maintainers — HPC admins and users can shoulder some of it.

signac/__main__.py
```diff
@@ -279,6 +279,12 @@ def main_job(args):
     if args.create:
         job.init()
     if args.workspace:
+        warnings.warn(
+            "The `-w/--workspace` parameter is deprecated. Use -p/--path instead",
+            FutureWarning,
+        )
```
csadorf, line 285: Should we mention "deprecated as of ... and will be removed in ..." and then add the corresponding assertion?

signac/contrib/project.py
```diff
@@ -789,8 +787,8 @@ def _contains_job_id(self, job_id):
             True if the job id is initialized for this project.
 
         """
-        # We can rely on the project workspace to be well-formed, so just use
-        # str.join with os.sep instead of os.path.join for speed.
+        # We can rely on the project workspace to be well-formed, so use
+        # os.sep.join instead of os.path.join for performance.
         return os.path.exists(os.sep.join((self.workspace(), job_id)))
```
csadorf, line 792:
```suggestion
        return os.path.exists(os.sep.join((self.workspace, job_id)))
```

csadorf (changes requested):

signac/contrib/job.py
vyasr: Right, so basically I am fine with this policy but we just need to be careful about what we state. I don't think that this constitutes "following" NEP 29, but I think it's consistent with how most packages are treating the proposal. I think our main action item is to amend the statement [here](https://docs.signac.io/projects/core/en/latest/installation.html?highlight=nep%2029#installation), [here](https://docs.signac.io/projects/dashboard/en/latest/installation.html?highlight=nep%2029#installation) and [here](https://docs.signac.io/projects/flow/en/latest/installation.html?highlight=nep%2029#installation) to say something like "signac uses NEP 29 as a guideline for when to drop support for Python and NumPy versions and does not guarantee support beyond the versions recommended in that proposal". That is in line with [Scipy's statement](https://docs.scipy.org/doc/scipy-1.8.0/html-scipyorg/dev/toolchain.html#python-versions), which to my knowledge is the only such statement made by any of the major Python packages behind NEP 29.

bdice: I addressed this in https://github.com/glotzerlab/signac/pull/715/commits/1b18ffab3f3a5497b7911f4e82f089ca4242e989.

signac/__main__.py
vyasr: Good question. I don't think this change needs a particularly long deprecation, so are we fine with removing this in 2.0?

signac/contrib/project.py
vyasr: This probably needs a more global find-and-replace, I'll take care of that.

signac/__main__.py
csadorf: I would be.

bdice: Yes, I’m also in favor.

csadorf (approved):
Perfect! 👍

signac/contrib/project.py
```diff
@@ -789,9 +785,9 @@ def _contains_job_id(self, job_id):
             True if the job id is initialized for this project.
 
         """
-        # We can rely on the project workspace to be well-formed, so just use
-        # str.join with os.sep instead of os.path.join for speed.
-        return os.path.exists(os.sep.join((self.workspace(), job_id)))
+        # We can rely on the project workspace to be well-formed, so use
+        # os.sep.join instead of os.path.join for performance.
```
bdice, line 789: The changes in this comment aren't aligned with the comment in `Job.path`. Either is fine but I slightly prefer the other for clarifying "`str.join` with `os.sep`":

```suggestion
        # We can rely on the project workspace to be well-formed, so just
        # use str.join with os.sep instead of os.path.join for speed.
```

signac/contrib/project.py
```diff
@@ -789,9 +785,9 @@ def _contains_job_id(self, job_id):
...
+        return os.path.exists(os.sep.join((self.workspace, job_id)))
```
bdice, line 790: Do we mean to use `os.path.isdir` instead of `os.path.exists` here? That is, files shouldn't be allowed.

If we used `isdir`, symbolic links are followed (as we desire). It looks like this would be the same number of syscalls from the implementation I see:

https://github.com/python/cpython/blob/3751b6b030b4a3b88959b4f3c4ef2e58d325e497/Lib/genericpath.py#L39-L45

https://github.com/python/cpython/blob/3751b6b030b4a3b88959b4f3c4ef2e58d325e497/Lib/genericpath.py#L16-L22

We'd need to investigate the performance impact of this proposed change a little more closely but it might eliminate a possibility of false positives from files while being as efficient as possible.

bdice (approved):
A few comments but nothing that should block this PR. Some review comments may warrant future work / may be worth tracking in an issue.

signac/__main__.py
```diff
@@ -258,7 +258,7 @@ def main_project(args):
             print(json.dumps(doc))
         return
     if args.workspace:
-        print(project.workspace())
+        print(project.workspace)
```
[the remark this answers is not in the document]
vyasr: Yeah I don't know if `main_project` even needs to exist in the future. Maybe we just remove it altogether?

signac/contrib/job.py
```diff
@@ -335,22 +338,15 @@ def __repr__(self):
             self.__class__.__name__, repr(self._project), self.statepoint
         )
 
+    @deprecated(
+        deprecated_in="1.8",
+        removed_in="2.0",
+        current_version=__version__,
+        details="Use Job.path instead.",
+    )
     def workspace(self):
```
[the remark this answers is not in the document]
vyasr: No that's a good call, I'll fix that in this PR.

signac/contrib/project.py
vyasr: Technically I think you are correct, although I don't think it matters in practice. If a file has been created with that hash we either have a bug in signac (which we probably don't know how to recover from without manual intervention) or a user has been manually monkeying with the workspace, which IMO is not something we need to protect against. We can certainly test out the performance implications, although `isdir` is slightly more complicated since it has to check whether the target is a file or a directory (the `stat.S_ISDIR(st.st_mode)` check) so it will definitely be slower (but hopefully not by much, maybe on the order of 50-100 ns).

signac/contrib/project.py
```diff
@@ -215,8 +228,11 @@ def __setitem__(self, key, value):
 
 def _invalidate_config_cache(project):
     """Invalidate cached properties derived from a project config."""
+    # TODO: Consider using functools.cached_property for these instead. That simplify our code,
+    # although at the expense of creating multiple locks (each cached_property has its own lock
+    # AFAICT from looking at the implementation).
```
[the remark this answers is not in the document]
vyasr: > for non-mutable properties one can just use @property and @cache in combination

The docs discussing that combination aren't referring to immutable properties, they're referring to classes where `__slots__` is defined so that instances don't have a `__dict__`. That being said, you are partially correct that this combination would give an immutable property, but there are some dealbreaking caveats. Note that `functools.cache` wasn't introduced until 3.9 so we would have to use `functools.lru_cache` instead, but that's not an issue. The problem will be that the cache will be extremely cumbersome to access, and it will be class-wide and cannot be cleared on a per-instance basis.

> The gain in simplicity might not outweigh the potential performance cost of creating multiple locks.

The actual result will be far fewer locks because it will only create one lock per class, not one per instance. In a GIL-free world that would lead to excessive serialization, but we don't live in that world so only having one instance set the property at a time shouldn't be a problem. And for subsequent accesses the same implementation details that unfortunately make `cached_property` writeable (and that require the existence of `__dict__`) will also make it much faster than a normal `property` for typical usage. Here's a benchmark of the different options including @csadorf's proposal above
```
In [40]: class A:
    ...:     def __init__(self):
    ...:         self._z = None
    ...:
    ...:     @cached_property
    ...:     def w(self):
    ...:         return 0
    ...:
    ...:     @property
    ...:     @lru_cache(None)
    ...:     def x(self):
    ...:         return 0
    ...:
    ...:     @property
    ...:     def y(self):
    ...:         if not hasattr(self, '_y'):
    ...:             self._y = 0
    ...:         return self._y
    ...:
    ...:     @property
    ...:     def z(self):
    ...:         if self._z is None:
    ...:             self._z = 0
    ...:         return self._z
    ...:

In [41]: a = A()

In [42]: %timeit a.w
31.1 ns ± 0.268 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [43]: %timeit a.x
67 ns ± 0.648 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [44]: %timeit a.y
109 ns ± 0.976 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [45]: %timeit a.z
97.9 ns ± 0.86 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

As you can see, `cached_property` is the clear winner, followed by the combination of `property` with `lru_cache`. You'll also see constructors get faster because the initialization only happens once for the class (also true for `property`, but our manual caching introduces extra work per instance).

> unless it improves thread safety.
I think the main benefit to thread safety that we get is that `cached_property`s will automatically get locked without us having to think about it. Conversely, there are some methods that need to be mutually locked and this class won't help us there, so we'll still end up needing to maintain per-instance locks for those cases. That is more what I was referring to in my comment about "the expense of multiple locks". I wasn't so much thinking of performance cost as the fact that multiple locks may not be workable for certain mutually interacting properties/methods.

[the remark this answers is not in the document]
csadorf: > The docs discussing that combination aren't referring to immutable properties, they're referring to classes where `__slots__` is defined so that instances don't have a `__dict__`. That being said, you are partially correct that this combination would give an immutable property, but there are some dealbreaking caveats. Note that `functools.cache` wasn't introduced until 3.9 so we would have to use `functools.lru_cache` instead, but that's not an issue. The problem will be that the cache will be extremely cumbersome to access, and it will be class-wide and cannot be cleared on a per-instance basis.

Thanks for the clarification, I indeed did not read the preceding paragraph careful enough, however – like you said – this would still be the recommended approach as far as I understand; see also the first paragraph:

> Transform a method of a class into a property whose value is computed once and then cached as a normal attribute for the life of the instance. Similar to [property()](https://docs.python.org/3/library/functions.html#property), with the addition of caching.

 Looks like that approach would already yield a slight performance boost.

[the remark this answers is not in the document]
vyasr: I'm a little confused about what "this" and "that approach" are referring to in your last comment :joy: 

To make sure that we're on the same page: Using `cached_property` is fine and the recommended approach, and as I showed in my last comment it does lead to a performance boost. Using `property` with `cache` also gives a performance boost. However, I think the fact that it creates a per-class rather than per-instance cache makes it unworkable for our purposes.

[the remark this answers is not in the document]
bdice: > Using `cached_property` is fine and the recommended approach

...except that the resulting attribute from `cached_property` is settable, which we don't want. I mentioned this in my first comment and I'm confused about why we're evaluating `cached_property` as a viable alternative to the status quo.

```python
>>> from functools import cached_property
>>> class A:
...   @cached_property
...   def id(self):
...     return "abcdef"
...
>>> a = A()
>>> a.id
'abcdef'
>>> a.id = "this is settable"
>>> a.id
'this is settable'
```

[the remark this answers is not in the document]
csadorf: I'm a bit confused as well. I don't think that `cached_property` should be discussed as a viable alternative. And why are you saying that `@propert+@cache` is caching class-wide? Here is a simple test which demonstrates that it is per-instance:
```python
from functools import cache, lru_cache


class A:
    @property
    @cache
    def foo(self):
        print("calc foo.", id(self))
        return "bar"
```
When I evaluate this for
```python


a = A()
b = A()

print(a.foo)
print(a.foo)
print(b.foo)
print(b.foo)
```
I get:
```console
% python test.py
calc foo. 140312263293920
bar
bar
calc foo. 140312263292816
bar
bar
```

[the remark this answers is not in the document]
vyasr: Let me clarify: _I no longer think this comment is actionable in any way, and I will remove it_.

> Using cached_property is fine and the recommended approach

That comment was about what the docs recommend for caching properties, not what I was recommending for signac. I know `cached_property` is settable, but I wasn't thinking of that as a serious problem until @bdice raised it in his first comment. Not every library views it this way (scikit-learn is a prominent example of using raw attributes, for instance, and PyTorch does so in some cases as well), so it doesn't have to be a showstopper. I'm fine with deciding against that approach in signac, though.

> And why are you saying that @propert+@cache is caching class-wide?

The _values_ are per-instance, but the _cache_ is per class. A property is functionally like a `classmethod` that is called on instances. So in your example `a.foo` is really closer to `A.foo(a)`. Therefore, the cache is stored in the class `A` and keyed on the instance of `a` that it is called on. You cannot clear the cache for just a single instance, you have to clear the cache for all instances (and doing so is extremely cumbersome since you have to bypass normal attribute access patterns to touch the cache), and that's a non-starter when for example you have many jobs and you modify one of their statepoints: you don't want to lose cached information on _all_ job instances just because one statepoint was modified.

signac/contrib/project.py
bdice: https://github.com/bdice/python_microbenchmarks/blob/main/microbenchmarks/os_path_isdir.py

This microbenchmark suggests that `os.path.exists` takes about 749 ns, while `os.path.isdir` takes about 815 ns. An interesting result, but nothing actionable here imo. I'll resolve this.

signac/contrib/project.py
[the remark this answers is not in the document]
csadorf: > Not every library views it this way (scikit-learn is a prominent example of using raw attributes, for instance, and PyTorch does so in some cases as well), so it doesn't have to be a showstopper. I'm fine with deciding against that approach in signac, though.

Might not be a showstopper, but since signac has been using immutable properties until now, would be unexpected to change the API in this way, no?

> > And why are you saying that @Propert+@cache is caching class-wide?
> 
> The _values_ are per-instance, but the _cache_ is per class. A property is functionally like a `classmethod` that is called on instances. So in your example `a.foo` is really closer to `A.foo(a)`. Therefore, the cache is stored in the class `A` and keyed on the instance of `a` that it is called on. You cannot clear the cache for just a single instance, you have to clear the cache for all instances (and doing so is extremely cumbersome since you have to bypass normal attribute access patterns to touch the cache), and that's a non-starter when for example you have many jobs and you modify one of their statepoints: you don't want to lose cached information on _all_ job instances just because one statepoint was modified.

I get what you mean now. That said, I do wonder though whether the performance penalty might be acceptable, since modifying the state point shouldn't be a routine operation, should it?

signac/__main__.py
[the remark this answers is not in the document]
vyasr: I deprecated the command.

signac/contrib/project.py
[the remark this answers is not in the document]
vyasr: > Might not be a showstopper, but since signac has been using immutable properties until now, would be unexpected to change the API in this way, no?

Honestly I'm not sure that's true. You can't effectively query `help` on properties or attributes since you'll just get the docstring for the returned value, so unless someone goes to our online documentation and actively searches for the specific docs they won't even notice the change. It will largely only be relevant for new users, who won't have seen the original code. I admit that it is potentially a way for users to do something nonsensical, but again I think plenty of libraries get away with this and the potential performance benefits are compelling.

> I get what you mean now. That said, I do wonder though whether the performance penalty might be acceptable, since modifying the state point shouldn't be a routine operation, should it?

That is a good point. The tradeoff is probably worth it. Part of me wonders if any of the caching questions here are even relevant for the benchmarks @bdice has looked at most frequently since IIRC those almost always involve a single pass (or maybe a very small number of passes) through the data space, in which case the caching may not be a huge performance improvement either. We need to evaluate concrete benchmarks to make a call here.

signac/contrib/project.py
vyasr: Sure.

vyasr:
We'll discuss the cached property question at the next dev meeting then I think this PR is good to go.

signac/contrib/project.py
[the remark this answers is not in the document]
vyasr: In the interest of moving forward, I'm going to resolve this and merge the PR. We can benchmark alternatives at a later point.

PR #686, opened 2022-02-22
Allow PR template to catch incomplete tasks

cbkerr:
# Final checklist:

## Description of bug fix

This makes all checkboxes required by adding an "or" to the ones that were previously under a separate heading of "if necessary".

[test of relative link](../tree/master/doc)

I propose creating a label for possibly breaking changes.

This PR uses the new template **before reviews**

## Motivation and Context

glotzerlab/signac-flow/pull/508 was merged with unchecked boxes. It will be easier to catch this if we don't use checkboxes as radio buttons or otherwise optional items.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -19,5 +19,5 @@
 - [ ] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
 - [ ] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
 - [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.
-- [ ] If this PR changes user-facing features or breaks things, I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) at: glotzerlab/signac-docs/pull#**XXX**
+- [ ] If this PR changes user-facing features or breaks existing APIs, I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and included a link here.
 - [ ] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added all related issue and pull request numbers for future reference (if applicable).
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,10 +1,8 @@
 <!-- Provide a general summary of your changes in the Title above -->
 
-
 ## Description
 <!-- Describe your changes in detail. -->
 <!-- If possible, please indicate if this change breaks or may break existing functionality, so we can mark it as 'breaking'. -->
-
 
 ## Motivation and Context
 <!-- Why is this change required? What problem does it solve? -->
@@ -12,7 +10,7 @@
 
 
 ## Checklist:
-<!-- Please complete the checklist before the PR can be merged. Reviews can begin while the checklist is in progress. -->
+<!-- This checklist must be complete before merging the pull request. -->
 <!-- If you are unsure about any of these items, do not hesitate to ask! -->
 - [ ] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
 - [ ] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,7 +3,7 @@
 
 ## Description
 <!-- Describe your changes in detail. -->
-<!-- Please indicate if this change breaks or may break existing functionality, so we can mark it as 'breaking'. -->
+<!-- If possible, please indicate if this change breaks or may break existing functionality, so we can mark it as 'breaking'. -->
 
 
 ## Motivation and Context
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,7 +1,7 @@
 <!-- Provide a general summary of your changes in the Title above -->
 
 
-## Description of [new feature/bug fix/ documentation update]
+## Description
 <!-- Describe your changes in detail. -->
 <!-- Please indicate if this change breaks or may break existing functionality, so we can mark it as 'breaking'. -->
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -18,7 +18,7 @@
 - [ ] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
 - [ ] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
 - [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.
-- [ ] I have determined whether the [framework documentation](https://docs.signac.io/) requires an update (see guidance below). If so, I have created a separate pull request to update it on [signac-docs](https://github.com/glotzerlab/signac-docs) at this link: glotzerlab/signac-docs#**XXX**
+- [ ] I have determined whether the [framework documentation](https://docs.signac.io/) requires an update (see guidance below). If so, I have created a separate pull request to update it on [signac-docs](https://github.com/glotzerlab/signac-docs) at this link: glotzerlab/signac-docs/pull#**XXX**
 <!-- The framework documentation must be updated if:
         This PR will change user-facing features (labeled "enhancement").
         This PR will break things (labeled "breaking").
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -16,5 +16,5 @@
 - [ ] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
 - [ ] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
 - [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.
-- [ ] The package documentation and [framework documentation](https://docs.signac.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.
+- [ ] The [package documentation](../doc) and [framework documentation](https://docs.signac.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.
 - [ ] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added any related issue and pull request numbers for future reference.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -12,7 +12,7 @@
 
 
 ## Checklist:
-<!-- Please select all items that apply either now or after creating the pull request. -->
+<!-- Please complete the checklist before the PR can be merged. Reviews can begin while the checklist is in progress. -->
 <!-- If you are unsure about any of these items, do not hesitate to ask! -->
 - [ ] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
 - [ ] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -20,4 +20,4 @@
 - [ ] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
 - [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.
 - [ ] If this PR changes user-facing features or breaks existing APIs, I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and included a link here.
-- [ ] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added all related issue and pull request numbers for future reference (if applicable).
+- [ ] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added any related issue and pull request numbers for future reference.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,8 +2,8 @@
 
 
 ## Description of [new feature/bug fix/ documentation update]
-<!-- If this change breaks (or has the potential to break) existing functionality, add the `breaking` label. -->
-<!-- Describe your changes in detail -->
+<!-- Describe your changes in detail. -->
+<!-- Please indicate if this change breaks or may break existing functionality, so we can mark it as 'breaking'. -->
 
 ## Motivation and Context
 <!-- Why is this change required? What problem does it solve? -->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -12,9 +12,9 @@
 ## Checklist:
 <!-- This checklist must be complete before merging the pull request. -->
 <!-- If you are unsure about any of these items, do not hesitate to ask! -->
-- [ ] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
-- [ ] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
+- [ ] I am familiar with the [Contributing Guidelines](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
+- [ ] I agree with the terms of the [Contributor Agreement](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
 - [ ] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
-- [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.
+- [ ] The changes introduced by this pull request are covered by existing or newly introduced **tests**.
 - [ ] The [package documentation](../tree/master/doc) and [framework documentation](https://docs.signac.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.
 - [ ] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added any related issue and pull request numbers for future reference.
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,20 +1,15 @@
 <!-- Provide a general summary of your changes in the Title above -->
 
-## Description
+
+## Description of [new feature/bug fix/ documentation update]
+If this change breaks (or has the potential to break) existing functionality, add the `breaking` label.
+
```
cbkerr, line 6:
```suggestion
<!-- If this change breaks (or has the potential to break) existing functionality, add the `breaking` label. -->
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,20 +1,14 @@
 <!-- Provide a general summary of your changes in the Title above -->
 
-## Description
+
+## Description of [new feature/bug fix/ documentation update]
+<!-- If this change breaks (or has the potential to break) existing functionality, add the `breaking` label. -->
```
bdice, line 5: Not all PR authors have privileges to add a label.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -24,10 +18,7 @@
 - [ ] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
 - [ ] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
 - [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.
-
-If necessary:
-- [ ] I have updated the API documentation as part of the package doc-strings.
-- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
+- [ ] I have updated the API documentation as part of the package doc-strings, or this is not necessary.
```
bdice, line 21: As a reviewer I think the “or not necessary” language is potentially unclear. “Not necessary” is sometimes a subjective decision, whereas the existing criteria of updating docs/tests is more objective (did or did not change). Is there a way to make all the criteria more objective?

.github/PULL_REQUEST_TEMPLATE.md
csadorf: Should we say:

> Please make sure to mention whether this change breaks (or has the potential to break) existing functionality.

Whoever triages this, should add the `breaking` label then.

.github/PULL_REQUEST_TEMPLATE.md
csadorf: I think we can completely drop this item, because the presence and correct formatting of doc-strings should ideally be enforced by a linter. There is simply no scenario where checking this box is either not necessary or provides significant value IMO. The same is not true for the next item (see comment below).

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -24,10 +18,7 @@
...
+- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here, or this is not necessary.
```
csadorf, line 22:
```suggestion
- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here. / No change to the framework documentation needed (see below for guidance).
```

Here is a first _crude_ attempt at this guidance:

> #### Guidance on [framework documentation](https://docs.signac.io) updates:
>
> - Any change that affects user-facing features (labeled `enhancement`) must be documented.
> - Breaking changes (labeled `breaking`) may require a revision.
> - Changes that are only affecting purely internal components do not require an update.

csadorf (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
mikemhenry: Agreed

mikemhenry (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
cbkerr: I've revised these guidelines and made the checklist item start by determining if an update is needed.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -24,10 +18,13 @@
 - [ ] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
 - [ ] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
 - [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.
-
-If necessary:
-- [ ] I have updated the API documentation as part of the package doc-strings.
-- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
+- [ ] I have determined whether the [framework documentation](https://docs.signac.io/) requires an update (see guidance below). If so, I have created a separate pull request to update it on [signac-docs](https://github.com/glotzerlab/signac-docs) at this link: glotzerlab/signac-docs/pull#**XXX**
```
bdice, line 21: When this text is rendered, the “guidance below” will be invisible in a comment. Readers will not know what this guidance says, only the author of the PR. I think that is potentially confusing. However, we also want to minimize the amount of boilerplate text in each PR. I might suggest a link to the CONTRIBUTING document, and include the description there.

We might also want to unify some of our CONTRIBUTING document with the developer guide in our docs… a single source of info might be better?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -24,10 +18,13 @@
...
+<!-- The framework documentation must be updated if:
+        This PR will change user-facing features (labeled "enhancement").
+        This PR will break things (labeled "breaking").
+-->
+<!-- The framework documentation can be left if:
```
bdice, line 26: This is probably redundant, the “else” may not need to be stated.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -24,10 +18,13 @@
...
+        This PR affects only internal components.
+-->
 - [ ] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added all related issue and pull request numbers for future reference (if applicable). See example below.
```
bdice, line 29: Similar issue here with “example below” which is hidden. 🙃

bdice (commented):
Comments attached. Overall I would be happy to see this document streamlined — I think less content is better than more in the template. Large blocks of instructions are often ignored, in my experience with creating PRs on other projects and seeing PRs from new contributors on projects I maintain. Maybe we can outsource more content to the CONTRIBUTING document.

csadorf (commented):
I agree with @bdice 's comments. Please address those and then re-request review from me. Thank you.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -24,10 +18,13 @@
 - [ ] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
 - [ ] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
 - [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.
-
-If necessary:
-- [ ] I have updated the API documentation as part of the package doc-strings.
-- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
+- [ ] I have determined whether the [framework documentation](https://docs.signac.io/) requires an update (see guidance below). If so, I have created a separate pull request to update it on [signac-docs](https://github.com/glotzerlab/signac-docs) at this link: glotzerlab/signac-docs/pull#**XXX**
+<!-- The framework documentation must be updated if:
+        This PR will change user-facing features (labeled "enhancement").
```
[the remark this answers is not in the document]
cbkerr: That wasn't obvious to me. Could we rename the label "enhancement" to "internal"?

.github/PULL_REQUEST_TEMPLATE.md
cbkerr: I'll just get rid of it because people can refer to the changelog file itself for examples

.github/PULL_REQUEST_TEMPLATE.md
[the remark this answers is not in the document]
cbkerr: I was thinking that any PR not marked as a bug fix or docs update would be an enhancement. (Even bug fixes are a kind of enhancement)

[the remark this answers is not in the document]
bdice: > I was thinking that any PR not marked as a bug fix or docs update would be an enhancement.

That's my stance as well. I just wanted to change the language "will change user-facing features" because it's not broad enough.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -24,10 +19,5 @@
...
 - [ ] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
 - [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.
-
-If necessary:
-- [ ] I have updated the API documentation as part of the package doc-strings.
-- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
-- [ ] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added all related issue and pull request numbers for future reference (if applicable). See example below.
-
-<!-- Example for a changelog entry: `Fix issue with launching rockets to the moon (#101, #212).` -->
+- [ ] If this PR changes user-facing features or breaks things, I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) at: glotzerlab/signac-docs/pull#**XXX**
```
bdice, line 22:
```suggestion
- [ ] If this PR changes user-facing features or breaks existing APIs, I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and included a link here.
```

bdice (commented):
One small suggestion. Otherwise LGTM.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,33 +1,23 @@
 <!-- Provide a general summary of your changes in the Title above -->
 
+
 ## Description
-<!-- Describe your changes in detail -->
+<!-- Describe your changes in detail. -->
+<!-- Please indicate if this change breaks or may break existing functionality, so we can mark it as 'breaking'. -->
```
csadorf, line 6: I think, especially for new or inexperienced contributors, this is very hard to assess.

```suggestion
<!-- If possible, please indicate if this change breaks or may break existing functionality, so we can mark it as 'breaking'. -->
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,33 +1,23 @@
...
 - [ ] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
 - [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.
-
-If necessary:
-- [ ] I have updated the API documentation as part of the package doc-strings.
-- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
-- [ ] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added all related issue and pull request numbers for future reference (if applicable). See example below.
-
-<!-- Example for a changelog entry: `Fix issue with launching rockets to the moon (#101, #212).` -->
+- [ ] If this PR changes user-facing features or breaks existing APIs, I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and included a link here.
```
csadorf, line 22: Still not sure whether it is clear that this should be checked in case this does _not_ apply. Should the reviewers do that?

csadorf (commented):
I have one minor suggestion, but overall this is fine.

I still feel like the list is overall maybe a bit too long. Lack of testing is automatically controlled by codecov for example. Code style is automatically checked by linters and autoformatters. How would you feel about dropping all items that should be automatically enforced? @bdice Also interested in your opinion here.

cbkerr:
> Code style is automatically checked by linters and autoformatters. 

I like getting rid of the code style check, since that is part of the code guidelines

> Lack of testing is automatically controlled by codecov for example

I think having the box for tests would help remind me to include tests while I'm starting the PR as opposed to when reviews are nearly done.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,33 +1,22 @@
...
-- [ ] Documentation update
-- [ ] Bug fix
-- [ ] New feature
-- [ ] Breaking change<sup>1</sup>
-
-<sup>1</sup>The change breaks (or has the potential to break) existing functionality.
 
 ## Checklist:
-<!-- Please select all items that apply either now or after creating the pull request. -->
+<!-- Please complete the checklist before the PR can be merged. Reviews can begin while the checklist is in progress. -->
```
bdice, line 15: The part about reviews seems unnecessary. Reviews can begin at any point, even on draft PRs. Also made the language more direct/clear.
```suggestion
<!-- This checklist must be complete before merging the pull request. -->
```

.github/PULL_REQUEST_TEMPLATE.md
bdice: API breaks don't always require a pull request to signac-docs. I'd even bet that _most_ API breaks do not require changes in the framework documentation.

I agree with Simon that it's not clear that the intended action for most pull requests is to check this box without needing to make a separate change in the documentation. I am having trouble coming up with a wording that is concise and indicates that this box should be checked by default.

Given the difficulty in describing this concisely and the relatively small number of PRs to which it pertains, should we remove the signac-docs item from the checklist? Changes in signac-docs happen pretty infrequently, and perhaps it's better to remove this checkbox and leave it to committers to request that as a part of review. This could be described in the Contributing Guidelines instead.

bdice:
> I still feel like the list is overall maybe a bit too long. Lack of testing is automatically controlled by codecov for example.

Code coverage isn't really a metric of whether code is _tested_ adequately. I wouldn't rely on codecov, since it is very prone to noise and doesn't really indicate whether the code has bugs -- only that it executes. Removing a line of documentation decreases net coverage, while splitting a single line of code into multiple lines sometimes increases coverage...

> Code style is automatically checked by linters and autoformatters. How would you feel about dropping all items that should be automatically enforced?

Code style can be removed from the checklist, that is sufficiently enforced by our CI at this point.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,33 +1,22 @@
 <!-- Provide a general summary of your changes in the Title above -->
 
+
 ## Description
-<!-- Describe your changes in detail -->
+<!-- Describe your changes in detail. -->
+<!-- If possible, please indicate if this change breaks or may break existing functionality, so we can mark it as 'breaking'. -->
```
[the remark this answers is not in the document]
mikemhenry: :+1:

.github/PULL_REQUEST_TEMPLATE.md
cbkerr: I was debating whether to keep the "please" :)

.github/PULL_REQUEST_TEMPLATE.md
cbkerr: I agree it's still not clear.
I think we should keep SOME reminder about the docs because failing to update docs (and still passing review) is what precipitated this PR. Even though they are small in number, they are more important to get right.

Maybe create some visible text or subsection called "Do you think the documentation needs to be updated?"

bdice: We’re probably just as guilty of forgetting the package docs, let alone the framework docs. Maybe this:

```suggestion
- [ ] The package documentation and [framework documentation](https://docs.signac.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.
```

cbkerr: I like this a lot!

cbkerr:
I think this is ready for final review. I updated the PR description with the new checklist.
Do I need to update the changelog?

bdice:
> Do I need to update the changelog?

Nope, this kind of thing isn't put in the changelog (that's also a funny question, in a "meta" way 😉).

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,33 +1,20 @@
...
-- [ ] My code follows the [code style guideline](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md#code-style) of this project.
-- [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.
-
-If necessary:
-- [ ] I have updated the API documentation as part of the package doc-strings.
-- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
-- [ ] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added all related issue and pull request numbers for future reference (if applicable). See example below.
-
-<!-- Example for a changelog entry: `Fix issue with launching rockets to the moon (#101, #212).` -->
+- [ ] The changes introduced by this pull request are covered by existing or newly introduced **tests**.
```
bdice, line 18: It's weird to have _one_ bold word in the entire document. I agreed with the change to remove bold font from the other linked words. Let's not bold this.
```suggestion
- [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,33 +1,20 @@
...
+- [ ] The [package documentation](../tree/master/doc) and [framework documentation](https://docs.signac.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.
```
bdice, line 19: I suspect this relative link will not work. It doesn't work in the document preview, at least: https://github.com/glotzerlab/signac/blob/68c01377a0a42b8c699220bcf0334f7f2c6b3a07/.github/PULL_REQUEST_TEMPLATE.md

Can we just use an absolute link like the other links in this document?
```suggestion
- [ ] The [package documentation](https://github.com/glotzerlab/signac/tree/master/doc) and [framework documentation](https://docs.signac.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.
```

bdice: If it's not too much information, we could state both the hosted URL and source file locations for both package docs and framework docs:
```suggestion
- [ ] The [package documentation](https://docs.signac.io/projects/core/) in [this repository](https://github.com/glotzerlab/signac/tree/master/doc) and [framework documentation](https://docs.signac.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.
```

bdice (approved):
I approve. This is (after many iterations and great patience by all involved) an improvement over the current PR template! Thanks @cbkerr.

mikemhenry (approved):

bdice:
It looks like we have a [test failure related to pytest-xdist](https://app.circleci.com/pipelines/github/glotzerlab/signac/2814/workflows/306a6c62-fa6d-4ee7-8a7e-73edb70579d7/jobs/19461). It looks like worker 0 collected some tests for MongoDB that were not collected by worker 1, and one of those tests failed. 😕 I guess we have a few paths forward:
- Disable pytest-xdist
- Only run signac tests in parallel and run synced collections tests separately
- Ignore this failure and hope it's infrequent

vyasr:
The two workers collecting different threads is a symptom, not the problem. If you look at the failure, it's in the code that actually decides whether or not to run MongoDB tests by testing whether a MongoDB server can be accessed. The problem is almost certainly coming from a race condition between the two processes in performing the test write to the same MongoDB server. The brute force solution is to replace the except clause with a blanket `except Exception`. Worst case we just don't run MongoDB tests. If we want to be a little safer to avoid failing in serial contexts, we could [detect whether or not we're running with pytest-xdist](https://pypi.org/project/pytest-xdist/#identifying-the-worker-process-during-a-test) before deciding whether to allow arbitrary exceptions through.

PR #704, opened 2022-03-02
Run asv benchmarks instead of `benchmark.py` in CI.

bdice:
## Description
This PR runs benchmarks with asv instead of `benchmark.py`. This uses `asv continuous`, which is intended for CI testing: https://asv.readthedocs.io/en/stable/commands.html#asv-continuous

Currently, I am not very satisfied with the results I get here. I have experimented locally and in CI, and I have not found a good way to keep the benchmark runtime low while ensuring a small runtime variance (through repetitions). asv's approach spawns processes for each test and creates virtual environments for each commit, which may help with reproducibility but costs a lot of time. Previously, the CI benchmarks took less than 15 seconds for jobs like [this one](https://app.circleci.com/pipelines/github/glotzerlab/signac/2760/workflows/39215278-126c-4f55-89be-04503d706af2/jobs/19027). Meanwhile, the current [asv CI benchmarks](https://app.circleci.com/pipelines/github/glotzerlab/signac/2776/workflows/4eb5e0d7-980a-4324-9661-eaa88fcd0432/jobs/19153) take longer and have an unacceptably variance in runtime, leading to some [random failures](https://app.circleci.com/pipelines/github/glotzerlab/signac/2777/workflows/0b38d126-1104-41d0-bd9e-c8e5999adc4d/jobs/19161). Increasing the numbers of repetitions/rounds/samples/etc. can help with variance but will also increase runtime. I don't think we can get asv benchmarks to run as quickly as `benchmark.py` (or within, say, 1 minute) while keeping the limits on variance that we need for accuracy.

I'm not sure what to do with this and am near the limit of my interest in retaining CI benchmarks, so I might recommend removing CI benchmarks entirely (the existing CI benchmarks also have somewhat high variance).

## Motivation and Context
`benchmark.py` relies on `signac.Collection` which we plan to remove in #683. Using asv instead of `benchmark.py` is one way to keep CI benchmarks while removing that script.

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,8 +92,7 @@ jobs:
               . venv/bin/activate
               ${PYTHON} -m pip install --progress-bar off -r requirements/requirements-benchmark.txt
               ${PYTHON} -m pip freeze
-              ${PYTHON} -m asv machine --machine circleci
-              ${PYTHON} -m asv continuous --machine circleci master "${CIRCLE_SHA1}"
+              ${PYTHON} -m asv continuous master "${CIRCLE_SHA1}"
               ${PYTHON} benchmark.py run -N 100 1000  # this revision
               ${PYTHON} benchmark.py report
               git reset --hard origin/master
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,6 +92,8 @@ jobs:
               . venv/bin/activate
               ${PYTHON} -m pip install --progress-bar off -r requirements/requirements-benchmark.txt
               ${PYTHON} -m pip freeze
+              ${PYTHON} -m asv check
+              ${PYTHON} -m asv machine --yes
               ${PYTHON} -m asv continuous master "${CIRCLE_SHA1}"
               ${PYTHON} benchmark.py run -N 100 1000  # this revision
               ${PYTHON} benchmark.py report
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,6 +92,8 @@ jobs:
               . venv/bin/activate
               ${PYTHON} -m pip install --progress-bar off -r requirements/requirements-benchmark.txt
               ${PYTHON} -m pip freeze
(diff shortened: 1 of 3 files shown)


vyasr:
I have never felt great about running benchmarks in CI. Unless we have dedicated machines the variance is invariably going to be a problem. We ran into this plenty with freud, but I guess signac benchmarks are small enough that we've gotten away with it up to now. In lieu of acquiring dedicated hardware, IMO the best solution would be to store benchmark data to some remote repo after each commit to master and then use a weekly (or some periodic) CI run that analyzes the last week of benchmarks and looks for any spikes that are outside the variance of recent runs. That approach would be much less susceptible to pure noise. It is a lot of work, though, and I'm fine just creating an issue to track suggestions on how to address this problem going forward and disabling CI benchmarks for now.

bdice:
@vyasr Do you have an opinion on whether we need to keep the `benchmark.py` file? `benchmark.py` will quit working entirely once `signac.Collection` is removed in #683 unless we refactor it. I'd support disabling CI benchmarks but we'd still be leaving the `benchmark.py` file in a "broken" state for now.

vyasr:
I assumed that we would be getting rid of benchmark.py when we removed Collection. IIRC benchmark.py is just formalizing the benchmarks Simon, Paul and I wrote for the original paper. The asv benchmarks have basically just copied that logic, except putting it into a framework that can be run with asv, right? i.e. there's no extra benchmarks in benchmark.py? So it really boils down to the question of why does asv have so much overhead relative to benchmark.py, and whether that's a blocker for us getting rid of the simple benchmark script.

bdice:
Yes. I think I need to do a bit more investigation into whether asv’s results exactly match those of the current script. I am not 100% sure about the caching behavior. Then we’d be safe to delete the benchmark script and disable CI benchmarks.

PR #705, opened 2022-03-03
Run tests in parallel with pytest-xdist.

bdice:
## Description
This PR adds the plugin [pytest-xdist](https://pypi.org/project/pytest-xdist/) to run tests in parallel. Each CircleCI machine has two cores. This cuts off around 45 seconds from each CI job, which adds up to a decent amount of time across all jobs. The output is a little more verbose.

Before:
image

After (ignore pypy, I pushed another commit before it finished):
image

## Motivation and Context
Might as well use all of the CI resources we are given. Saves a bit of time and probably reduces the energy needed by the server since it can sleep sooner.

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -15,7 +15,7 @@ Added
 
  - Official support for Python 3.10 (#631).
  - Benchmarks can be run using the ``asv`` (airspeed velocity) tool (#629).
- - Tests run in parallel with ``pytest-xdist`` (#705).
+ - Continuous integration tests run in parallel with ``pytest-xdist`` (#705).
 
 Changed
 +++++++
--- a/changelog.txt
+++ b/changelog.txt
@@ -15,6 +15,7 @@ Added
 
  - Official support for Python 3.10 (#631).
  - Benchmarks can be run using the ``asv`` (airspeed velocity) tool (#629).
+ - Tests run in parallel with ``pytest-xdist`` (#705).
 
 Changed
 +++++++
--- a/.circleci/ci-oldest-reqs.txt
+++ b/.circleci/ci-oldest-reqs.txt
@@ -8,6 +8,7 @@ packaging==15.0
 pandas==1.0.0
 pymongo==3.0.0
 pytest-cov==2.10.1
+pytest-xdist==2.5.0
 pytest==6.2.1
 tables==3.4.2
 tqdm==4.10.0
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,7 +75,7 @@ jobs:
           name: Run tests
           command: |
             . venv/bin/activate
-            ${PYTHON} -m pytest --cov signac --cov-config=setup.cfg --cov-report=xml tests/ -v
+            ${PYTHON} -m pytest -n auto --cov=signac --cov-config=setup.cfg --cov-report=xml tests/ -v
             codecov
 
       - store_artifacts:
@@ -171,7 +171,7 @@ jobs:
       - run:
           name: Run tests
           command: |
-            ${PYTHON} -m pytest --cov=signac --cov-report=xml tests/ -v
+            ${PYTHON} -m pytest -n auto --cov=signac --cov-report=xml tests/ -v
             codecov
 
   macos-python-3:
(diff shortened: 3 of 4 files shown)


mikemhenry:
Maybe do `-n auto` so that if we ever have more cores/different CI we don't need to change the number?

bdice:
> Maybe do `-n auto` so that if we ever have more cores/different CI we don't need to change the number?

Unfortunately the CI machine claims to have 36 cores but only schedules 2. Auto ends up oversubscribing really badly.

mikemhenry:
> > Maybe do `-n auto` so that if we ever have more cores/different CI we don't need to change the number?
> 
> Unfortunately the CI machine claims to have 36 cores but only schedules 2. Auto ends up oversubscribing really badly.

Yikes! Okay that makes sense!

joaander:
> > > Maybe do `-n auto` so that if we ever have more cores/different CI we don't need to change the number?
> > 
> > 
> > Unfortunately the CI machine claims to have 36 cores but only schedules 2. Auto ends up oversubscribing really badly.
> 
> Yikes! Okay that makes sense! (sorry I see you tried that here: [e77fcdc](https://github.com/glotzerlab/signac/commit/e77fcdcaf5b77546fd5a9802fce759020286ddf8))

As far as I know, this is specific to circleci. GitHub Actions and Azure pipelines use VMs that advertise the correct number of cores.

vyasr:
I think we can probably run with this change for now and revisit if we ever switch to a different CI executor, then?

joaander:
> I think we can probably run with this change for now and revisit if we ever switch to a different CI executor, then?

Yes. I wasn't suggesting action - just providing information for context.

changelog.txt
```diff
@@ -15,6 +15,7 @@ Added
 
  - Official support for Python 3.10 (#631).
  - Benchmarks can be run using the ``asv`` (airspeed velocity) tool (#629).
+ - Tests run in parallel with ``pytest-xdist`` (#705).
```
cbkerr, line 18: Because it doesn't affect running locally:
```suggestion
 - Continuous integration tests run in parallel with ``pytest-xdist`` (#705).
```

cbkerr (approved):
Nice speedup! Make the changelog entry more specific then this is good to go!

csadorf (commented):
Great! Thanks a lot.

Do you know what is going on with the failing py38 test? Looks like some multi-threaded tests are failing?

vyasr:
> Great! Thanks a lot.
> 
> Do you know what is going on with the failing py38 test? Looks like some multi-threaded tests are failing?

It looks like one of the errors is that [this line](https://github.com/glotzerlab/signac/blob/master/tests/test_synced_collections/synced_collection_test.py#L484) in the test should be modified to catch a KeyError. The existing except clause isn't broad enough in detecting the ways in which multithreaded execution can fail when multithreading support is disabled.

My best guess is that the other failure is due to #397, specifically the case that I decided was out of scope when closing that issue: process-parallelism safe buffering. I suspect that we only see failures on PyPy because it executes much faster than CPython. I observed the same type of issue many times when constructing this test, specifically because sequences of operations that could fail when multithreading support was disabled would succeed in PyPy because threads completed too quickly to be preempted and so data was never corrupted even without acquiring mutexes. 

Buffering safe for multiple processes is still out of scope (especially in conjunction with PyPy), so I don't think diagnosing the exact failure here is worth much further investigation. My recommendation is to check if we're using pytest-xdist (either using the `PYTEST_XDIST_WORKER` environment variable or another mechanism described [in the pytest-xdist docs](https://pypi.org/project/pytest-xdist/#identifying-the-worker-process-during-a-test)) and whether we're running in PyPy (the `platform` check is already being done for other things in this file), and if both of those conditions are true just skip the multithreaded test in buffered mode. Note that we do not want to skip the multithreaded test in unbuffered collections, so you'll probably need to do something like override the test defined in `SyncedCollectionTest` with another test defined in `BufferedJSONCollectionTest` that skips the xdist+PyPy case and otherwise calls the `super()` test.

vyasr:
The same failure of the multithreaded test with PyPy is showing up on #706. It's possible that tests are suddenly detecting a failure in the multithreading logic that they haven't for the past year, although I don't know why that would be. A possible culprit is #658, which bumped our testing environment for pypy from 3.7 to 3.8. Perhaps something is different in PyPy 3.8 vs 3.7.

Now that we've seen this failure on two PRs, my recommendation would be to merge this PR **as is** and to investigate further in a separate PR. The easiest solution would just be to disable threading support on PyPy if we can't find an underlying cause. I invested a lot of time trying to get to the bottom of this issue in the past and I am unlikely to do so again in the near future.

vyasr:
It looks like the first issue that I diagnosed was the underlying problem. I pushed a fix in #710 that resolved the problem for me locally, let's see if tests pass on CI.

tommy-waltmann (approved):

PR #706, opened 2022-03-03
Change Project constructor to use root directory

vyasr:
## Description

The Project constructor no longer requires a config object. Projects are constructed using a root directory alone.

## Motivation and Context

Part of #643. Since projects no longer have ids and are entirely identified by their root directory in signac 2.0, the constructor can be simplified to reflect that.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -170,10 +170,7 @@ class Project:
         return str(self.root_directory())
 
     def __repr__(self):
-        # TODO: Fix this too.
-        return "{type}.get_project({root})".format(
-            type=self.__class__.__name__, root=repr(self.root_directory())
-        )
+        return f"{self.__class__.__name__}({repr(self.root_directory())})"
 
     def _repr_html_(self):
         """Project details in HTML format for use in IPython environment.
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -89,15 +89,16 @@ class _ProjectConfig(Config):
 class Project:
     """The handle on a signac project.
 
-    Application developers should usually not need to
-    directly instantiate this class, but use
-    :meth:`~signac.get_project` instead.
+    A :class:`Project` may only be constructed in a directory that is already a
+    signac project, i.e. a directory in which :func:`~signac.init_project` has
+    already been run. If project discovery (searching upwards in the folder
+    hierarchy until a project root is discovered) is desired, users should
+    instead invoke :func:`~signac.get_project` or :meth:`Project.get_project`.
 
     Parameters
     ----------
-    config :
-        The project configuration to use. By default, it loads the first signac
-        project configuration found while searching upward from the current
+    root : str, optional
+        The project root directory. By default, the current working directory
         working directory (Default value = None).
 
     """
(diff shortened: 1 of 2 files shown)


signac/contrib/project.py
```diff
@@ -1608,7 +1609,7 @@ def get_project(cls, root=None, search=True, **kwargs):
                 )
             )
 
-        return cls(config=config, **kwargs)
+        return cls(root=config["project_dir"], **kwargs)
```
bdice, line 1612: The current approach finds and loads a project config and then uses its `"project_dir"` key to construct the `Project`. Can this be simplified somehow, so that we check for config _existence_ and then pass the `root` to the constructor? We shouldn't need to load the config twice (in `get_project` and again in the constructor), right?

tests/test_shell.py
```diff
@@ -49,6 +49,7 @@ def setUp(self, request):
             pythonpath = [os.getcwd()] + pythonpath.split(":")
         os.environ["PYTHONPATH"] = ":".join(pythonpath)
         self.tmpdir = TemporaryDirectory(prefix="signac_")
+        self.project_name = f"Project({repr(os.path.realpath(self.tmpdir.name))})"
```
bdice, line 52: Is this attribute used? I don't see what this line is meant to do.

bdice (changes requested):

tests/test_shell.py
vyasr: Probably an unnecessary holdover from rebasing.

signac/contrib/project.py
vyasr: Hmm that's a good point. This question is related to the last piece of #643, which is changing `load_config` to not populate the `project_dir` into the config but to instead change `load_config` to return a tuple of `(root_dir, config)`. The reason I originally wanted to make that change is because actually modifying the contents of the config with something that isn't contained in any of the config files is an unexpected side effect that could cause confusion. Perhaps your question here is an indication that the discovery aspect should be separated into a completely separate function, and that `load_config` would stop discovering altogether? Alternatively, `load_config` could still be allowed to discover even if we added a separate `discover_config` function, but that seems like unnecessary redundancy.

signac/contrib/project.py
```diff
@@ -89,15 +89,16 @@ def __setitem__(self, key, value):
...
+    hierarchy until a project root is discovered) is desired, users should
+    instead invoke :func:`~signac.get_project` or :meth:`Project.get_project`.
 
     Parameters
     ----------
-    config :
-        The project configuration to use. By default, it loads the first signac
-        project configuration found while searching upward from the current
+    root : str, optional
+        The project root directory. By default, the current working directory
```
kidrahahjo, line 101:
```suggestion
        The project root directory. By default, the current
```
Is this repeated in line 102?

signac/contrib/project.py
vyasr: @bdice let me know what you think here. I can split those changes into a separate PR if that's the direction that we want to go.

tests/test_shell.py
vyasr: I think it's a holdover from an old commit where the `__str__` was defined differently.

signac/contrib/project.py
vyasr: Ha good catch thanks.

signac/contrib/project.py
bdice: It might be a better design to separate discovery from config loading (parsing). A separate PR might be best, I agree.

bdice (approved):

kidrahahjo (approved):

PR #708, opened 2022-03-04
Move internal signac files into .signac subdirectory

### LINKED ISSUE #197: Use more compact schema for root directory files

vyasr:
## Description

This PR moves the shell history and state point cache files into the .signac directory.

## Motivation and Context

Resolves #197. I have intentionally chosen to do nothing with `signac_statepoints.json` since that file would only have been created by users actually calling the `write_statepoints` function and therefore I think we should leave it to users to delete that file if it is present.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -93,6 +93,8 @@ Synchronize your project with the temporary project, for example with:
 )
 
 
+SHELL_HISTORY_FN = os.sep.join((".signac", "shell_history"))
+
 warnings.simplefilter("default")
 
 
@@ -864,7 +866,7 @@ def main_shell(args):
         else:  # interactive
             if READLINE:
                 if "PyPy" not in platform.python_implementation():
-                    fn_hist = project.fn(".signac_shell_history")
+                    fn_hist = project.fn(SHELL_HISTORY_FN)
                     try:
                         readline.read_history_file(fn_hist)
                         readline.set_history_length(1000)
--- a/signac/contrib/migration/v1_to_v2.py
+++ b/signac/contrib/migration/v1_to_v2.py
@@ -50,3 +50,12 @@ def _migrate_v1_to_v2(root_directory):
     v2_fn = _get_project_config_fn(root_directory)
     os.mkdir(os.path.dirname(v2_fn))
     os.rename(v1_fn, v2_fn)
+
+    # Now move all other files.
+    files_to_move = {
+        ".signac_shell_history": os.sep.join((".signac", "shell_history")),
+    }
+    for src, dst in files_to_move.items():
+        os.rename(
+            os.sep.join((root_directory, src)), os.sep.join((root_directory, dst))
+        )
--- a/signac/contrib/migration/v1_to_v2.py
+++ b/signac/contrib/migration/v1_to_v2.py
@@ -49,14 +49,16 @@ def _migrate_v1_to_v2(root_directory):
     v1_fn = os.path.join(root_directory, "signac.rc")
     v2_fn = _get_project_config_fn(root_directory)
     os.mkdir(os.path.dirname(v2_fn))
-    os.rename(v1_fn, v2_fn)
+    os.replace(v1_fn, v2_fn)
 
     # Now move all other files.
     files_to_move = {
         ".signac_shell_history": os.sep.join((".signac", "shell_history")),
-        ".signac_sp_cache.json.gz": os.sep.join((".signac", "sp_cache.json.gz")),
+        ".signac_sp_cache.json.gz": os.sep.join(
+            (".signac", "statepoint_cache.json.gz")
+        ),
     }
     for src, dst in files_to_move.items():
-        os.rename(
+        os.replace(
             os.sep.join((root_directory, src)), os.sep.join((root_directory, dst))
         )
--- a/signac/contrib/migration/v1_to_v2.py
+++ b/signac/contrib/migration/v1_to_v2.py
@@ -54,6 +54,7 @@ def _migrate_v1_to_v2(root_directory):
     # Now move all other files.
     files_to_move = {
         ".signac_shell_history": os.sep.join((".signac", "shell_history")),
+        ".signac_sp_cache.json.gz": os.sep.join((".signac", "sp_cache.json.gz")),
     }
     for src, dst in files_to_move.items():
         os.rename(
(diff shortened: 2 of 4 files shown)


signac/__main__.py
```diff
@@ -93,6 +93,8 @@
 )
 
 
+SHELL_HISTORY_FN = os.sep.join((".signac", "shell_history"))
```
bdice, line 96: I'd like to see this definition centralized, like in `project.py` next to `FN_CACHE`. That would eliminate the duplicate definition between this file and `migration/v1_to_v2.py`.

signac/contrib/migration/v1_to_v2.py
```diff
@@ -50,3 +50,13 @@ def _migrate_v1_to_v2(root_directory):
     v2_fn = _get_project_config_fn(root_directory)
     os.mkdir(os.path.dirname(v2_fn))
     os.rename(v1_fn, v2_fn)
+
+    # Now move all other files.
+    files_to_move = {
+        ".signac_shell_history": os.sep.join((".signac", "shell_history")),
+        ".signac_sp_cache.json.gz": os.sep.join((".signac", "sp_cache.json.gz")),
```
bdice, line 57: If it's being renamed anyway, especially in a hidden folder, we can be more explicit about the name.
```suggestion
        ".signac_sp_cache.json.gz": os.sep.join((".signac", "statepoint_cache.json.gz")),
```

signac/contrib/project.py
```diff
@@ -110,7 +110,7 @@ class Project:
     KEY_DATA = "signac_data"
     "The project's datastore key."
 
-    FN_CACHE = ".signac_sp_cache.json.gz"
+    FN_CACHE = os.sep.join((".signac", "sp_cache.json.gz"))
```
bdice, line 113:
```suggestion
    FN_CACHE = os.sep.join((".signac", "statepoint_cache.json.gz"))
```

tests/test_project.py
```diff
@@ -2491,13 +2495,34 @@ def test_project_schema_version_migration(self, implicit_version):
...
 
+            # Create a shell history file.
+            history_fn = os.path.join(dirname, ".signac_shell_history")
+            with open(history_fn, "w") as f:
+                f.write(
+                    textwrap.dedent(
+                        """\
+                        print(project)"""
+                    )
+                )
```
bdice, line 2506: Since there are no newlines in the text being written, we should just do this:
```suggestion
                f.write("print(project)")
```

tests/test_project.py
```diff
@@ -23,6 +24,7 @@
 from test_job import TestJobBase
 
 import signac
+from signac.__main__ import SHELL_HISTORY_FN
```
bdice, line 27: In general, we should not import from `__main__` modules. See previous comment about centralizing this definition.

signac/contrib/migration/v1_to_v2.py
```diff
@@ -50,3 +50,13 @@ def _migrate_v1_to_v2(root_directory):
...
+    }
+    for src, dst in files_to_move.items():
+        os.rename(
```
bdice, line 60: There are some subtleties about [`os.rename`](https://docs.python.org/3/library/os.html#os.rename) vs. [`os.replace`](https://docs.python.org/3/library/os.html#os.replace) on Windows. I think we always use `os.replace` in signac. I see `os.rename` also comes up elsewhere in this file, so we might want to double-check that globally.
```suggestion
        os.replace(
```

tests/test_project.py
```diff
@@ -2491,13 +2495,34 @@ def test_project_schema_version_migration(self, implicit_version):
...
+                    )
+                )
+
+            # Create a statepoint cache. Note that this cache does not
+            # correspond to actual statepoints since we don't currently have
+            # any in this project, but that's fine for migration testing.
+            history_fn = os.path.join(dirname, ".signac_sp_cache.json.gz")
+            sp = {"a": 1}
+            with gzip.open(history_fn, "wb") as f:
+                f.write(json.dumps({calc_id(sp): sp}).encode())
```
bdice, line 2514: This might be a better regression test if it did not re-use any internal machinery from the current signac implementation? Perhaps this only matters if we change our data schema more dramatically.
```suggestion
                f.write(b'{"42b7b4f2921788ea14dac5566e6f06d0": {"a": 1}}')
```

bdice (changes requested):

signac/__main__.py
vyasr: The migration file intentionally doesn't import these kinds of variables from anywhere. We don't import `FN_CACHE` either. By avoiding building that dependency in we don't have to update these when signac's internal schema changes, and it provides a more accurate record of what the schema migration contains.

vyasr: I guess that doesn't address your first sentence. Why would you put a variable in `project.py` that has nothing to do with the project in general, only the command line interface? That would be out of place.

tests/test_project.py
vyasr: See my reply. I don't think moving the definition just for the sake of the test makes sense; if it comes down to that, perhaps we really should be hardcoding the values into the test instead.

signac/contrib/migration/v1_to_v2.py
vyasr: Good call, I forgot about this.

tests/test_project.py
vyasr: We're not testing the contents of this file at all right now. We could make the test more stringent to validate that the file isn't corrupted during the migration, I just didn't go that far. I'm not sure that it's worth the effort right now.

signac/__main__.py
bdice: Okay, I see your argument. This can be resolved.

tests/test_project.py
bdice: Based on your argument on the other thread, I’m fine with hardcoding it.

tests/test_project.py
bdice: Okay. Maybe add a TODO, otherwise I agree we don’t need to invest more effort here.

vyasr: I think it's probably not worth worrying about. Worst case scenario, the user just has to run update-cache again.

bdice (approved):

PR #710, opened 2022-03-05
Catch KeyError as another possible failure of multithreading

vyasr:
## Description

When multithreading support is disabled, it is possible for operations on synced dicts to fail with a KeyError (as well as the other errors that were already being caught), so the multithreading test now catches that failure. Note that this is a bug in the test only, not in signac itself.

## Motivation and Context

PyPy3 tests of multithreading have been failing due to this omission (see #705 and #706). Additionally, once this failure occurs, the buffer is in an invalid state, so future multithreading tests can fail in unpredictable ways. I reproduced the issue by running `test_json_buffered_collection.py::TestBufferedJSONDictWriteConcern::test_multithreaded` using pytest-repeat to run the test 100 times. I invariably observed failures within this number of tests. I then verified that catching the KeyError and clearing the collection prevents the error by running the test a few times with 1000 repeats (with pytest-xdist just to make sure that #705 was safe). I observed no further errors.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/tests/test_synced_collections/synced_collection_test.py
+++ b/tests/test_synced_collections/synced_collection_test.py
@@ -481,7 +481,7 @@ class SyncedDictTest(SyncedCollectionTest):
         try:
             with ThreadPoolExecutor(max_workers=num_threads) as executor:
                 list(executor.map(set_value, [synced_collection] * num_threads * 10))
-        except (RuntimeError, JSONDecodeError):
+        except (RuntimeError, JSONDecodeError, KeyError):
             # This line may raise an exception, or it may successfully complete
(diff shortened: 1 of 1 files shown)


bdice:
@vyasr What did the `KeyError` say in the message/error value?

bdice (approved):
LGTM, this is probably safe to merge if it fixed the issue locally for you. I am curious what the `KeyError` said.

vyasr:
@bdice have a look at the failure on #705, you can see the failure there.

PR #711, opened 2022-03-05
Move discovery to separate functions.

vyasr:
## Description

This PR removes the local vs global search option from `config.load_config` API in favor of a separate function to handle discovery. It also removes the `project_dir` key that was previously being automatically populated into the config when calling `load_config`.

## Motivation and Context

Populating the `project_dir` key is an unexpected side effect since it adds a key that doesn't actually exist in any config file. One option would be to change `load_config` to return a tuple containing `(root, config)` instead of just the config, but that mixes two distinct pieces of functionality and also makes config loading more expensive than it needs to be. By separating the two functions, calling code can find the config file and load it separately.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -363,7 +363,7 @@ def main_sync(args):
     """Handle sync subcommand."""
     # TODO: This function appears to be untested.
     #
-    # Valid provided argument combinations
+    # Validate provided argument combinations
     #
     if args.archive:
         args.recursive = True
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -412,6 +412,7 @@ def main_sync(args):
         destination = get_project(root=args.destination)
     except LookupError:
         if args.allow_workspace:
+            # TODO: Remove allow_workspace entirely.
             destination = Project(os.path.relpath(args.destination))
         else:
             _print_err(
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -361,6 +361,7 @@ def main_schema(args):
 
 def main_sync(args):
     """Handle sync subcommand."""
+    # TODO: This function appears to be untested.
     #
     # Valid provided argument combinations
     #
@@ -411,13 +412,7 @@ def main_sync(args):
         destination = get_project(root=args.destination)
     except LookupError:
         if args.allow_workspace:
-            destination = Project(
-                config={
-                    "project": os.path.relpath(args.destination),
-                    "project_dir": args.destination,
-                    "workspace_dir": ".",
-                }
-            )
+            destination = Project(os.path.relpath(args.destination))
         else:
             _print_err(
                 "WARNING: The destination appears to not be a project path. "
@@ -723,7 +718,7 @@ def main_config_show(args):
     elif args.globalcfg:
         cfg = config.read_config_file(config.USER_CONFIG_FN)
     else:
-        cfg = config.load_config()
+        cfg = config.load_config(config._locate_config_dir(os.getcwd()))
     if not cfg:
         if args.local:
             mode = "local"
@@ -756,7 +751,7 @@ def main_config_verify(args):
     elif args.globalcfg:
         cfg = config.read_config_file(config.USER_CONFIG_FN)
     else:
-        cfg = config.load_config()
+        cfg = config.load_config(config._locate_config_dir(os.getcwd()))
     if not cfg:
         if args.local:
             mode = "local"
--- a/signac/common/config.py
+++ b/signac/common/config.py
@@ -20,25 +20,6 @@ def _get_project_config_fn(root):
     return os.path.abspath(os.path.join(root, PROJECT_CONFIG_FN))
 
 
-def _contains_config_file(root):
-    """Determine if the root directory contains a signac configuration file.
-
-    Parameters
-    ----------
-    root : str
-        Path to search. Uses ``os.getcwd()`` if None (Default value = None).
-
-    Returns
-    --------
-    str or None
-        The path to the configuration file if one is found, otherwise None.
-    """
-    fn_ = _get_project_config_fn(root)
-    if os.path.isfile(fn_):
-        return root
-    return None
-
-
 def _locate_config_dir(root):
     """Locates root directory containing a signac configuration file in a directory hierarchy.
 
@@ -52,10 +33,13 @@ def _locate_config_dir(root):
     str or None
         The root directory containing the configuration file if one is found, otherwise None.
     """
+    root = os.path.abspath(root)
     while True:
-        if _contains_config_file(root):
+        if os.path.isfile(_get_project_config_fn(root)):
             return root
-        up = os.path.abspath(os.path.join(root, ".."))
+        # TODO: Could use the walrus operator here when we completely drop
+        # Python 3.7 support if we like the operator.
+        up = os.path.dirname(root)
         if up == root:
             logger.debug("Reached filesystem root, no config found.")
             return None
@@ -66,18 +50,19 @@ def _locate_config_dir(root):
 def read_config_file(filename, configspec=None, *args, **kwargs):
     """Read a configuration file.
 
-        Parameters
-        ----------
-        filename : str
-            The path to the file to read.
-        configspec : List[str], optional
-            The key-value pairs supported in the config.
+    Parameters
+    ----------
+    filename : str
+        The path to the file to read.
+    configspec : List[str], optional
+        The key-value pairs supported in the config. If None, uses the default
+        pairs supported by signac, see :data:`signac.common.validate.cfg`
+        (Default value = None).
 
-        Returns
-        --------
-        :class:`Config`
-
-    The config contained in the file.
+    Returns
+    --------
+    :class:`Config`
+        The config contained in the file.
     """
     logger.debug(f"Reading config file '{filename}'.")
     if configspec is None:
@@ -99,31 +84,31 @@ def read_config_file(filename, configspec=None, *args, **kwargs):
 
 
 def load_config(root=None):
-    """Load configuration at root directory.
+    """Load configuration from a project directory.
 
     Parameters
     ----------
     root : str
-        The path from which to load the local config.
+        The project path to pull project-local configuration data from.
 
     Returns
     --------
     :class:`Config`
-        The composite configuration including both local and global config data
-        if requested. Note that because this config is a composite,
+        The composite configuration including both project-local and global
+        config data if requested. Note that because this config is a composite,
         modifications to the returned value will not be reflected in the files.
     """
     if root is None:
         root = os.getcwd()
     config = Config(configspec=cfg.split("\n"))
 
-    # Add in any global or user config files. For now this search only finds user-specific
+    # Add in any global or user config files. For now this only finds user-specific
     # files, but it could be updated in the future to support e.g. system-wide config files.
     for fn in (USER_CONFIG_FN,):
         if os.path.isfile(fn):
             config.merge(read_config_file(fn))
 
-    if _contains_config_file(root):
+    if os.path.isfile(_get_project_config_fn(root)):
         config.merge(read_config_file(_get_project_config_fn(root)))
     return config
 
--- a/signac/common/config.py
+++ b/signac/common/config.py
@@ -20,65 +20,64 @@ def _get_project_config_fn(root):
     return os.path.abspath(os.path.join(root, PROJECT_CONFIG_FN))
 
 
-def _get_config_dirname(fn):
-    # We could use pathlib's `.paths` attribute to remove the trailing project
-    # config filename, but that has significant performance
-    # implications: this takes ~200 ns, whereas
-    # os.path.join(*(pathlib.PosixPath(fn).parts[:-2])) takes ~6 us, almost a
-    # 20x slowdown.
-    return fn.replace(os.sep + PROJECT_CONFIG_FN, "")
-
-
-def _search_local(root):
-    fn_ = _get_project_config_fn(root)
-    if os.path.isfile(fn_):
-        yield fn_
-
-
-def _search_tree(root=None):
-    """Locates signac configuration files in a directory hierarchy.
+def _contains_config_file(root):
+    """Determine if the root directory contains a signac configuration file.
 
     Parameters
     ----------
     root : str
         Path to search. Uses ``os.getcwd()`` if None (Default value = None).
 
+    Returns
+    --------
+    str or None
+        The path to the configuration file if one is found, otherwise None.
     """
-    if root is None:
-        root = os.getcwd()
+    fn_ = _get_project_config_fn(root)
+    if os.path.isfile(fn_):
+        return root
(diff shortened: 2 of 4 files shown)


signac/common/config.py
```diff
@@ -20,65 +20,64 @@ def _get_project_config_fn(root):
...
+    str or None
+        The root directory containing the configuration file if one is found, otherwise None.
     """
-    if root is None:
-        root = os.getcwd()
     while True:
-        yield from _search_local(root)
+        if _contains_config_file(root):
+            return root
         up = os.path.abspath(os.path.join(root, ".."))
```
bdice, line 58: We should create an absolute path once, before entering the `while True:`, then use `os.path.dirname` each time we want to go up. Performance is better if we (1) use `dirname` to go up instead of a join with `".."` and (2) call `abspath` fewer times.

```python
root = os.path.abspath(root)
while True:
    # ...
    up = os.path.dirname(root)
    # ...
```

bdice: https://github.com/bdice/python_microbenchmarks/blob/main/microbenchmarks/os_path_dirname.py

Relative cost of each function call:
```
dirname: 0.0445243
join: 0.0719918
abs-dirname: 0.176682
abs-join: 0.234717
```

signac/common/config.py
```diff
@@ -20,65 +20,64 @@ def _get_project_config_fn(root):
...
-def _search_tree(root=None):
-    """Locates signac configuration files in a directory hierarchy.
+def _locate_config_dir(root):
+    """Locates root directory containing a signac configuration file in a directory hierarchy.
 
     Parameters
     ----------
     root : str
-        Path to search. Uses ``os.getcwd()`` if None (Default value = None).
+        Starting path to search.
```
bdice, line 48: Do we need/want to use [`os.path.expanduser`](https://docs.python.org/3/library/os.path.html#os.path.expanduser) or [`os.path.expandvars`](https://docs.python.org/3/library/os.path.html#os.path.expandvars) in any of our path handling? I think the answer is no (users can do that if needed), but I wanted to ask anyway.

signac/common/config.py
```diff
@@ -20,65 +20,64 @@ def _get_project_config_fn(root):
...
 def read_config_file(filename, configspec=None, *args, **kwargs):
     """Read a configuration file.
 
-    Parameters
-    ----------
-    filename : str
-        The path to the file to read.
-    configspec : List[str], optional
-        The key-value pairs supported in the config.
+        Parameters
```
bdice, line 69: It looks like an extra indentation was added to the next few lines. The return section was also messed up?

signac/common/config.py
```diff
@@ -99,45 +98,33 @@ def read_config_file(filename, configspec=None, *args, **kwargs):
     return config
 
 
-def load_config(root=None, local=False):
-    """Load configuration, searching upward from a root path if desired.
+def load_config(root=None):
+    """Load configuration at root directory.
```
bdice, line 102:
```suggestion
    """Load configuration from a project directory.
```

signac/common/config.py
```diff
@@ -99,45 +98,33 @@ def read_config_file(filename, configspec=None, *args, **kwargs):
...
 
     Parameters
     ----------
     root : str
-        The path from which to begin searching for config files.
-    local : bool, optional
-        If ``True``, only search in the provided directory and do not traverse
-        upwards through the filesystem (Default value: False).
+        The path from which to load the local config.
```
bdice, line 107: This sounds like it expects a config path, but I think it takes a project path. Also the use of the word "local" feels a bit out of place.

signac/common/config.py
```diff
@@ -99,45 +98,33 @@ def read_config_file(filename, configspec=None, *args, **kwargs):
...
 
     Returns
     --------
     :class:`Config`
         The composite configuration including both local and global config data
-        if requested.
+        if requested. Note that because this config is a composite,
+        modifications to the returned value will not be reflected in the files.
```
bdice, line 114: Modifications to a `Config` object are _never_ reflected in a file, right? It's not like a synced collection iirc, but has methods to write to a file? Can't remember.

signac/common/config.py
```diff
@@ -99,45 +98,33 @@ def read_config_file(filename, configspec=None, *args, **kwargs):
...
     """
     if root is None:
         root = os.getcwd()
     config = Config(configspec=cfg.split("\n"))
```
bdice, line 118: We need to rename `cfg` to indicate it's a module-level constant, like `CONFIG_SCHEMA` or similar. It took me a moment to recognize what was going on here.

signac/common/config.py
```diff
@@ -99,45 +98,33 @@ def read_config_file(filename, configspec=None, *args, **kwargs):
...
-    if local:
-        search_func = _search_local
-    else:
-        for fn in _search_standard_dirs():
+
+    # Add in any global or user config files. For now this search only finds user-specific
```
bdice, line 120: Slightly redundant...?
```suggestion
    # Add in any global or user config files. For now this only finds user-specific
```

signac/contrib/project.py
```diff
@@ -1599,18 +1606,20 @@ def get_project(cls, root=None, search=True, **kwargs):
...
+            old_root = root
+            root = _locate_config_dir(root)
+            if not root:
+                raise LookupError(
+                    f"Unable to find project at path '{os.path.abspath(old_root)}'."
                 )
-            )
 
-        return cls(root=config["project_dir"], **kwargs)
+        return cls(root=root, **kwargs)
```
bdice, line 1622: Simplify this. I think this is equivalent? Maybe there's something even better in between the current state and this rewrite. My primary goal is to raise the `LookupError` only in one path.
```python
config_dir = _locate_config_dir(root) if search else root
if config_dir is None or not _contains_config_file(config_dir):
    raise LookupError(f"Unable to find project at path '{os.path.abspath(config_dir)}'.")
return cls(root=config_dir, **kwargs)
```

bdice (changes requested):
First round of feedback. Thanks for working on this. :)

signac/common/config.py
```diff
@@ -99,45 +98,33 @@ def read_config_file(filename, configspec=None, *args, **kwargs):
     return config
 
 
-def load_config(root=None, local=False):
-    """Load configuration, searching upward from a root path if desired.
+def load_config(root=None):
```
bdice, line 101: Does this file need to have _any_ public methods?

vyasr:
This is odd. A number of my changes got discarded somehow, leading to some of the weird half-done work that you commented on. Doesn't just look like uncommitted changes either. Not sure what happened, but in any case I'll fix and push. Most of your requests were changes I thought I made :(

signac/common/config.py
vyasr: Good call.

signac/common/config.py
vyasr: See if the new wording is clearer to you.

signac/common/config.py
vyasr: The `configobj` API is pretty confusing in this respect. You are correct in the sense that it will not automatically reflect changes to a file. However, there `ConfigObj.write` behaves differently depending on whether you initialize the object with a filename or not. If you do, then `write` will create if necessary and write to the file. Otherwise, if you just do something like `c = ConfigObj(); c['a'] = 'b'; c.write()` it will just print the output. That means that a `Config` returned from `load_config` can't be written back to file, while a `Config` returned from `read_config_file` can be. I've been bitten by this many times in the past, so I wanted to add a note. I am open to a more robust solution in the long term that avoids this though, and I'm hopeful that whatever package we use for TOML support when we switch will allow us to avoid this problem. Alternatively, maybe in 2.0 we can remove all functions that return writeable configs and force any mutating code path to invoke the underlying TOML library directly. I'm not sure what the best approach is here, but for now I figured documentation was a start.

bdice: Sounds good to me.

signac/common/config.py
vyasr: Agreed. It is a public attribute, though, which isn't something that's really feasible to deprecate with a warning. I think we'll just have to rename this in 2.0. Chances are nobody is using it, but I know signac-flow used to (I believe that I removed that dependence recently when I removed monkey-patching), which is why I'm advocating treating it as a breaking change for 2.0 rather than something to put in this PR.

signac/contrib/project.py
vyasr: The problem is that the error message becomes useless in the case that `_locate_config_dir` returns `None`. I disliked this code as well but couldn't come up with an alternative that gets around that elegantly. You could do something like this:
```
old_root = root
root = _locate_config_dir(root) if search else root
if root is not None or not os.path.isfile(_get_project_config_fn(root)):
    raise LookupError(
        f"Unable to find project at path '{os.path.abspath(old_root)}'."
    )
```
which gets around my above objection, but it still has the downside of adding an extra pair of `os` calls in the `search` case because you'll run the `os.path.isfile(_get_project_config_fn(root))` even though the current value of `root` (the output of `_locate_config_dir`) is already sufficient information. I don't think that's worthwhile. The best we can do is just set a local `err` variable in both branches and raise the `LookupError` outside the `if-else`, which sort of addresses your concern but still isn't great.

signac/common/config.py
vyasr: That is a fair question. Ultimately this boils down to a question of how we expect users to interact with config files in signac 2.0. The files will be hidden, so most casual users probably won't think to do something like
```
import toml  # or whatever package
config = toml.open("/path/to/project/.signac/config") # or whatever the API looks like
```
We could really expect them to treat it like git and edit directly with a text editor, but that's also not really entry-level. I only muck with the git config manually if I really have to. That leaves the command-line config interface. I'm fine with that being the primary entry point as it currently is, the question is whether we're OK with that being the _only_ entry point. Is that what we want?

signac/__main__.py
```diff
@@ -409,13 +410,7 @@ def _sig(st):
...
     except LookupError:
         if args.allow_workspace:
-            destination = Project(
-                config={
-                    "project": os.path.relpath(args.destination),
-                    "project_dir": args.destination,
-                    "workspace_dir": ".",
-                }
-            )
+            destination = Project(os.path.relpath(args.destination))
```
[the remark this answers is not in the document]
vyasr: Good point, I obviously didn't really think this through. I'm not sure what to do with this. In the case where the workspace is a direct subdirectory of the project we could still support this (for the lifetime of 1.x) by doing
```
try:
    destination = Project(os.path.dirname(os.path.relpath(args.destination)))
except LookupError:
    ...
```
Is that worthwhile to avoid a breaking change here?

signac/common/config.py
vyasr: I'm inclined to agree, let users do that. It's simpler and saves us from incurring the cost unconditionally. However, this point may be moot after we decide what to do on your last comment about whether any of these should be public APIs.

signac/common/config.py
bdice: I would be fine with making it CLI / manual editing only. We have never recommended using these APIs, all our docs and error/warning messages point to using the CLI.

signac/contrib/project.py
bdice: Ultimately this is pretty minor and I don’t want to get hung up on this. If you think one of these alternatives is better, go for it.

signac/__main__.py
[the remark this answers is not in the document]
bdice: I personally think the “allow workspace” feature is a little unintuitive (or unexpected?) when everything else in signac is based on a project. There are no other functions I know of that deal with raw workspace directories. I’d be fine with removing it. I think @csadorf might have added this feature, so maybe helpful to get another opinion on whether we need it in 2.0.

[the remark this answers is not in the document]
bdice: As for the lifetime of 1.x, this branch targets 2.0, right? I don’t see the connection. A breaking change is allowable.

signac/common/config.py
```diff
@@ -20,51 +20,33 @@ def _get_project_config_fn(root):
...
-
-def _search_local(root):
-    fn_ = _get_project_config_fn(root)
-    if os.path.isfile(fn_):
-        yield fn_
-
-
-def _search_tree(root=None):
-    """Locates signac configuration files in a directory hierarchy.
+def _locate_config_dir(root):
```
bdice, line 23: A better name might be `search_path`. The _returned_ value is the "root" of the project, not the argument here.
```suggestion
def _locate_config_dir(search_path):
```

signac/__main__.py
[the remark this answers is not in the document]
bdice: It's been a couple days since I looked at this but I hadn't finished my review. I think the best action to take here is to remove `--allow-workspace` in a separate PR. We could do that first, or we can mark this as TODO and leave it partially broken with an issue to fix it later.

signac/common/config.py
```diff
@@ -73,7 +55,9 @@ def read_config_file(filename, configspec=None, *args, **kwargs):
     filename : str
         The path to the file to read.
     configspec : List[str], optional
-        The key-value pairs supported in the config.
+        The key-value pairs supported in the config. If None, uses the default
+        pairs supported by signac, see :data:`signac.common.validate.cfg`
+        (Default value = None).
```
bdice, line 60: Are `configspec, *args, **kwargs` ever used? I don't see this function being called with anything except `filename` provided. Let's remove those arguments since they're not doing anything for us.

signac/common/config.py
bdice: I recommend moving towards making all these functions private. Could be done in a later PR.

bdice (changes requested):

signac/__main__.py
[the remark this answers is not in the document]
vyasr: > As for the lifetime of 1.x, this branch targets 2.0, right? I don’t see the connection. A breaking change is allowable.
Yeah sorry I was context switching between too many PRs simultaneously when I wrote that. You're right, we can break this. I'll leave a TODO item for a future PR.

signac/common/config.py
vyasr: Again, forgot this was a PR where breaks were fine. That said, I think this is another future TODO IMO.

signac/common/config.py
vyasr: Sounds good, I'll keep that on the docket for future work.

signac/__main__.py
```diff
@@ -361,6 +361,7 @@ def main_schema(args):
 
 def main_sync(args):
     """Handle sync subcommand."""
+    # TODO: This function appears to be untested.
     #
     # Valid provided argument combinations
```
bdice, line 366: Unrelated to this PR, but I think this is supposed to say:
```suggestion
    # Validate provided argument combinations
```

bdice (approved):
Nice. Let's merge this, so that we can unblock other work.

PR #714, opened 2022-03-14
Remove configurable workspace directory 

### LINKED ISSUE #390: Requiring workspace to be a subdirectory of the project

vyasr:
## Description

This PR removes the functionality for users to use a custom workspace directory. Going forward, the workspace will always be an immediate subdirectory of the project called "workspace".

## Motivation and Context

Resolves #390.

## Checklist:

### DIFF
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -409,7 +409,7 @@ def main_sync(args):
     try:
         destination = get_project(root=args.destination)
     except LookupError:
-        _print_err("WARNING: The destination appears to not be a project path. ")
+        _print_err("WARNING: The destination does not appear to be a project path.")
         raise
     selection = find_with_filter_or_none(args)
 
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -29,7 +29,7 @@ except ImportError:
 else:
     READLINE = True
 
-from . import Project, get_project, init_project
+from . import get_project, init_project
 from .common import config
 from .common.configobj import Section, flatten_errors
 from .contrib.filterparse import _add_prefix, parse_filter_arg
@@ -74,7 +74,6 @@ SHELL_BANNER = """Python {python_version}
 signac {signac_version} 🎨
 
 Project:\t{root_path}{job_banner}
-Workspace:\t{workspace_path}
 Size:\t\t{size}
 
 Interact with the project interface using the "project" or "pr" variable.
@@ -343,7 +342,7 @@ def main_view(args):
 
 def main_init(args):
     """Handle init subcommand."""
-    init_project(name=args.project_id, root=os.getcwd(), workspace=args.workspace)
+    init_project(name=args.project_id, root=os.getcwd())
     _print_err("Initialized project.")
 
 
@@ -410,21 +409,8 @@ def main_sync(args):
     try:
         destination = get_project(root=args.destination)
     except LookupError:
-        if args.allow_workspace:
-            destination = Project(
-                config={
-                    "project": os.path.relpath(args.destination),
-                    "project_dir": args.destination,
-                    "workspace_dir": ".",
-                }
-            )
-        else:
-            _print_err(
-                "WARNING: The destination appears to not be a project path. "
-                "Use the '-w/--allow-workspace' option if you want to "
-                "synchronize to a workspace directory directly."
-            )
-            raise
+        _print_err("WARNING: The destination appears to not be a project path. ")
+        raise
     selection = find_with_filter_or_none(args)
 
     if args.strategy:
@@ -552,7 +538,6 @@ def _main_import_interactive(project, origin, args):
                     signac_version=__version__,
                     job_banner="",
                     root_path=project.root_directory(),
-                    workspace_path=project.workspace(),
                     size=len(project),
                     origin=args.origin,
                 ),
@@ -894,7 +879,6 @@ def main_shell(args):
                     signac_version=__version__,
                     job_banner=f"\nJob:\t\t{job.id}" if job is not None else "",
                     root_path=project.root_directory(),
-                    workspace_path=project.workspace(),
                     size=len(project),
                 ),
             )
@@ -923,13 +907,6 @@ def main():
 
     parser_init = subparsers.add_parser("init")
     parser_init.add_argument("project_id", nargs="?", help=argparse.SUPPRESS)
-    parser_init.add_argument(
-        "-w",
-        "--workspace",
-        type=str,
-        default="workspace",
-        help="The path to the workspace directory.",
-    )
     parser_init.set_defaults(func=main_init)
 
     parser_project = subparsers.add_parser("project")
@@ -1469,13 +1446,6 @@ job documents."
         "--no-keys", action="store_true", help="Never overwrite any conflicting keys."
     )
 
-    parser_sync.add_argument(
-        "-w",
-        "--allow-workspace",
-        action="store_true",
-        help="Allow the specification of a workspace (instead of a project) directory "
-        "as the destination path.",
-    )
     parser_sync.add_argument(
         "--force", action="store_true", help="Ignore all warnings, just synchronize."
     )
--- a/signac/contrib/migration/v1_to_v2.py
+++ b/signac/contrib/migration/v1_to_v2.py
@@ -19,7 +19,6 @@ from .v0_to_v1 import _load_config_v1
 # A minimal v2 config.
 _cfg = """
 schema_version = string(default='0')
-workspace_dir = string(default='workspace')
 """
 
 
@@ -37,10 +36,28 @@ def _load_config_v2(root_directory):
 
 def _migrate_v1_to_v2(root_directory):
     """Migrate from schema version 1 to version 2."""
-    # Delete project name from config and store in project doc.
+    # Load the v1 config.
     cfg = _load_config_v1(root_directory)
     fn_doc = os.path.join(root_directory, Project.FN_DOCUMENT)
     doc = BufferedJSONAttrDict(filename=fn_doc, write_concern=True)
+
+    # Try to migrate a custom workspace directory if one exists.
+    current_workspace_name = cfg.get("workspace_dir")
+    if current_workspace_name is not None:
+        if current_workspace_name != "workspace":
+            current_workspace = os.path.join(root_directory, current_workspace_name)
+            new_workspace = os.path.join(root_directory, "workspace")
+            if os.path.exists(new_workspace):
+                raise RuntimeError(
+                    "Workspace directories are no longer configurable in schema version 2, and "
+                    f"must be 'workspace', but {new_workspace} already exists. Please remove or "
+                    f"move it so that the currently configured workspace directory "
+                    "{current_workspace} can be moved to {new_workspace}."
+                )
+            os.replace(current_workspace, new_workspace)
+        del cfg["workspace_dir"]
+
+    # Delete project name from config and store in project doc.
     doc["signac_project_name"] = cfg["project"]
     del cfg["project"]
     cfg.write()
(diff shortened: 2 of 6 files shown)


signac/__main__.py
```diff
@@ -410,21 +409,8 @@ def _sig(st):
...
-                }
-            )
-        else:
-            _print_err(
-                "WARNING: The destination appears to not be a project path. "
-                "Use the '-w/--allow-workspace' option if you want to "
-                "synchronize to a workspace directory directly."
-            )
-            raise
+        _print_err("WARNING: The destination appears to not be a project path. ")
```
bdice, line 412:
```suggestion
        _print_err("WARNING: The destination does not appear to be a project path.")
```

signac/contrib/migration/v1_to_v2.py
```diff
@@ -37,10 +36,28 @@ def _load_config_v2(root_directory):
...
+    if current_workspace_name is not None:
+        if current_workspace_name != "workspace":
+            current_workspace = os.path.join(root_directory, current_workspace_name)
+            new_workspace = os.path.join(root_directory, "workspace")
+            if os.path.exists(new_workspace):
+                raise RuntimeError(
+                    "Workspace directories are no longer configurable in schema version 2, and "
+                    f"must be 'workspace', but {new_workspace} already exists. Please remove or "
+                    f"move it so that the currently configured workspace directory "
+                    "{current_workspace} can be moved to {new_workspace}."
```
bdice, line 55:
```suggestion
                    f"{current_workspace} can be moved to {new_workspace}."
```

signac/contrib/project.py
```diff
@@ -1467,8 +1463,9 @@ def temporary_project(self, dir=None):
         with TemporaryProject(cls=type(self), dir=dir) as tmp_project:
             yield tmp_project
 
+    # TODO: Is there a reason to support the make_dir parameter?
```
bdice, line 1466: Do things fail if `make_dir` is false? I wasn't aware of this parameter. I think it would be fine to remove and always assume it's true.

tests/test_job.py
```diff
@@ -70,12 +70,10 @@ def setUp(self, request):
         self._tmp_dir = TemporaryDirectory(prefix="signac_")
         request.addfinalizer(self._tmp_dir.cleanup)
         self._tmp_pr = os.path.join(self._tmp_dir.name, "pr")
-        self._tmp_wd = os.path.join(self._tmp_dir.name, "wd")
+        self._tmp_wd = os.path.join(self._tmp_dir.name, "pr", "workspace")
```
bdice, line 73: At first I thought this was unused -- it's only used in `test_project.py` which has tests inheriting from this class, and it's only used twice. Maybe we can consolidate that somehow.

tests/test_project.py
```diff
@@ -245,13 +214,11 @@ def test_no_workspace_warn_on_find(self, caplog):
     @pytest.mark.skipif(WINDOWS, reason="Symbolic links are unsupported on Windows.")
     def test_workspace_broken_link_error_on_find(self):
         with TemporaryDirectory() as tmp_dir:
-            project = self.project_class.init_project(
-                root=tmp_dir, workspace="workspace-link"
-            )
-            os.rmdir(os.path.join(tmp_dir, "workspace-link"))
+            project = self.project_class.init_project(root=tmp_dir)
+            os.rmdir(os.path.join(tmp_dir, "workspace"))
```
bdice, line 218:
```suggestion
            os.rmdir(project.workspace())
```

tests/test_project.py
```diff
@@ -245,13 +214,11 @@ def test_no_workspace_warn_on_find(self, caplog):
...
             os.symlink(
                 os.path.join(tmp_dir, "workspace~"),
-                os.path.join(tmp_dir, "workspace-link"),
+                os.path.join(tmp_dir, "workspace"),
```
bdice, line 221:
```suggestion
                project.workspace(),
```

tests/test_project.py
```diff
@@ -2571,12 +2548,10 @@ def setUp_base_h5Store(self, request):
         self._tmp_dir = TemporaryDirectory(prefix="signac_")
         request.addfinalizer(self._tmp_dir.cleanup)
         self._tmp_pr = os.path.join(self._tmp_dir.name, "pr")
-        self._tmp_wd = os.path.join(self._tmp_dir.name, "wd")
+        self._tmp_wd = os.path.join(self._tmp_dir.name, "pr", "workspace")
```
bdice, line 2551:
```suggestion
```

bdice (changes requested):

signac/contrib/project.py
vyasr: Yeah you get a LookupError. I think we can safely remove it as well.

tests/test_job.py
vyasr: I just removed one of the places, which wasn't really testing anything useful, and hardcoded the path in the one test that verifies that the workspace directory is set to the expected path.

tests/test_project.py
```diff
@@ -102,29 +108,17 @@ def test_root_directory(self):
...
     def test_workspace_directory(self):
-        assert self._tmp_wd == self.project.workspace()
+        assert os.path.join(self._tmp_pr, "workspace") == self.project.workspace()
 
     def test_config_modification(self):
         # In-memory modification of the project configuration is
         # deprecated as of 1.3, and will be removed in version 2.0.
         # This unit test should reflect that change beginning 2.0,
         # and check that the project configuration is immutable.
+        assert _MAJOR_VERSION < _VERSION_3
```
vyasr, line 118:
```suggestion
        assert _MAJOR_VERSION < _VERSION_2
```

signac/contrib/project.py
```diff
@@ -1474,7 +1470,7 @@ def temporary_project(self, dir=None):
             yield tmp_project
 
     @classmethod
-    def init_project(cls, *args, root=None, workspace=None, make_dir=True, **kwargs):
+    def init_project(cls, *args, root=None, **kwargs):
```
bdice, line 1473: I still think the choice to postpone some breaking changes until 3.0 is arbitrary and should be done in 2.0 instead. At this point, we're changing enough behavior that I expect user scripts to break. We might as well change everything we want to change, rather than maintaining the large and awkward API shims like in this function.

signac/contrib/project.py
```diff
@@ -1565,25 +1555,15 @@ def init_project(cls, *args, root=None, workspace=None, make_dir=True, **kwargs)
...
+            _mkdir_p(os.path.dirname(fn_config))
             config = read_config_file(fn_config)
-            if workspace is not None:
-                config["workspace_dir"] = workspace
             config["schema_version"] = SCHEMA_VERSION
             config.write()
             project = cls.get_project(root=root)
             if name is not None:
                 project.doc[name_key] = name
             return project
```
bdice, line 1565: Both the `except` and the `else` have a `return project` at the end. I would move that return out of the `try/except/else` and remove the `else`.

bdice (approved):
One comment, one refactor, one suggestion for naming. Looks good, thank you!

signac/contrib/project.py
mikemhenry: That echos what we discussed in our dev meeting today, the more breakage we do now, the less we will have to break later/make a premature major version release when we could have set ourselves up better now.

vyasr: I don't think it's arbitrary, I think it's based on a reasonable guess of usage patterns based on empirical evidence, not to mention the fact that we're only going to be surfacing deprecation warnings for a short period of time because we're only now switching from DeprecationWarning to FutureWarning.

Having said that, I would be on board with just accepting that a major release involves breakage and not trying to maintain these patches. I agree that this is in line with what we discussed at the meeting. @csadorf what do you think? @mikemhenry could you add this as a topic for our next meeting?

csadorf: I agree with @vyasr that this is not arbitrary, but I also understand @bdice and @mikemhenry 's concern with respect to the overall pace of changes and this specific change. Version 1.0.0 of signac was released in Feb 2019, which means that we provided a very stable API for over **three years** now. That's a great accomplishment, but also comes with some responsibility to maintain that stability.

As a user of the _basic functions_* of a library that uses SemVer within the Python data ecosystem, I typically have the following expectations:

1. A major version change _will typically_ require a few high impact changes (adaptation effort on the order of hours) and will sometimes require substantial effort (weeks).
2. A minor version change _should not, but might_ require a few changes, but certainly with very low impact (minutes).
3. Any change required for a patch version upgrade is a bug/regression.

I would recommend that we move forward with all breaking changes for 2.0 without mitigation measures except where we estimate that the user effort for adaptation would be significantly more than 1 hour per project. For example, a breaking change to `init_project()` would be acceptable, because there are like only very few uses of this particular method for each project. A change to `.sp` or `.workspace` on the other hand would have a significantly larger impact.

In short: Let's move ahead with _all warranted_ breaking changes, but try to evaluate their potential impact and continue to support deprecated API with clear user instructions on how to upgrade as part of the warnings only for individual high impact changes.

*) Usually those shown as part of the quickstart examples and basic tutorials.

mikemhenry: RE: https://github.com/glotzerlab/signac/pull/714#discussion_r834971063
@cbkerr will be the community manager for the next meeting

bdice: @csadorf What would you recommend for an API break that requires more than one hour to complete? There isn't anything bigger than a major version. Users can pin to signac 1.x for as long as they want, if they aren't willing to upgrade. I think it's in the long-term interest of the project to be open to API adjustments. Both freud 2.0 and HOOMD 2.0 / 3.0 involved major API breaks. I just don't see a convincing argument for 2.0 to be released with deprecated code paths that won't be removed until some indeterminate future time (we have no guarantee that a signac 3.0 will _ever_ exist, based on the slowing pace of development and high stability of the package).

vyasr: Yeah I'm starting to come around to this point of view as well, especially given 

> we have no guarantee that a signac 3.0 will ever exist, based on the slowing pace of core feature development and high stability of the package

Breaking changes are what major releases are for. While I still don't think our decisions were arbitrary, I don't know that there's a good enough reason to avoid the breakage going forward. That said... let's discuss this at the next dev meeting. I think that's a more efficient medium for this kind of discussion than async Github threads.

csadorf: > @csadorf What would you recommend for an API break that requires more than one hour for users to adapt? There isn't anything bigger than a major version.

@bdice I think I was not fully successful in conveying the message I tried to send. A major breaking change can absolutely require users to make adaptations that require more than one hour of effort. There is really no limit here. When I wrote about "orders of hours to weeks", I was really just talking about my personal expectations, this does not really have much to do with SemVer _per se_.

My main point is that this is not a black and white issue. SemVer allows you to release basically a completely new package that has an entirely different API (**zero compatibility!**) with a new major version. Is that "legitimate"? Yes. Would you be better off just releasing a new package under a different name? Maybe.

What I am advocating for is to _make any breaking change that we want_. Only for individual high impact changes, let's have a close look at wether it is a) necessary and b) could be easily mitigated. If the answers are a) yes and b) no, move ahead with the breaking change.

vyasr:
This PR is awaiting resolution of the above discussion on deprecations, and then it should be good to go.

signac/contrib/project.py
vyasr: I'm going to resolve this conversation and make a follow-up PR removing backwards compatibility layers. If there are specific ones that anyone feels strongly that we should keep, they can be discussed in that PR.

PR #719, opened 2022-03-26
Remove non-str key support

bdice:
## Description
This PR closes a handful of TODO items related to automatically converting non-string keys to strings for synced collections.

## Motivation and Context
See #316 for the original motivation to remove non-string keys.

## Checklist:

### DIFF
--- a/signac/synced_collections/backends/collection_json.py
+++ b/signac/synced_collections/backends/collection_json.py
@@ -7,7 +7,6 @@ import errno
 import json
 import os
 import uuid
-import warnings
 from collections.abc import Mapping, Sequence
 from typing import Callable, FrozenSet
 from typing import Sequence as Sequence_t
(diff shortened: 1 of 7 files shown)


vyasr:
@bdice thanks for tackling this! Could you also rerun some of the benchmarks that we looked at in conjunction with #484 to see how much they improve with this applied? IIRC the switch to synced collections increased the cost of converting keys to strings, and that increase was exacerbated because I implemented it as a separate validator (with the expectation that it would be removed), so we should see a marked improvement now.

bdice:
@vyasr I looked at the benchmarks but there is no apparent change. I believe that this is because we created a single validator `json_attr_dict_validator` that does all the logic of `require_string_key` and `json_format_validator` in a single function. The branch I eliminated from that function was the `elif isinstance(key, (int, bool, type(None))):`. Since the first `if isinstance(key, str):` is always hit by the benchmarking code, I don't think this would result in any performance changes.

vyasr:
@bdice you're right, I forgot that when I implemented that new validator I bypassed the perf issue entirely. I think I initially recognized that the sequence of validators would be slower but wrote it off as a temporary issue until 2.0, but then when you did your much more thorough benchmarking of #484 and we saw how bad it was I just implemented the custom validator to recover performance. So I think we're good here.

vyasr (approved):

PR #752, opened 2022-05-02
Remove compatibility layers and other deprecated code

vyasr:
## Description

This PR removes the various backwards compatibility layers that have been introduced on next to mitigate the pain of migrating to signac 2.0. Some other deprecated code that has previously slipped through the cracks is removed as well.

## Motivation and Context

Based on discussions in some recent PRs (primarily #685 and #714) as well as more extended consideration in [our developer meeting](https://github.com/glotzerlab/signac/wiki/Developer-Meeting-Minutes#2022-04-05), we have decided against including backwards compatibility layers signac 2.0. Previously, we had incorporated some level of compatibility for changes that we foresaw having a particularly wide impact, but per our more recent discussions we will be moving forward with making these changes breaking in 2.0 since none of them are very difficult for users to fix.

## Checklist:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -29,7 +29,7 @@ Removed
  - The pre-SyncedCollection synchronized dictionary classes, including SyncedDict, SyncedAttrDict, and JSONDict (#577).
  - The old custom JSON encoder and dumps wrapper (#577).
  - The MPIPool and the filesystems.py module (#575).
- - The following Project methods: ``get_id``, ``build_job_search_index``, ``build_job_statepoint_index``, ``find_job_ids``, ``reset_statepoint``, ``update_statepoint``, ``create_access_module``, ``index``, ``dump_statepoints``, ``get_statepoint``, ``read_statepoints``, ``write_statepoints``, ``groupbydoc``, ``root_directory`` (#574, #593, #599, #601, #752).
+ - The following Project methods: ``get_id``, ``build_job_search_index``, ``build_job_statepoint_index``, ``find_job_ids``, ``reset_statepoint``, ``update_statepoint``, ``create_access_module``, ``index``, ``dump_statepoints``, ``get_statepoint``, ``read_statepoints``, ``write_statepoints``, ``groupbydoc``, ``root_directory``, ``num_jobs`` (#574, #593, #599, #601, #752).
  - The following Job methods: ``get_id``, ``workspace``, ``ws``, ``reset_statepoint`` (#578, #752).
  - The ``syncutil.copytree`` method (#581).
  - All Crawlers, including ``RegexFileCrawler``, ``MainCrawler``, ``MasterCrawler``, ``SignacProjectCrawler``, and ``BaseCrawler``, in addition to all associated functionality in indexing.py (#580).
@@ -40,11 +40,13 @@ Removed
  - The ability to pass indexes to various ``Project`` methods (#599).
  - The following ``JobsCursor`` methods: ``groupbydoc``, ``next`` (#601, #604).
  - The ``Project.config`` property is no longer mutable. Use the command line ``$ signac config`` to modify configuration (#608, #246, #244).
- - The following config related functions: ``get_config_file``, ``search_standard_dirs`` (#674).
+ - The following config related functions: ``get_config``, ``search_standard_dirs`` (#674).
  - ``Project`` subclasses can no longer define a ``Job`` subclass to use (#588, #693).
  - The ``Collection`` class (#664, #667, #683).
  - The ``project`` command line command (#752).
  - The ``workspace parameter to the command line ``job`` command (#752).
+ - The ability to call ``Project.workspace``, it is strictly a property now (#752).
+ - ``ProjectSchema.__call__``, ``ProjectSchema.detect`` (#752).
 
 Version 1
 =========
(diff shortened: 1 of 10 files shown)


changelog.txt
```diff
@@ -40,9 +40,13 @@ Removed
  - The ability to pass indexes to various ``Project`` methods (#599).
  - The following ``JobsCursor`` methods: ``groupbydoc``, ``next`` (#601, #604).
  - The ``Project.config`` property is no longer mutable. Use the command line ``$ signac config`` to modify configuration (#608, #246, #244).
- - The following config related functions: ``get_config_file``, ``search_standard_dirs`` (#674).
+ - The following config related functions: ``get_config``, ``search_standard_dirs`` (#674).
  - ``Project`` subclasses can no longer define a ``Job`` subclass to use (#588, #693).
- - The ``Collection`` class has been removed (#664, #667, #683).
+ - The ``Collection`` class (#664, #667, #683).
+ - The ``project`` command line command (#752).
+ - The ``workspace parameter to the command line ``job`` command (#752).
```
bdice, line 47:
```suggestion
 - The ``--workspace`` parameter to the CLI ``job`` subcommand (#752).
```

changelog.txt
```diff
@@ -40,9 +40,13 @@ Removed
  - The ability to pass indexes to various ``Project`` methods (#599).
  - The following ``JobsCursor`` methods: ``groupbydoc``, ``next`` (#601, #604).
  - The ``Project.config`` property is no longer mutable. Use the command line ``$ signac config`` to modify configuration (#608, #246, #244).
- - The following config related functions: ``get_config_file``, ``search_standard_dirs`` (#674).
+ - The following config related functions: ``get_config``, ``search_standard_dirs`` (#674).
  - ``Project`` subclasses can no longer define a ``Job`` subclass to use (#588, #693).
- - The ``Collection`` class has been removed (#664, #667, #683).
+ - The ``Collection`` class (#664, #667, #683).
+ - The ``project`` command line command (#752).
```
bdice, line 46: Sounds redundant. Maybe:
```suggestion
 - The ``project`` CLI subcommand (#752).
```

signac/contrib/project.py
```diff
@@ -51,36 +48,10 @@
...
 JOB_ID_REGEX = re.compile(f"[a-f0-9]{{{JOB_ID_LENGTH}}}")
 
-# The warning used for doc filter deprecation everywhere. Don't use
-# triple-quoted multi-line string to avoid inserting newlines.
-# (issue #725) TODO: In signac 2.0, remove all docstrings for doc_filter parameters. The
-# doc_filter parameters will only be preserved for backwards compatibility but
-# not advertised as part of the API in signac 2.0.
-DOC_FILTER_WARNING = (
-    "The doc_filter argument is deprecated as of version 1.7 and will be removed "
-    "in version 3.0. Users should instead use a filter with a 'doc.' prefix. For "
```
bdice, line 61: Needs a backport to `2.0` in the `master` branch.

signac/contrib/project.py
```diff
@@ -51,36 +48,10 @@
...
-    "example, `doc_filter={'foo': 'bar'}` is equivalent to `filter={'doc.foo': 'bar'}`. "
-    "See https://docs.signac.io/en/latest/query.html#query-namespaces for more "
-    "information."
-)
-
 # Temporary default for project names until they are removed entirely in signac 2.0
 _DEFAULT_PROJECT_NAME = None
```
bdice, line 52: Are we ready to remove this too?

signac/contrib/project.py
```diff
@@ -158,9 +129,7 @@ def __init__(self, root=None):
         # Prepare root directory and workspace paths.
         # os.path is used instead of pathlib.Path for performance.
         self._root_directory = os.path.abspath(root)
```
bdice, line 131: I'd rename the internal `self._root_directory` to `self._path`.

signac/contrib/project.py
```diff
@@ -1468,7 +1403,7 @@ def temporary_project(self, dir=None):
             yield tmp_project
 
     @classmethod
-    def init_project(cls, *args, root=None, **kwargs):
+    def init_project(cls, *, root=None):
```
bdice, line 1406: Do we want to keep the requirement that `root` is a keyword-only argument? All our previous discussions about compatibility layers are somewhat moot now, so I would be willing to revisit that.

It could even be named `path` for consistency with the new terminology. The more I think about it, it'd be really nice to call it `path`. :)

tests/test_project.py
```diff
@@ -763,14 +711,7 @@ def test_schema(self):
             "f.f2",
         ):
             assert k in s
-            if "." in k:
-                with pytest.warns(FutureWarning):
-                    assert k.split(".") in s
-            # The following calls should not error out.
             s[k]
```
bdice, line 714: I think the previous comment was somewhat useful, since it's weird to see a bare `__getitem__` outside of an assignment statement or function call.
```suggestion
            # The following call should not error out.
            s[k]
```

bdice (approved):
I have a few comments. This PR is somewhat large and is mostly self-contained, so the changes might be easier to handle in a second PR. This also needs documentation updates for the removed methods.

signac/contrib/project.py
vyasr: Good idea. I do like this. Let's just break everything now and be done with it 🙂

signac/contrib/project.py
vyasr: Good call. This is a tiny change so I'm just going to push it directly onto master.

signac/contrib/project.py
vyasr: Actually I'm going to make this an issue to address in a subsequent PR. This change will be large and will substantially bloat the scope of this PR. It will also require changing functions like `get_project`. We should figure out the full scope and address this in one fell swoop.

vyasr: I created #757 to follow up on this.

PR #754, opened 2022-05-06
Catch warnings on creation of deprecated password cache.

vyasr:
## Description

This PR prevents signac from emitting warnings when it creates a deprecated SimpleKeyring object on import because it is a module-level constant in the deprecated db module.

## Motivation and Context

We shouldn't be surfacing warnings to the user that are expected consequences of normal signac usage. This issue will disappear when we actually remove the deprecated code in 2.0

## Checklist:


bdice (approved):
Thanks. You have a simpler solution than I had been planning!

PR #758, opened 2022-05-10
Rename `root` parameter to `path`.

### LINKED ISSUE #757: Rename `root` parameter for all `Project` methods to `path`
For consistency with the new terminology that we're standardizing on for both `Job` and `Project`, we should rename the `root` parameters of `Project` methods to `path`. This change affects `get_project` and `init_project`, as well as the `Project` constructor. Since we're moving forward with the breaking changes in #752, w should also change the signature of `init_project` to not require `path` to be a keyword-only argument. There may be other related changes as well that we need to consider, but I don't think there is any way to construct e.g. a `Job` at a specific path (we always construct by `id` or `statepoint` so I can't think of anything else that I might be missing right now.

**Original comment**
Do we want to keep the requirement that `root` is a keyword-only argument? All our previous discussions about compatibility layers are somewhat moot now, so I would be willing to revisit that.

It could even be named `path` for consistency with the new terminology. The more I think about it, it'd be really nice to call it `path`. :)

_Originally posted by @bdice in https://github.com/glotzerlab/signac/pull/752#discussion_r863999527_

bdice:
## Description
See #757.

## Motivation and Context
Resolves #757.

## Checklist:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -22,6 +22,7 @@ Changed
    - Projects no longer have names and are identified solely by their root directories. This change also affects a number of public APIs where names are including, most prominently ``signac.init_project`` and ``Project.init_project``. Projects can now be constructed with just a root directory rather than a preloaded config (#677, #684, #706).
    - Project workspaces are no longer configurable, but are instead always defined as a subdirectory of the Project's root directory called ``workspace`` (#714).
  - Rather than searching upwards until the root, ``load_config`` will only load configuration files in the specified directory, which is assumed to be a project directory, as well as the user's home directory (#711).
+ - The ``signac.get_project`` and ``signac.init_project`` functions and corresponding ``Project`` methods accept the parameter ``path`` instead of ``root`` (#757, #758).
 
 Removed
 +++++++
--- a/changelog.txt
+++ b/changelog.txt
@@ -22,7 +22,7 @@ Changed
    - Projects no longer have names and are identified solely by their root directories. This change also affects a number of public APIs where names are including, most prominently ``signac.init_project`` and ``Project.init_project``. Projects can now be constructed with just a root directory rather than a preloaded config (#677, #684, #706).
    - Project workspaces are no longer configurable, but are instead always defined as a subdirectory of the Project's root directory called ``workspace`` (#714).
  - Rather than searching upwards until the root, ``load_config`` will only load configuration files in the specified directory, which is assumed to be a project directory, as well as the user's home directory (#711).
- - The ``signac.get_project`` and ``signac.init_project`` functions and corresponding ``Project`` methods accept the parameter ``path`` instead of ``root`` (#757, #758).
+ - Changed the ``root`` parameter to ``path`` in the ``signac.get_project`` and ``signac.init_project`` functions and corresponding ``Project`` methods (#757, #758).
 
 Removed
 +++++++
--- a/signac/common/config.py
+++ b/signac/common/config.py
@@ -24,7 +24,7 @@ def _get_project_config_fn(path):
 
 
 def _locate_config_dir(search_path):
-    """Locates root directory containing a signac configuration file in a directory hierarchy.
+    """Locates directory containing a signac configuration file in a directory hierarchy.
 
     Parameters
     ----------
@@ -34,17 +34,17 @@ def _locate_config_dir(search_path):
     Returns
     --------
     str or None
-        The root directory containing the configuration file if one is found, otherwise None.
+        The directory containing the configuration file if one is found, otherwise None.
     """
-    root = os.path.abspath(search_path)
+    search_path = os.path.abspath(search_path)
     while True:
-        if os.path.isfile(_get_project_config_fn(root)):
-            return root
-        if (up := os.path.dirname(root)) == root:
+        if os.path.isfile(_get_project_config_fn(search_path)):
+            return search_path
+        if (up := os.path.dirname(search_path)) == search_path:
             logger.debug("Reached filesystem root, no config found.")
             return None
         else:
-            root = up
+            search_path = up
 
 
 def read_config_file(filename):
--- a/signac/common/config.py
+++ b/signac/common/config.py
@@ -19,8 +19,8 @@ USER_CONFIG_FN = os.path.expanduser(os.path.join("~", ".signacrc"))
 # functions from the public API.
 
 
-def _get_project_config_fn(root):
-    return os.path.abspath(os.path.join(root, PROJECT_CONFIG_FN))
+def _get_project_config_fn(path):
+    return os.path.abspath(os.path.join(path, PROJECT_CONFIG_FN))
 
 
 def _locate_config_dir(search_path):
@@ -28,7 +28,7 @@ def _locate_config_dir(search_path):
 
     Parameters
     ----------
-    root : str
+    search_path : str
         Starting path to search.
 
     Returns
@@ -40,10 +40,7 @@ def _locate_config_dir(search_path):
     while True:
         if os.path.isfile(_get_project_config_fn(root)):
             return root
-        # TODO: Could use the walrus operator here when we completely drop
-        # Python 3.7 support if we like the operator.
-        up = os.path.dirname(root)
-        if up == root:
+        if (up := os.path.dirname(root)) == root:
             logger.debug("Reached filesystem root, no config found.")
             return None
         else:
@@ -81,12 +78,12 @@ def read_config_file(filename):
     return config
 
 
-def load_config(root=None):
+def load_config(path=None):
     """Load configuration from a project directory.
 
     Parameters
     ----------
-    root : str
+    path : str
         The project path to pull project-local configuration data from.
 
     Returns
@@ -96,8 +93,8 @@ def load_config(root=None):
         config data if requested. Note that because this config is a composite,
         modifications to the returned value will not be reflected in the files.
     """
-    if root is None:
-        root = os.getcwd()
+    if path is None:
+        path = os.getcwd()
     config = Config(configspec=cfg.split("\n"))
 
     # Add in any global or user config files. For now this only finds user-specific
@@ -106,8 +103,8 @@ def load_config(root=None):
         if os.path.isfile(fn):
             config.merge(read_config_file(fn))
 
-    if os.path.isfile(_get_project_config_fn(root)):
-        config.merge(read_config_file(_get_project_config_fn(root)))
+    if os.path.isfile(_get_project_config_fn(path)):
+        config.merge(read_config_file(_get_project_config_fn(path)))
     return config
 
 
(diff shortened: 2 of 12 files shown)


signac/common/config.py
```diff
@@ -40,10 +40,7 @@ def _locate_config_dir(search_path):
     while True:
         if os.path.isfile(_get_project_config_fn(root)):
             return root
-        # TODO: Could use the walrus operator here when we completely drop
-        # Python 3.7 support if we like the operator.
-        up = os.path.dirname(root)
-        if up == root:
+        if (up := os.path.dirname(root)) == root:
```
syjlee, line 43: are `root`s remaining here without being changed into search_path?

changelog.txt
```diff
@@ -22,6 +22,7 @@ Changed
    - Projects no longer have names and are identified solely by their root directories. This change also affects a number of public APIs where names are including, most prominently ``signac.init_project`` and ``Project.init_project``. Projects can now be constructed with just a root directory rather than a preloaded config (#677, #684, #706).
    - Project workspaces are no longer configurable, but are instead always defined as a subdirectory of the Project's root directory called ``workspace`` (#714).
  - Rather than searching upwards until the root, ``load_config`` will only load configuration files in the specified directory, which is assumed to be a project directory, as well as the user's home directory (#711).
+ - The ``signac.get_project`` and ``signac.init_project`` functions and corresponding ``Project`` methods accept the parameter ``path`` instead of ``root`` (#757, #758).
```
cbkerr, line 25: Moved the important change to the beginning
```suggestion
 - Changed the ``root`` parameter to ``path`` in the ``signac.get_project`` and ``signac.init_project`` functions and corresponding ``Project`` methods (#757, #758).
```

signac/contrib/project.py
```diff
@@ -78,13 +78,13 @@ class Project:
     A :class:`Project` may only be constructed in a directory that is already a
     signac project, i.e. a directory in which :func:`~signac.init_project` has
     already been run. If project discovery (searching upwards in the folder
-    hierarchy until a project root is discovered) is desired, users should
+    hierarchy until a project is discovered) is desired, users should
```
cbkerr, line 81: Can we say how a project is discovered? Like the presence of a `signac.rc` file?

signac/contrib/project.py
```diff
@@ -264,7 +264,7 @@ def min_len_unique_id(self):
         return i
 
     def fn(self, filename):
-        """Prepend a filename with the project's root directory path.
+        """Prepend a filename with the project's path.
```
cbkerr, line 267: "The project's path" makes it sound like the path is owned by the project, when actually you could copy paste a project directory and then the new path would be different.

(This may get into something deep about how projects are defined in v2.0 that I may have missed)

```suggestion
        """Prepend a filename with the path to the project.
```

signac/contrib/project.py
```diff
@@ -1370,7 +1370,7 @@ def _read_cache(self):
     def temporary_project(self, dir=None):
         """Context manager for the initialization of a temporary project.
 
-        The temporary project is by default created within the root project's
+        The temporary project is by default created within the parent project's
```
cbkerr, line 1373: Parent project doesn't sound like the same as "root" in this context. To me, "parent" sounds like a higher level then the project.

signac/contrib/project.py
```diff
@@ -1424,38 +1424,38 @@ def init_project(cls, *, root=None):
...
             The starting point to search for a project, defaults to the
             current working directory.
         search : bool
             If True, search for project configurations inside and above
-            the specified root directory, otherwise only return projects
-            with a root directory identical to the specified root argument
-            (Default value = True).
+            the specified directory, otherwise only return projects with
+            a directory whose path is identical to the specified path
+            argument (Default value = True).
```
cbkerr, line 1458: I would like to see the updated signac-docs page on what a project is now. See my other comment about whether a project "has" a path.

signac/contrib/project.py
```diff
@@ -2055,18 +2054,18 @@ def get_project(root=None, search=True, **kwargs):
...
     Parameters
     ----------
-    root : str
-        The job root directory.
-        If no root directory is given, the current working directory is
-        assumed to be within the current job workspace directory (Default value = None).
+    path : str
+        The job directory. If no path is given, the current working
+        directory is assumed to be within the current job workspace
+        directory (Default value = None).
```
cbkerr, line 2068: Critical distinction!



```suggestion
    path : str
        The job directory. If no path is given, the current working
        directory is assumed to be within the current project workspace
        directory (Default value = None).
```

(If you meant "job workspace directory", that is a term we are phasing out and this should instead be "job directory".)

tests/test_project.py
```diff
@@ -2242,9 +2242,9 @@ def test_get_job_subdir(self):
 
     def test_get_job_nested_project_subdir(self):
         # Test case: Get a job from a sub-directory of the job workspace dir
-        # when the job workspace is also a project root dir
-        root = self._tmp_dir.name
-        project = signac.init_project(root=root)
+        # when the job workspace is also a project dir.
```
cbkerr, line 2245: Critical
```suggestion
        # when the job directory is also a project directory
```

cbkerr (changes requested):
There's still a bit of old terminology (like "job workspace") that needs to be fixed.

signac/contrib/project.py
bdice: I think is supposed to say "job directory."
```suggestion
    path : str
        The job directory. If no path is given, the current working
        directory is assumed to be within the current job directory
        (Default value = None).
```

signac/contrib/project.py
bdice: How about:
```suggestion
        """Prepend a filename with the project path.
```

signac/contrib/project.py
bdice: Well... it is sort of a parent. The temporary project is created in a subdirectory of the project instance calling this. This is awkward almost any way I can think of writing it, since it's hard to explain the benefit of this implementation detail.

signac/contrib/project.py
bdice: I would need to consult the implementation to refresh myself but I think it's defined by the existence of a `.signac/config` file (?). Either way, it may be better to write this definition in one place (not in every method/function docstring), and it's outside the scope of this PR which is focused on an API change.

signac/common/config.py
bdice: Good idea. I refactored this in ba109a2ba371079ac36eaf21b3fb7c189abe5090.

signac/contrib/project.py
bdice: If you were to copy-paste a project, the copied project's path is different. I don't really understand this distinction.

bdice: Refactored in de6d897d99ae5865e865b05df04c7f5f16c1b9c7.

signac/contrib/project.py
cbkerr: Agreed https://github.com/orgs/glotzerlab/projects/5#card-81869156

signac/contrib/project.py
cbkerr: Proposal to also rewrite the beginning of the docstring to be:
```python
...
"""Context manager that initializes a temporary project within the calling Project.

The temporary project path defaults to that of the calling project workspace.

...

"""
```

btw - we should rename the argument from `dir` to `path`.

signac/contrib/project.py
cbkerr: I agree with the change.

signac/contrib/project.py
cbkerr: I misread the text. I made a commit with a rewording.

signac/contrib/job.py
```diff
@@ -828,13 +828,13 @@ def fn(self, filename):
         Returns
         -------
         str
-            The full workspace path of the file.
+            The absolute path of the file.
```
cbkerr, line 831:
```suggestion
            The absolute path to the file.
```

signac/contrib/job.py
```diff
@@ -828,13 +828,13 @@ def fn(self, filename):
...
 
         """
         return os.path.join(self.path, filename)
 
     def isfile(self, filename):
-        """Return True if file exists in the job's workspace.
+        """Check if a filename exists in the job path.
```
cbkerr, line 837: Do files exist "in" paths?

Perhaps the formal way would be "Check if a file exists in the directory at the job path", but that's long.

And it might just be simpler to stick with one word (path)

```suggestion
        """Check if a filename exists in the job directory.
```

signac/contrib/job.py
```diff
@@ -844,7 +844,7 @@ def isfile(self, filename):
         Returns
         -------
         bool
-            True if file with filename exists in workspace.
+            True if filename exists in the job path.
```
cbkerr, line 847:
```suggestion
            True if filename exists in the job directory.
```

cbkerr (changes requested):
Check 2 rewordings in this commit: 74d7f788
plus this comment: https://github.com/glotzerlab/signac/pull/758#discussion_r871890853

signac/contrib/job.py
bdice: @cbkerr Please feel free to apply changes to definitions / grammar in separate PRs, but this is far from the intended scope of this PR's focus on a singular API change. I tried to clean up a few egregious issues in the docs that seemed related to "root", but I am not able to invest the time needed to fully revamp the docs nor do I think it should be done in this PR.

bdice: If you're okay with limiting the scope here, you may apply the suggestions you've already made on this PR but please keep it limited. I'm only able to commit a small amount of time to the project and multiple iterations on things like this make it hard to keep PRs small and focused.

signac/contrib/project.py
bdice: This PR is focused on changing `root` to `path`. A follow-up PR to address this would be appropriate.

bdice:
I'm not sure if https://github.com/glotzerlab/signac/commit/74d7f788bd2af95f31b001c6518e39251df54e3a is quite accurate. If `signac.get_job()` is called from a subdirectory of a job directory, e.g. `my_project/workspace/4abbdfd8bbd68bb88e726bb8978ffa14/asdf/`, `signac.get_job()` will return the job from `4abbdf...`.

Perhaps it should say something like:
```The path from which to search. The first job found in or above this path is returned.```

cbkerr (approved):
Thanks for bearing with me!
Notes for future PRs:
1. [Functionality and description of `Project.temporary_project`](https://github.com/glotzerlab/signac/pull/758#discussion_r871890853)
2. Review docstrings for consistent terminology.

syjlee (approved):
Looks good to me. It reminds me of where's waldo

PR #761, opened 2022-05-19
Add links to the query api page in view and find --help

cbkerr:
## Description

Adds plaintext links to the output of `signac view --help` and `signac find --help` 

## Motivation and Context

I and others often can't find the query api page from the docs about finding and viewing.

Tested by building locally.

## Checklist:

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,4 +1,3 @@
-argparse-formatter>=1.4
 sphinx>=4.0.0
 sphinx_rtd_theme>=1.0.0
 sphinxcontrib-programoutput
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,3 +1,4 @@
+argparse-formatter>=1.4
 sphinx>=4.0.0
 sphinx_rtd_theme>=1.0.0
 sphinxcontrib-programoutput
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -17,6 +17,7 @@ import platform
 import re
 import shutil
 import sys
+from textwrap import dedent
 import warnings
 from pprint import pformat, pprint
 from rlcompleter import Completer
@@ -1468,11 +1469,17 @@ def main():
 
     parser_find = subparsers.add_parser(
         "find",
-        description="""All filter arguments may be provided either directly in JSON
-                       encoding or in a simplified form using the query API. See
-                       https://docs.signac.io/en/latest/query.html#simplified-syntax-on-the-command-line.
-                       $ signac find a 42
-                       is equivalent to $ signac find '{"a": 42}'.""",  # noqa:E501
+        formatter_class = argparse.RawDescriptionHelpFormatter,
+        description=dedent("""\
+        All filter arguments may be provided either directly in JSON
+        encoding or in a simplified form using the query API. For
+        example, the following two bash commands are equivalent:
+
+        $ signac find key 42
+        $ signac find '{"a": 42}'
+
+        For the full syntax, see:
+        https://docs.signac.io/en/latest/query.html#simplified-filter"""),  # noqa:E501
     )
     parser_find.add_argument(
         "filter",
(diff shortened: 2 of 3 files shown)


signac/__main__.py
```diff
@@ -1469,8 +1469,10 @@ def main():
     parser_find = subparsers.add_parser(
         "find",
         description="""All filter arguments may be provided either directly in JSON
```
bdice, line 1471: I think it could be helpful to use some intentional line breaks in this message. I looked at `man` pages for a few executables and commands like `apt search --help`, `docker run --help`. I'd recommend the following:

```
Filter arguments may be provided directly as JSON or in a simplified form using the query API. For, example the following commands are equivalent:

$ signac find key 42
$ signac find '{"a": 42}'

For more information, see: https://docs.signac.io/en/latest/query.html#simplified-filter
```
Note that we have an anchor in the source that can be used to shorten this URL: https://github.com/glotzerlab/signac-docs/blame/ddd4a06b23f82b722303d31b3d1eb7338a2fcd97/docs/source/query.rst#L251

I'd do something similar for the other case, too. This is helpful, thanks!

cbkerr: Good idea. The included `argparse.RawDescriptionHelpFormatter` and `argparse.RawTextHelpFormatter` don't allow wrapping of some text while preserving blank lines. There is a [long running feature request for a better formatter for argparse](https://github.com/python/cpython/issues/57015).

I propose adding a new dependency [argparse-formatter](https://pypi.org/project/argparse-formatter/) and its `FlexiFormatter`, which has been [proposed for inclusion in Python](https://github.com/python/cpython/pull/22129).

See the [source of the formatter](https://github.com/davesteele/argparse_formatter) for docs.

doc/requirements.txt
```diff
@@ -1,3 +1,4 @@
+argparse-formatter>=1.4
```
bdice, line 1: These are requirements only needed for building the docs. The package requirements will already be met by installing signac, so this can be removed.

signac/__main__.py
```diff
@@ -1468,9 +1469,17 @@ def main():
 
     parser_find = subparsers.add_parser(
         "find",
-        description="""All filter arguments may be provided either directly in JSON
-                       encoding or in a simplified form, e.g., -- $ signac find a 42 --
-                       is equivalent to -- $ signac find '{"a": 42}'.""",
+        formatter_class=FlexiFormatter,
```
bdice, line 1472: Can we just write the raw string we want to print? I don’t think adding a dependency is the right course of action if a literal expression of the desired string (with continuation symbols and explicit \n newlines) is viable.

cbkerr: I don't think so.

I tried
```
description=r""" blah blah with\
trailing lines and \n etc"""``` and it gets line wrapped.

bdice: This formatter is in the standard library and lets you manually write the string: https://docs.python.org/3/library/argparse.html#argparse.RawDescriptionHelpFormatter

I pushed a proposed change for this in b2efd854b40a2a7c44b7d8b89fd704582492b228.

bdice:
@cbkerr For some context: you may have observed from making this PR that `argparse` has some undesirable and awkward nuances in its design, and I think many Python developers feel the same way about it. However, it's the only real option we have in the standard library (forgoing the deprecated `optparse`). The most popular package beyond the standard library for CLI design is [click](https://click.palletsprojects.com/en/8.1.x/). Previous efforts to rewrite signac's CLI using click stalled because of some nuances in the signac CLI that click didn't support. See conversations on #309. We might be able to adopt `click` in signac 2.0, if someone is passionate about doing that, but it has not been a high priority task in the past.

signac/__main__.py
cbkerr: That works, thanks! I'd also rather not add a dependency but it seemed like it was about to get merged into Python.

I tried RawDescriptionHelpFormatter first but I didn't know about the Python automatic string continuation to have it wrap lines while preserving newlines.

bdice (approved):

PR #762, opened 2022-05-23
Refactor and simplify code printing to stderr.

bdice:
## Description
Refactors and simplifies code printing to `sys.stderr`. This eliminates some helper functions and reduces imports.

## Motivation and Context
Found opportunity for simplification while looking at internal documentation. No changelog entry is needed because this does not touch public APIs.

## Checklist:

### DIFF
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -34,7 +34,7 @@ from .common import config
 from .common.configobj import Section, flatten_errors
 from .contrib.filterparse import _add_prefix, parse_filter_arg
 from .contrib.import_export import _SchemaPathEvaluationError, export_jobs
-from .contrib.utility import _add_verbosity_argument, _query_yes_no
+from .contrib.utility import _add_verbosity_argument, _print_err, _query_yes_no
 from .diff import diff_jobs
 from .errors import (
     DestinationExistsError,
@@ -95,10 +95,6 @@ Synchronize your project with the temporary project, for example with:
 SHELL_HISTORY_FN = os.sep.join((".signac", "shell_history"))
 
 warnings.simplefilter("default")
-
-
-def _print_err(msg=None, *args):
-    print(msg, *args, file=sys.stderr)
 
 
 def _fmt_bytes(nbytes, suffix="B"):
@@ -1664,7 +1660,6 @@ job documents."
     try:
         args.func(args)
     except KeyboardInterrupt:
-        _print_err()
         _print_err("Interrupted.")
         if args.debug:
             raise
--- a/tests/test_find_command_line_interface.py
+++ b/tests/test_find_command_line_interface.py
@@ -2,9 +2,7 @@
 # All rights reserved.
 # This software is licensed under the BSD 3-Clause License.
 import json
-import os
-import sys
-from contextlib import contextmanager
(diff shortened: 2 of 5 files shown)


signac/__main__.py
```diff
@@ -97,10 +97,6 @@
 warnings.simplefilter("default")
 
 
-def _print_err(msg=None, *args):
```
iblanco11981870, line 100: Does it matter that the _print_err is now doing (msg, file=sys.stderr, *args) instead of the original position given?  (more an aesthetic question rather than functionality)

tests/test_find_command_line_interface.py
```diff
@@ -64,24 +62,11 @@
...
-    finally:
-        sys.stderr = old_target
-
-
 class TestFindCommandLineInterface:
     @staticmethod
     def _parse(args):
-        with open(os.devnull, "w") as null:
-            return parse_filter_arg(args, file=null)
+        with redirect_stderr(StringIO()):
```
iblanco11981870, line 68: This is more of a question for understanding- I was looking at this function, and it seems that it has created errors to people before (https://stackoverflow.com/questions/38291744/redirect-stderr-does-not-work-python-3-5). I was wondering if this is resolved because it is a @staticmethod or whether we should accommodate for this.

iblanco11981870 (approved):
Overall looks great, the only questions that I added are for 1) my own understanding/ 2) verifying that the stack overflow problem does not apply to us.

signac/__main__.py
bdice: The function is now defined as:
```python
def _print_err(*args, **kwargs):
    print(*args, file=sys.stderr, **kwargs)
```
This use of `*args` and `**kwargs` will transparently pass through all positional and keyword arguments, with one exception: we always provide `file=sys.stderr`. If the caller provides a keyword argument for `file`, the function will raise an error. This is okay because we're only using it as a shorthand utility for printing to `stderr`.

The previous function definition did not support other keyword arguments like `print("some", "words", end="")`, which provides an empty `end` value (rather than including a new line at the end of the output). The new definition corrects this behavior, which is needed for this call:
https://github.com/glotzerlab/signac/blob/8303d2e579b2b39621aa8a8cb35bc743a496b0b1/signac/contrib/migration/__init__.py#L113-L116

cbkerr (approved):
Looks good!

tests/test_find_command_line_interface.py
bdice: That StackOverflow question is misusing `redirect_stderr`. See the answer below: "You need to actually write to stderr, it is not a tool to catch exceptions."

The use of a `@staticmethod` here is to indicate that the method implementation is not dependent on the current instance `self` (as in a "normal" method) or the class `cls` (as in a `@classmethod`). A `@staticmethod` is equivalent to a free function (defined outside of a class) but has the benefit of being "scoped" through its definition in the class. To be honest, I rarely use static methods in Python because they are kind of awkward. Some developers argue that you should always use a free function (putting `def _parse` outside the class) but I didn't want to change this part of the test. See also: https://stackoverflow.com/questions/136097/difference-between-staticmethod-and-classmethod

PR #763, opened 2022-05-23
Add "optional" to parameter docstrings.

### LINKED ISSUE #344: NumPy docstyle should denote "optional" arguments

bdice:
## Description
Resolves #344. Note: I am going to open a couple more PRs with related docstring revisions that may require more thoughtful review. This PR is a subset that should be non-controversial and easy to approve.

## Checklist:

### DIFF
--- a/signac/contrib/_searchindexer.py
+++ b/signac/contrib/_searchindexer.py
@@ -380,7 +380,7 @@ class _SearchIndexer(dict):
         Parameters
         ----------
         expr : dict
-            The filter of expressions to match (Default value = None).
+            The filter of expressions to match.
 
         Returns
         -------
--- a/signac/contrib/_searchindexer.py
+++ b/signac/contrib/_searchindexer.py
@@ -98,7 +98,7 @@ class _TypedSetDefaultDict(dict):
         ----------
         key : str, float
             The key to get the value.
-        default :
+        default : object, optional
             Default value if type of key is not float (Default value = None).
 
         Returns
@@ -471,7 +471,7 @@ class _SearchIndexer(dict):
 
         Parameters
         ----------
-        filter_ : dict
+        filter_ : dict, optional
             The filter of expressions to match (Default value = None).
 
         Returns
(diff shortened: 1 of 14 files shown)


iblanco11981870 (approved):
The only changes performed that I saw were: 1) adding on the comments the word optional and 2) instead of returning the name of the job as a string, we are now passing the id of the job. (line 1830) for def keyfunction. Overall, these changes did not change any functionality of the program.

bdice:
Thanks for your reviews @shihkual @iblanco11981870! I'm going to merge this now.

PR #764, opened 2022-05-23
Always encode nested dicts.

bdice:
## Description
This PR refactors a utility function `_nested_dicts_to_dotted_keys` to remove a default parameter that is never used by any callers. This simplifies the logic and removes unnecessary functionality. Because this is an internal change, no changelog entry is needed.

## Checklist:

### DIFF
--- a/signac/contrib/utility.py
+++ b/signac/contrib/utility.py
@@ -213,25 +213,6 @@ def _to_hashable(obj):
         return obj
 
 
-def _encode_tree(x):
-    """Encode if type of x is list.
-
-    Parameters
-    ----------
-    x :
-        type to encode.
-
-    Returns
-    -------
-    Hashable version of ``x``.
-
-    """
-    if type(x) is list:
-        return _to_hashable(x)
-    else:
-        return x
-
-
 def _nested_dicts_to_dotted_keys(d, key=None):
     """Generate tuples of key in dotted string format and value from nested dict.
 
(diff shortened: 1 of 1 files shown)


iblanco11981870 (approved):
Changes reviewed: a) changing variables to make it easier to read/ more intuitive . b) always encode the nested dicts by removal of the "if is not None" statement, which previously did not allow the an empty/non existing encode value to be encoded.

signac/contrib/utility.py
```diff
@@ -252,14 +250,13 @@ def _nested_dicts_to_dotted_keys(t, encode=_encode_tree, key=None):
         Tuples of dotted key and values e.g. ('a.b', 'c')
 
     """
-    if encode is not None:
-        t = encode(t)
-    if isinstance(t, Mapping):
-        if t:
-            for k in t:
+    d = _encode_tree(d)
```
kidrahahjo, line 253: As mentioned in the docs, we're always expecting `d` to be a `dict`. Hence `_encode_tree` will simply return the non-encoded object, do we even need to use this then?

bdice: The docs are not entirely clear here. We expect callers of this function to provide `d` as a dict. However, this function is called recursively, and it may provide non-dict inputs to its recursive calls. See:

```python
yield from _nested_dicts_to_dotted_keys(d[k], key=k_)
```

However, this points out to me that we may only need to call `_encode_tree` in the `else:` case below, to handle lists. And moreover, it could be written as an inline function instead of existing as a separate `_encode_tree`. I'll refactor this a bit.

bdice: Updated in 39165c14.

kidrahahjo (approved):
Build fails on windows testing but I guess it's an issue with CircleCI. I approve of these changes.

bdice:
Thanks @kidrahahjo! We’re ignoring the Windows CI results for now.

PR #767, opened 2022-05-23
Fix copytree docs and unify default parameter logic.

bdice:
## Description
While working on #763, I unified the `copytree` docs and default parameter behavior across the codebase. Now, the default parameter in public APIs should always be `None` and we should consistently document that `None` means that signac will use `shutil.copytree`.

I also updated a few docstrings related to synchronization.

## Checklist:

### DIFF
--- a/signac/contrib/import_export.py
+++ b/signac/contrib/import_export.py
@@ -474,10 +474,10 @@ def export_jobs(jobs, target, path=None, copytree=None):
         A path to a directory or archive file to export to.
     path : str or callable
         The path (function) used to structure the exported data space. (Default value = None)
-    copytree : callable
-        The function used for copying of directory tree
-        structures. Defaults to :func:`shutil.copytree`.
-        Can only be used when the target is a directory.
+    copytree : callable, optional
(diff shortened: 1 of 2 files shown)


csadorf (approved):

bdice:
> addition of an if statement that changes the empty input, labeled as known from the previous change, to shutil.copytree (question: why were these two changes necessary? It seems like we broke up one line into 3 lines)

@iblanco11981870 Good question. I think we had decided at some point in the past that we didn't like using the function `shutil.copytree` directly as a default parameter, and instead preferred to let it be `None` and replace `None` with that default value in the function. I don't know the rationale for that choice anymore. This PR's primary goal was to establish a consistent style for all the functions accepting a `copytree`-like function. We could have gone the other way, and used `shutil.copytree` instead of `None` in all cases, but I went with using `None` because I think it gives us the most flexibility to change the internals going forward without altering the signac public API. For example, we could replace the internal use of `shutil.copytree` with some better/faster copy function while retaining the default value of `None`. This is unlikely, but it's best to be defensive with preventing future breakage of public APIs.

PR #773, opened 2022-06-01
Use GitHub Actions.

bdice:
## Description
This PR overhauls our CI configurations to build and test signac with GitHub Actions. We've been planning to move to GitHub Actions for a long time and the breakage of our Windows builds (limited monthly minutes) and limited parallel runners are finally motivating a change here.

Before merging:
- [ ] Ensure we have builds for all supported platforms (Linux, macOS, Windows) and configurations (Python versions, oldest/newest dependencies). I'd like to see the following:
  - [ ] Linux, Python 3.10
  - [ ] Linux, Python 3.10 "minimal" (no optional dependencies)
  - [ ] Linux, Python 3.9
  - [ ] Linux, Python 3.8
  - [ ] Linux, Python 3.8 "oldest" (minimum versions of supported dependencies)
  - [ ] macOS, some supported Python version
  - [ ] Windows, some supported Python version
- [ ] Make an attempt at the "deploy to PyPI" logic that should release a source distribution (tarball) and a wheel tagged `py3-none-any` uploaded from one of those test configurations mentioned above. See https://github.com/pypa/gh-action-pypi-publish

## Motivation and Context
I don't think an issue exists for this but we've been discussing moving to GitHub Actions for a while. cibuildwheel is a convenient and well-tested / easily configurable way to do so.

## Checklist:
- [x] I am familiar with the [Contributing Guidelines](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [Contributor Agreement](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
- [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.
- [ ] The [package documentation](https://github.com/glotzerlab/signac/tree/master/doc) and [framework documentation](https://docs.signac.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.
- [ ] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added any related issue and pull request numbers for future reference.

PR #776, opened 2022-06-19
Use GitHub Actions for CI

bdice:
## Description
This PR overhauls our CI configurations to build and test signac with GitHub Actions. We've been planning to move to GitHub Actions for a long time and the breakage of our Windows builds (limited monthly minutes) and limited parallel runners are finally motivating a change here.

Depends on #777.

I'm going to leave CircleCI active (except for the failing Windows jobs) for a little while to ensure a smooth transition.

Before merging:
- [x] Ensure we have builds for all supported platforms (Linux, macOS, Windows) and configurations (Python versions, oldest/newest dependencies). I have the following:
  - [x] Linux, Python 3.10
  - [x] Linux, Python 3.10 "minimal" (no optional dependencies)
  - [x] Linux, Python 3.9
  - [x] Linux, Python 3.8
  - [x] Linux, Python 3.8 "oldest" (minimum versions of supported dependencies)
  - [x] macOS, some supported Python version
  - [x] Windows, some supported Python version
  - ❌ PyPy. I don't really think we need to keep testing PyPy. If another maintainer would like to add that to the CI, feel free.
  - ❌ Benchmarks. I haven't attempted to port these over from CircleCI. We deleted the CI script for benchmarks from signac 2.0 anyway, because it's not reliable and we want to use `asv` instead.
- ❌ Make an attempt at the "deploy to PyPI" logic that should release a source distribution (tarball) and a wheel tagged `py3-none-any` uploaded from one of those test configurations mentioned above. See https://github.com/pypa/gh-action-pypi-publish
  - I'm going to revisit this at a later time using the wonderful template from @matthewfeickert linked below.
  - We will need to port the `check-metadata` job and the `deploy.bash` script.

## Motivation and Context
I don't think an issue exists for this but we've been discussing moving to GitHub Actions for a while. Supersedes previous attempt in #773.

## Checklist:

### DIFF
--- a/.github/workflows/run-pytest.yml
+++ b/.github/workflows/run-pytest.yml
@@ -24,6 +24,13 @@ jobs:
                   {python: '3.10', dependencies: 'newest'},
                   {python: '3.10', dependencies: 'minimal'},
                   {python: '3.8', dependencies: 'oldest'} ]
+        exclude:
+          # macOS, Python 3.10 is excluded because zarr dependency numcodecs is
+          # not yet available as a wheel for that configuration and fails to
+          # build from source. See
+          # https://github.com/zarr-developers/numcodecs/issues/326
+          - os: macos-latest
+            config: {python: '3.10', dependencies: 'newest'}
 
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/run-pytest.yml
+++ b/.github/workflows/run-pytest.yml
@@ -1,0 +1,74 @@
+name: Run pytest
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-20.04, windows-2019, macos-10.15]
+        python-version: ["3.8", "3.9", "3.10"]
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v3
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install the package
+      run: |
+        pip install -e requirements/requirements-test.txt
+        pip install -e requirements/requirements-test-optional.txt
+        pip install -e .
+    - name: Run MongoDB
+      uses: supercharge/mongodb-github-action@1.7.0
+    - name: Run Redis
+      uses: supercharge/redis-github-action@1.4.0
+    - name: Test with pytest
+      run: |
+        pytest -n 2 --dist loadfile --cov=signac --cov-config=setup.cfg --cov-report=xml tests/ -v
+    - uses: codecov/codecov-action@v2
+
+  test-minimal:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.10"]
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v3
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install the package
+      run: |
+        pip install -e requirements/requirements-test.txt
+        pip install -e .
+    - name: Test with pytest
+      run: |
+        pytest -n 2 --dist loadfile --cov=signac --cov-config=setup.cfg --cov-report=xml tests/ -v
+    - uses: codecov/codecov-action@v2
+
+  test-oldest:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.8"]
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v3
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install the package
+      run: |
+        pip install -r .circleci/ci-oldest-reqs.txt
+        pip install -e .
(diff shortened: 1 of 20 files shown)


matthewfeickert:
> * Make an attempt at the "deploy to PyPI" logic that should release a source distribution (tarball) and a wheel tagged `py3-none-any` uploaded from one of those test configurations mentioned above. See https://github.com/pypa/gh-action-pypi-publish

Would highly recommend taking advantage of the idea that GHA supports many workflows and make this a separate workflow that is triggered either on a release and/or `workflow_dispatch`. c.f. [`pyhf`'s worfkflow for an example](https://github.com/scikit-hep/pyhf/blob/86229ad3257ddbbdb78548d7ceade5c9ab683a0c/.github/workflows/publish-package.yml).

bdice:
I confirmed that the number of tests that pass/skip is the same between CircleCI and GitHub Actions for the Linux tests (newest, oldest, minimal dependencies). This confirms that we're correctly testing what we intend to test with respect to optional dependencies (MongoDB, Redis, etc.).

bdice:
Tagging @mikemhenry and @cbkerr for review since you've had some interest in CI in the past.

.github/workflows/run-pytest.yml
```diff
@@ -0,0 +1,66 @@
+name: Run Unit Tests
+
+on:
+  # trigger on pull requests
+  pull_request:
```
mikemhenry, line 5: Do we want this to run on all PRs or ones that target a given branch?

.github/workflows/run-pytest.yml
```diff
@@ -0,0 +1,66 @@
...
+
+  # trigger on all commits to master
+  push:
+    branches:
+      - 'master'
```
mikemhenry, line 10:
```suggestion
      - 'master'
      - 'main'
```
If/when we migrate from master to main as our default branch, this will help future-proof it (but we will have to update all the badges so as long as you remember to update the workflow as well, this is less useful)

joaander:
> > * Make an attempt at the "deploy to PyPI" logic that should release a source distribution (tarball) and a wheel tagged `py3-none-any` uploaded from one of those test configurations mentioned above. See https://github.com/pypa/gh-action-pypi-publish
> 
> @bdice Would highly recommend taking advantage of the idea that GHA supports many workflows and make this a separate workflow that is triggered either on a release and/or `workflow_dispatch`. c.f. [`pyhf`'s worfkflow for an example](https://github.com/scikit-hep/pyhf/blob/86229ad3257ddbbdb78548d7ceade5c9ab683a0c/.github/workflows/publish-package.yml).

This is what `gsd` does: https://github.com/glotzerlab/gsd/blob/ca4f7c133594dd88f46d33b6e99c34a31b5706e8/.github/workflows/build_wheels.yml#L65-L96

.github/workflows/run-pytest.yml
```diff
@@ -0,0 +1,66 @@
...
+          - os: macos-latest
+            config: {python: '3.10', dependencies: 'newest'}
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python ${{ matrix.config.python }}
+      uses: actions/setup-python@v3
+      with:
+        python-version: ${{ matrix.config.python }}
+    - name: Install newest dependencies
```
joaander, line 41: You can use `actions/cache` to cache the files pip downloads and reduce the CI run time slightly: https://github.com/actions/cache

.github/workflows/run-pytest.yml
```diff
@@ -0,0 +1,66 @@
...
+  pull_request:
+
+  # trigger on all commits to master
+  push:
+    branches:
+      - 'master'
+
+  # trigger on request
+  workflow_dispatch:
+
```
mikemhenry, line 14:
```suggestion
concurrency:
  group: "${{ github.workflow }}-${{ github.ref }}"
  cancel-in-progress: true
```

Something like this ^  will kill the workflow when another commit is made, instead of running the workflow twice. https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow
I like it since it helps reduce CI waste when someone makes a quick commit after pushing to fix some minor thing.

mikemhenry (approved):
Just a few notes!

cbkerr (approved):
I only had time for a quick look, but I didn't see any obvious problems.
Thanks!

.github/workflows/run-pytest.yml
bdice: We'll be doing a mass find-replace when we make that branch transition, so I'm going to hold off until then.

.github/workflows/run-pytest.yml
bdice: I think this is a good idea but I'm going to leave it out-of-scope for this PR. Looks like this explains how to set it up for pip caching: https://github.com/actions/cache/blob/main/examples.md#simple-example

.github/workflows/run-pytest.yml
bdice: All PRs. 🤷‍♂️

.github/workflows/run-pytest.yml
bdice: Caching might not work well with cross-platform CI (on Windows in particular). 🤔

joaander: Caching works fine on multi-platform CI. You include the platform and any other relevant system specific quantities (like python version) in the cache key.

Your "install newest" dependencies step takes ~3 m 18 s (out of 10m 56 s total test time). Caching would remove the download times from that 3m 18s, but not the install times.

Issue #779, opened 2022-06-21
get_project 'root' keyword fails silently


rskye144:
### Description
When the `get_project` function is called with a project root that does not exist, it silently defaults to a project in the current directory. 

### To reproduce

```python
import signac
project1 = signac.init_project()
project2 = signac.get_project("some/directory/that/does/not/exist")  # this should raise to prevent user error, rather than searching upward from a nonexistent directory
assert project1 == project2  # this is true but we shouldn't get this far
```

Error also occurs because paths with `~` are not expanded, producing an unexpected error; @bdice suggests could consider adding:

```python
if "~" in root:
    warnings.warn("The character '~' is in the provided root path. Did you mean to call os.path.expanduser(root)?", UserWarning)
```

### Error output

No error is raised (which is wrong, an error/warning should be raised)

mikemhenry:
@rskye144 :wave: thanks for the bug report :heart_eyes: 

Are you interested in contributing a PR to fix this issue? We can provide some help/guidance if you are interested!

rskye144:
I can give a PR a try! It will be my first time doing that, so any advice is appreciated.

bdice:
One key piece that I didn't notice earlier: this bug occurs fails for _relative paths contained inside the current directory_, when the current directory contains a signac project. Getting a nonexistent absolute path like `signac.get_project("/a/fake/path")` fails with a `LookupError`.

vyasr:
Resolved by #792

PR #792, opened 2022-07-27
Issue #779: get_project() root keyword fails silently

rskye144:
When loading a nonexistent project with get_project(root="/does/not/exist"), signac silently fails to find the root and defaults to the current directory. If there is a project already in that directory, this causes user error when a different project is loaded than is expected.

## Description
I added a line suggested by @bdice to the config.py load_config() function to check if root is a valid directory, and if not then it raises a ValueError.

I do not think this change should break functionality?

## Motivation and Context
This change fixes issue #779 

## Checklist:

- [X] I am familiar with the [Contributing Guidelines](https://github.com/glotzerlab/signac/blob/master/CONTRIBUTING.md).
- [X] I agree with the terms of the [Contributor Agreement](https://github.com/glotzerlab/signac/blob/master/ContributorAgreement.md).
- [X] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/master/contributors.yaml).
- [X] The changes introduced by this pull request are covered by existing or newly introduced tests.
- [X] I have updated the [changelog](https://github.com/glotzerlab/signac/blob/master/changelog.txt) and added any related issue and pull request numbers for future reference.


rskye144:
I'm giving this a start - several things still need to be done but I created this draft so I can ask for help. First issue: when trying to run tests, the test_project and test_shell tests raised errors, I believe from expected behavior (test accidentally activating the root = does/not/exist), but I'm not sure.  I added a `pytest.raises(ValueError)` but I'm unsure how to use that safely, so advice is appreciated.

changelog.txt
```diff
@@ -6,6 +6,11 @@ The **signac** package follows `semantic versioning <https://semver.org/>`_.
 
 Version 1
 =========
+next
+----
+
+- Fix issue get_project() "root" keyword failing silently
```
bdice, line 12: Can you move this into the section under `[1.8.0]` / `Fixed` and include the issue / PR numbers? 1.8.0 hasn't been released yet, so that's the release that will contain this fix.

signac/common/config.py
```diff
@@ -145,6 +147,9 @@ def load_config(root=None, local=False):
     """Load configuration, searching upward from a root path."""
     if root is None:
         root = os.getcwd()
+    if not os.path.isdir(root):
+        with pytest.raises(ValueError):
+            raise ValueError(f"The root directory '{root}' does not exist.")
```
bdice, line 152: The `with pytest.raises` piece should be written in a test for this function, but doesn't need to be used in the implementation.
```suggestion
    if not os.path.isdir(root):
        raise ValueError(f"The root directory '{root}' does not exist.")
```

For the test, you can look at [test_project.py](https://github.com/glotzerlab/signac/blob/master/tests/test_project.py) and add a couple lines above these ones: https://github.com/glotzerlab/signac/blob/4bf9bceacfc780de49a879e16a45382ae0a932c9/tests/test_project.py#L2286-L2287

I would add these lines to the `test_get_project` function:
```python
        with pytest.raises(ValueError):
            signac.get_project(root="/a/path/that/does/not/exist")
```

bdice (commented):
@rskye144 Thank you very much for the contribution! I have a couple suggestions -- please feel free to ask for clarification if anything is unclear. 👍

signac/common/config.py
rskye144: Thank you for the suggestions! I'll add them in. I want to make sure that the error currently being raised is actually as expected.  I'm getting 12 warnings, all in the new ValueError I set up. I'll copy an example below. As far as I can tell, the temporary directory set up for testing is registering as Not A Directory when tested, though in non-tests and non-temp directories the load_config() works as expected. Is it expected for the tempfile.TemproraryDirectory() to behave like this?

`root = '/var/folders/cl/2y21xs_55pdg2wg0rgf130qc0000gn/T/signac_u3cgrvrc/a', local = False

    def load_config(root=None, local=False):
        """Load configuration, searching upward from a root path."""
        if root is None:
            root = os.getcwd()
        if not os.path.isdir(root):
>           raise ValueError(f"The root directory '{root}' does not exist.")
E           ValueError: The root directory '/var/folders/cl/2y21xs_55pdg2wg0rgf130qc0000gn/T/signac_u3cgrvrc/a' does not exist.`

bdice: I'm not sure if I understand. Can you push your proposed changes to this branch, and I can try to reproduce locally?

rskye144: Yes, thank you! I've uploaded it; I hope it merges? Here's a screenshot of part of the output of "python -m pytest tests/" for extra clarity.

Basically, I think I could make it pass the tests by going through and specifying "raises(ValueError)" for all of these points, but I'm unsure that would actually catch issues as we want it to.
image

bdice:
@rskye144 I think I misunderstood the original problem. The fix might need to look more like this: https://github.com/glotzerlab/signac/compare/fix/issue-779?expand=1

The part I missed was that this issue occurred for _relative_ paths and not absolute paths. The default behavior for `get_project` is to search upward from the provided path until it finds a project directory (`search=True` by default). That makes it so that calling `get_project` from within a job directory will return the containing project. However, we never check that the path actually exists. Maybe we can discuss this with @glotzerlab/signac-committers and get a consensus on (1) whether this is a bug or expected behavior and (2) whether we should issue a warning (and search upwards and return the first found project anyway) or throw an error when the user provides a nonexistent relative path.

rskye144:
Thank you @bdice ! That's a good point. Yeah, I noticed the search=True default when I was looking at the code but didn't want to mess with that. I'll look through it some more and see what it seems like could be a fix.

I agree that changing a default behavior like that requires a wider discussion. What, currently, is the intended use of that behavior? Does it help when trying to find a project when inside its workspace, for example?

mikemhenry:
I think that we should modify the behavior such that we don't do any searching for projects when a path is provided to `get_project`. If someone does `get_project()` (which I think works) then it is clear they want the search behavior (IMHO), but if they provide a path, we should use that and raise an error if the path doesn't exist or a project doesn't exist at the path location.

Thanks @rskye144 for working on this! I appreciate that you have uncovered this ambiguous behavior.

rskye144:
Hello! Thanks for all the help. I've added in a fix like @bdice suggested and pushed that for you all to check out. It passes tests correctly now! I'll check out the errors Github checks generated later. Thanks for the input, @mikemhenry . 
I messed around a bit, and it seems to behave as expected for relative paths like "../other_directory". It also fails intuitively for paths with "~" by printing out an obviously weird root in the error message:
image

I still need to fix all the contributor guidelines, so a few more steps left.

signac/contrib/project.py
```diff
@@ -2387,14 +2387,20 @@ def get_project(cls, root=None, search=True, **kwargs):
...
         config = load_config(root=root, local=False)
         if "project" not in config or (
             not search
             and os.path.realpath(config["project_dir"]) != os.path.realpath(root)
         ):
             raise LookupError(
                 "Unable to determine project id for path '{}'.".format(
-                    os.path.abspath(root)
+                    f"Unable to determine project id for path '{os.path.abspath(root)}'."
                 )
```
bdice, line 2404: This is performing a string formatting twice -- you can use `"Error: {}".format(message)` or an f-string like `f"Error: {message}"` but not both.

rskye144:
Okay, I'm tentatively ready for solidifying this PR?  As a last check, I'm not sure the docs need to be updated - this should be consistent with the current get_project() description which reads:
`root ([str](https://docs.python.org/3/library/stdtypes.html#str)) – The starting point to search for a project, defaults to the current working directory.`

rskye144:
Hmmm, actually, reviewing what @mikemhenry commented: the behavior of this current branch is that if the root does not exist, then it fails. However, if the root *does* exist but doesn't contain a project, then signac succesfully searches upwards, which as Mike said we may not want to do. 

As he commented, would it make sense to add a check for ``if root is not None: search = False`` ? The problem there is I don't know how to differentiate a user-set ``search=True`` from the default value ``search=True``.

bdice:
> I think that we should modify the behavior

Behavior should not be modified in this PR, aside from raising an error if the provided path is nonexistent (which I consider to be a bug). If we change behavior, it will need to wait until version 2.0.

changelog.txt
```diff
@@ -7,6 +7,7 @@ The **signac** package follows `semantic versioning <https://semver.org/>`_.
 Version 1
 =========
 
+
```
bdice, line 10:
```suggestion
```

changelog.txt
```diff
@@ -47,6 +48,7 @@ Fixed
  - User-provided path functions now raise an error if not unique (#666).
  - ``Collection`` class no longer raises an error when searching by a primary key that does not exist (#676).
  - Relative paths on Windows are not used if the current directory has no common prefix (#777).
+ - ``get_project()`` now returns an error if provided with a project root that does not exist (#779, #792)
```
bdice, line 51:
```suggestion
 - ``get_project()`` now raises an error if provided a root directory that does not exist (#779, #792).
```

signac/common/config.py
```diff
@@ -143,9 +145,12 @@ def get_config(infile=None, configspec=None, *args, **kwargs):  # noqa: D103
 
 def load_config(root=None, local=False):
     """Load configuration, searching upward from a root path."""
+
     if root is None:
         root = os.getcwd()
+
     config = Config(configspec=cfg.split("\n"))
+
```
bdice, line 153: Reverting whitespace changes.
```suggestion
    if root is None:
        root = os.getcwd()
    config = Config(configspec=cfg.split("\n"))
```

bdice (approved):
Approving with a few small suggestions, which I will apply.

signac/common/config.py
```diff
@@ -7,6 +7,8 @@
 import os
 import stat
 
+import pytest
+
```
bdice, line 11: Removing unused import.
```suggestion
```

contributors.yaml
```diff
@@ -119,4 +119,9 @@ contributors:
     given-names: "Matthew"
     affiliation: "University of Wisconsin-Madison"
     orcid: "https://orcid.org/0000-0003-4124-7862"
+
```
bdice, line 122: Fix missing dash (denotes that you're a separate person from the one above you).
```suggestion
  -
```

bdice:
Nice @rskye144! Thank you for PR. Thanks to @mikemhenry for the review.

vyasr:
> > I think that we should modify the behavior [...]
> 
> Behavior should not be modified in this PR, aside from raising an error if the provided path is nonexistent (which I consider to be a bug). If we change behavior, it will need to wait until version 2.0.

@bdice @mikemhenry Just a quick note here that the changes in #711 should facilitate this change if we do want to modify the behavior for 2.0.

PR #795, opened 2022-08-01
Remove --doc-filter option from CLI.

### LINKED ISSUE #613: Remove --doc-filter from CLI
### Feature description

From #586 conversation: https://github.com/glotzerlab/signac/pull/586#issuecomment-904777526

> Yes, I think --doc-filter should be removed in all cases. The signac-flow PR is less urgent since it only needs to go in before whichever flow version is the first to support signac 2.0, but adding the warning for find --doc-filter in this PR makes sense.

I decided to tackle this in a follow-up PR rather than extend #586.

### Proposed solution

`signac find --doc-filter ...` should be deprecated in 2.0 and removed in 3.0. A warning should be shown in signac 2.0.

`--doc-filter` should be deprecated and removed from signac-flow's CLI (e.g., `project.py status --doc-filter ...`).

bdice:
## Description
This PR resolves #613 by removing the `--doc-filter` argument from the signac CLI. Users who wish to search by a document filter should replace syntax as shown below:

```bash
Before (simple syntax):
$ signac find --doc-filter a 0
After (simple syntax):
$ signac find doc.a 0

Before (JSON syntax):
$ signac find --doc-filter '{"a": 0}'
After (JSON syntax):
$ signac find '{"doc.a": 0}'
```

## Motivation and Context
The general feature of searching by document filters has been replaced by a generalized filter syntax that prefixes state point keys with `sp.` and document keys with `doc.`. This applies that change from #586 to the CLI.

## Checklist:

### DIFF
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -32,7 +32,7 @@ else:
 from . import get_project, init_project
 from .common import config
 from .common.configobj import Section, flatten_errors
-from .contrib.filterparse import _add_prefix, parse_filter_arg
+from .contrib.filterparse import parse_filter_arg
 from .contrib.import_export import _SchemaPathEvaluationError, export_jobs
 from .contrib.utility import _add_verbosity_argument, _print_err, _query_yes_no
 from .core.utility import _safe_relpath
@@ -135,7 +135,7 @@ def _open_job_by_id(project, job_id):
 
 def find_with_filter_or_none(args):
     """Return a filtered subset of jobs or None."""
-    if args.job_id or args.filter or args.doc_filter:
+    if args.job_id or args.filter:
         return find_with_filter(args)
     else:
         return None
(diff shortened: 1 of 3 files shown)


kidrahahjo (approved):
Is it worth adding tests for these changes? I am not sure if we already have them or not. Also, I see a lot of lines not covered by us. Although not in scope for this PR, should we worry about this?

The changes looks good to me, if we decide to add tests, happy to re-review :)

bdice:
@kidrahahjo Thanks for the questions! The new unified filter syntax is already testing document filters, just below where I deleted the tests for the old `--doc-filter` flag.

Codecov is extra noisy on this PR but I don’t think we are really missing anything new. The baseline coverage measurement for the next branch may be wrong because I just force-pushed some changes. I think this PR’s coverage is fine because we’re just removing a feature.

signac/__main__.py
```diff
@@ -32,7 +32,7 @@
 from . import get_project, init_project
 from .common import config
 from .common.configobj import Section, flatten_errors
-from .contrib.filterparse import _add_prefix, parse_filter_arg
```
vyasr, line 35: My guess is that `_add_prefix` can be removed entirely at this point. I can verify that and create a separate issue for that though.

vyasr (approved):
I agree that we don't have to worry about the Codecov diff for now. Otherwise these changes LGTM.

signac/__main__.py
vyasr: #810

PR #807, opened 2022-08-10
Make synced collection a hidden package

### LINKED ISSUE #755: Remove synced collections
The synced collections subpackage introduced in #484 is a core component of signac, but it is complex and potentially useful enough to merit existence outside of the signac package itself. We plan to spin it out into a separate package, so we need to remove it in signac 2.0.

### LINKED ISSUE #790: make synced collections a "hidden" package

kidrahahjo:
## Description

Make synced collection package hidden. 

## Motivation and Context

Resolves #790, resolves #755

## Checklist:


changelog.txt
```diff
@@ -53,6 +53,7 @@ Removed
  - The ``--workspace`` option for the command line ``job`` subcommand (#752).
  - The ability to call ``Project.workspace``, it is strictly a property now (#752).
  - ``ProjectSchema.__call__``, ``ProjectSchema.detect`` (#752).
+ - Public support for the ``synced_collection`` package (#807, #790).
```
bdice, line 56:
```suggestion
 - The public API of the ``synced_collection`` subpackage is no longer accessible (#807, #790). The `synced_collections` code will be separated into its own package in the future.
```

signac/contrib/job.py
```diff
@@ -403,7 +403,7 @@ def statepoint(self):
             you can access a dict copy of the state point by calling it, e.g.
             ``sp_dict = job.statepoint()`` instead of ``sp = job.statepoint``.
             For more information, see
-            :class:`~signac.synced_collections.backends.collection_json.JSONAttrDict`.
+            :class:`~signac._synced_collections.backends.collection_json.JSONAttrDict`.
```
bdice, line 406: This class reference won't work because we're removing the synced collection docs. I wonder if we should add docs for the publicly imported name `signac.JSONDict` and point to that instead.

signac/contrib/migration/v1_to_v2.py
```diff
@@ -15,10 +15,10 @@
 
 import os
 
+from signac._synced_collections.backends.collection_json import BufferedJSONAttrDict
```
bdice, line 18: Maybe we should replace all references to `signac._synced_collections.backends.collection_json.BufferedJSONAttrDict` with `signac.JSONDict`, which is an alias that we publicly expose. Then the dependence on synced collections is more isolated.

signac/contrib/project.py
```diff
@@ -301,7 +301,7 @@ def document(self):
 
         Returns
         -------
-        :class:`~signac.synced_collections.backends.collection_json.BufferedJSONAttrDict`
+        :class:`~signac._synced_collections.backends.collection_json.BufferedJSONAttrDict`
```
bdice, line 304: Same as above: this class reference won't work because we're removing the synced collection docs. I wonder if we should add docs for the publicly imported name `signac.JSONDict` and point to that instead.

bdice (changes requested):
Awesome start! I suggested a few minor changes.

signac/contrib/migration/v1_to_v2.py
kidrahahjo: We also use `JSONAttrDict` in some parts of our code. Since I am not super familiar with the code I am not sure if it's safe to replace the use of `JSONAttrDict` with `BufferedJSONAttrDict` (as `signac.JSONDict` is a `BufferedJSONAttrDict`)

signac/contrib/job.py
kidrahahjo: Won't the docs for JSONDict (which is buffered) and JSONAttrDict be different?

bdice: 😕 This is tricky. We want to just point to public docs for a standalone synced collections package, but if we make synced collections hidden, we can't point to these docs.

Maybe this entire approach is flawed and we should leave `signac.synced_collections` as a public subpackage until we make `synced_collections` standalone.

cc: @glotzerlab/signac-committers

vyasr: See [my comment on the corresponding issue](https://github.com/glotzerlab/signac/issues/790#issuecomment-1229598098). Even if `synced_collections` stays public, I think pointing to a specific class here is a flawed approach. From a user's perspective, do they need any information other than the fact that this attribute is a dict-like object? If we _really_ want to explain the syncing, I would say that we could [point to the parent `SyncedDict` class](https://github.com/glotzerlab/signac/blob/master/signac/synced_collections/data_types/synced_dict.py#L24), but IMO that's exposing implementation details because synchronization is transparent.

There is only one minor exception that comes to my mind immediately, which is that the document currently supports buffering on a per-document basis:
```
with job.doc.buffered:
    # All I/O will be buffered in this block
```
but TBH I'm not sure that's something that we really need to promise at the signac level. The `signac.buffered` API should be the only entrypoint for buffering for signac users IMO:
```
with signac.buffered():
    # Buffer everything
```
since that abstracts over the details of specific backends and all instances. That abstraction guarantees that in a world where signac starts supporting different backends, users still have a single switch in the API to toggle when reads and writes are maximally buffered, and it's up to the implementation of signac to actually perform that buffering to the best of its ability. So I think that ultimately the only API promise (and therefore all that we should document) is that `[job|project].[sp|doc]` are dict-like objects, and perhaps that they support attribute-based access. We shouldn't link to a type because we don't actually promise the type of the object (again, see https://github.com/glotzerlab/signac/issues/790#issuecomment-1229598098 for why).

bdice: I agree with this approach. Thanks @vyasr. @kidrahahjo does this make sense on what to change? If not, I can take a look at changing it myself and push to this PR.

kidrahahjo: As far as I understood this, we do not need to expose any of the implementation details so we can either not specify the type or use MutableMappings (which is another way of saying its a dict like object) instead?

changelog.txt
vyasr: I'm going to accept the first half of this suggestion but not the second. I don't want to codify promises about a new package in our changelog.

signac/contrib/migration/v1_to_v2.py
vyasr: I say let's leave this as is for now. If and when we get around to making `synced_collections` a separate published package it'll get easier since we'll be referencing an external, public package.

vyasr:
@bdice @kidrahahjo I've standardized the docs in a relatively simple way. Let me know what you two think. In the interest of getting 2.0 done. I think we shouldn't belabor the documentation question too much further for now; we can revisit in 2.x, especially once `synced_collections` is actually a separate, accessible package, but I don't want to block 2.0 over that discussion.

bdice (approved):
Thanks for those changes @vyasr. This looks fine to me.

PR #814, opened 2022-09-18
Deprecate config module

### LINKED ISSUE #753: Consider removing config APIs
Currently signac allows users to interact with the config via both Python APIs and the CLI. While this model makes sense for most of signac, it is generally unnecessary for configs. In fact, changes like glotzerlab/signac-flow#573 are moving us towards completely decoupling the configuration logic and just sharing the file. We should investigate whether this approach is viable for signac-flow/other plugin compatibility (i.e. if packages built on top of signac could just rely on configobj directly), in which case we could remove config APIs entirely from signac.

vyasr:
## Description

This PR deprecates the `read_config_file` and `load_config` APIs, the remaining public components of the config module. As of glotzerlab/signac-flow#649 we have confirmed that higher-level signac projects should be able to interface with configuration information solely via the `Project.config` attribute.

In the interest of simplicity I have not bothered to deprecate methods like `search_tree` and `check_and_fix_permissions` since we have no reason to expect external users to be touching them. I plan to remove them without deprecation in 2.0.

## Motivation and Context

Resolves #753.

## Checklist:

### DIFF
--- a/signac/common/config.py
+++ b/signac/common/config.py
@@ -154,10 +154,12 @@ def _get_config(infile=None, configspec=None, *args, **kwargs):
     ),
 )
 def get_config(infile=None, configspec=None, *args, **kwargs):  # noqa: D103
+    """Get configuration from a file."""
     return _get_config(infile, configspec, *args, **kwargs)
 
 
 def _load_config(root=None, local=False):
+    """Load configuration, searching upward from a root path."""
     if root is None:
         root = os.getcwd()
     config = Config(configspec=cfg.split("\n"))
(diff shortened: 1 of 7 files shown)


signac/common/config.py
```diff
@@ -135,36 +148,52 @@ def _get_config(infile=None, configspec=None, *args, **kwargs):
...
+    deprecated_in="1.8",
+    removed_in="2.0",
+    current_version=__version__,
+    details=(
+        "The load_config method is deprecated. Configs should only be "
+        "accessed via a Project instance.",
+    ),
+)
+def load_config(root=None, local=False):
+    """Load configuration, searching upward from a root path."""
```
bdice, line 193: Maybe copy the docstring to the private method.

bdice (approved):

signac/common/config.py
vyasr: Done.

PR #816, opened 2022-09-18
Warn users to migrate when old schema is detected and enable migration CLI

vyasr:
## Description

This PR adds a check in project initialization to ensure that projects with old schemas are detected when signac fails to find a project with the current schema. The resulting check provides the user a meaningful error about how to proceed with updating their schema using our migration code. This PR also exposes a CLI for migration to complement the Python interface, providing a simpler and friendlier approach to one-off migration that is recommended by the error.

## Motivation and Context

Necessary to for projects how to migrate to v2.

## Checklist:

### DIFF
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -634,6 +634,7 @@ def main_update_cache(args):
         _print_err(f"Updated cache (size={n}).")
 
 
+# TODO: This can be uncommented now!
 # UNCOMMENT THE FOLLOWING BLOCK WHEN THE FIRST MIGRATION IS INTRODUCED.
 # def main_migrate(args):
 #     "Migrate the project's schema to the current schema version."
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -634,36 +634,36 @@ def main_update_cache(args):
         _print_err(f"Updated cache (size={n}).")
 
 
-# TODO: This can be uncommented now!
-# UNCOMMENT THE FOLLOWING BLOCK WHEN THE FIRST MIGRATION IS INTRODUCED.
-# def main_migrate(args):
-#     "Migrate the project's schema to the current schema version."
-#     from .contrib.migration import apply_migrations, _get_config_schema_version
-#     from packaging import version
-#     from .version import SCHEMA_VERSION
-#     project = get_project(_ignore_schema_version=True)
-#
-#     root = args.root_directory if args.root_directory else os.getcwd()
-#
-#     schema_version = version.parse(SCHEMA_VERSION)
-#     config_schema_version = _get_config_schema_version(root, schema_version)
-#
-#     if config_schema_version > schema_version:
-#         _print_err(
-#             "The schema version of the project ({}) is newer than the schema "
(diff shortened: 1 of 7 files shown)


signac/__main__.py
```diff
@@ -634,35 +634,36 @@ def main_update_cache(args):
...
+    root = args.root_directory if args.root_directory else os.getcwd()
+
+    schema_version = int(SCHEMA_VERSION)
+    config_schema_version = _get_config_schema_version(root, schema_version)
+
+    if config_schema_version > schema_version:
+        _print_err(
+            f"The schema version of the project ({config_schema_version}) is "
+            "newer than the schema version supported by signac version "
+            f"{__version__}: {schema_version}. Try updating signac."
```
bdice, line 651: I think I like this way better?
```suggestion
            f"newer than the schema version {schema_version} supported by "
            f"signac version {__version__}. Try updating signac."
```

bdice (approved):
I haven't tested this locally but I'm happy with the code in this PR! I'll test it with v1 projects on the `next` branch sometime in the near future.

vyasr:
> I haven't tested this locally but I'm happy with the code in this PR! I'll test it with v1 projects on the `next` branch sometime in the near future.

I'll hold off on merging to let you try a test or two in case you catch bugs that I missed in my testing.

signac/__main__.py
mikemhenry: :+1:

signac/common/config.py
```diff
@@ -15,14 +15,47 @@
...
+    """
+    from ..contrib.errors import IncompatibleSchemaVersion
+    from ..contrib.migration import _get_config_schema_version
+    from ..version import SCHEMA_VERSION, __version__
+
+    schema_version = int(SCHEMA_VERSION)
+
+    try:
+        schema_version = _get_config_schema_version(root, schema_version)
+        assert schema_version != int(SCHEMA_VERSION), (
```
mikemhenry, line 44: Maybe an `if` is better here? This is one of those nit picky things but since `assert` isn't ran with `-O`

mikemhenry (approved):

bdice:
@vyasr Maybe we should wait until I rebase `next` and then rebase this PR?

vyasr:
> @vyasr Maybe we should wait until I rebase `next` and then rebase this PR?

Yes that was my plan. I was also hoping to get one or two additional testers aside from myself, but I think once the rebase is done if nobody has done additional testing at that point we can just go ahead and merge this and let people test it on the main branch.

signac/common/config.py
vyasr: Well, I sort of did that intentionally since this indicates developer error somewhere. This is really intended to be something more like a debug print. I can remove it or change to an if statement if you'd prefer though.

vyasr:
@bdice @mikemhenry if one of you would like to test this locally that would be great, otherwise I think we can go ahead and just merge this.

vyasr:
I'm going to go ahead and merge this. People can test it out on next.

PR #817, opened 2022-09-19
Clarify behavior of groupby default=None.

### LINKED ISSUE #744: No more KeyError after changing groupbydoc to groupby
### Description

When I was adding tests for fixing the [deprecation of groupbydoc](https://github.com/glotzerlab/signac/issues/328), I noticed that there is a [test that looks for a KeyError](https://github.com/glotzerlab/signac/blob/95157b43eddde667c99b941052cc23c1304fddd9/tests/test_project.py#L1073) that doesn't raise one with the new doc query namespace. This leads me to wonder if the sp query namespace should also raise a KeyError

Additionally, should an empty `project.groupby()` raise an error?

### To reproduce

```python
import signac

project = signac.init_project("project")

for i in range(3):
    job = project.open_job(
        dict(
            a=1,
            b=i
        )
    ).init()
    job.doc = dict(
        a=1,
        b=2,
        c=i
    )


# This raise a KeyError
for k,g in project.groupbydoc('ff'):
    print(k)


# These all don't raise errors
for k,g in project.groupby('doc.ff'):
    print(k, list(g))

for k, g in project.groupby('sp.ff'):
    print(k, list(g))

for k,g in project.groupby('ff'):
    print(k, list(g))
    
for k,g in project.groupbydoc():
    print(k, list(g))

for k,g in project.groupby():
    print(k, list(g))
```

bdice:
## Description
Resolves #744.

## Motivation and Context
Clarifies docs. No changelog required.

## Checklist:

### DIFF
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -2721,10 +2721,9 @@ class JobsCursor:
 
         Returns
         -------
-        key : str
-            Grouped key.
-        group : iterable of Jobs
-            Iterable of `Job` instances matching this group key.
+        Tuple[str, Iterable[Job]]
+            A pair whose first element is the grouped key and whose second element
+            is an iterable of `Job` instances matching the key.
 
         """
         _filter = self._filter
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -1082,7 +1082,10 @@ class Project:
         return iter(self.find_jobs())
 
     def groupby(self, key=None, default=None):
-        """Group jobs according to one or more state point parameters.
+        """Group jobs according to one or more state point or document parameters.
+
+        Prepend the key with 'sp.' or 'doc.' to specify the query namespace. If no prefix
+        is specified, group by state point key.
 
         This method can be called on any :class:`~signac.contrib.project.JobsCursor` such as
         the one returned by :meth:`~signac.Project.find_jobs` or by iterating over a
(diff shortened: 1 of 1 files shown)


vyasr (approved):
This LGTM but I'd like to get @cbkerr's opinion before merging just in case.

iblanco11981870 (approved):
looks good to me

cbkerr (approved):
I don't have time to make sure the description matches the code, but I corrected some other aspects of the `groupby` doc string.

signac/contrib/project.py
```diff
@@ -2694,17 +2704,28 @@ def groupby(self, key=None, default=None):
...
+            A default value to be used when a given key is not
             present. The value must be sortable and is only used if not None
             (Default value = None).
 
+        Returns
+        -------
+        key : str
+            Grouped key.
+        group : iterable of Jobs
+            Iterable of `Job` instances matching this group key.
```
vyasr, line 2727:
```suggestion
        Tuple[str, Iterable[Job]]
            A pair whose first element is the grouped key and whose second element
            is an iterable of `Job` instances matching the key.
```

bdice:
@vyasr @cbkerr The changes made after my commits (changes to the "Returns") are incorrect. `itertools.groupby` returns a generator. The generated values have the structure of `Tuple[str, Iterable[Job]]`. I think that aligning with numpydoc would suggest a pattern like this: https://numpydoc.readthedocs.io/en/latest/format.html#yields

```
Yields
------
key :
    Grouped key.
group : iterable of Jobs
    Iterable of `Job` instances matching this group key.
```

Also, we do _not_ know the type of `key` (though it is typed as `str` in the changes made in this PR) because a provided callable can return any type that can be sorted and equality-compared, e.g. `bool` or `int`.

I can file a second PR to fix this, but it might have been nice to let this one sit for a little longer instead of making substantial changes and then merging so quickly...

vyasr:
Ack sorry I wanted to get everything merged ASAP to let the release process kicked off and I didn't expect docs changes to cause problems.

PR #818, opened 2022-09-19
Use Yields in groupby docs.

bdice:
## Description
See https://github.com/glotzerlab/signac/pull/817#issuecomment-1251342595

## Motivation and Context
Fixes docs.

## Checklist:

### DIFF
--- a/signac/contrib/project.py
+++ b/signac/contrib/project.py
@@ -1135,11 +1135,12 @@ class Project:
             present. The value must be sortable and is only used if not None
             (Default value = None).
 
-        Returns
-        -------
-        Tuple[str, Iterable[Job]]
-            A pair whose first element is the grouped key and whose second element
-            is an iterable of `Job` instances matching the key.
+        Yields
+        ------
+        key :
+            Grouped key.
+        group : iterable of Jobs
+            Iterable of `Job` instances matching this group key.
 
         """
         return self.find_jobs().groupby(key, default=default)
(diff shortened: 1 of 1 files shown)


signac/contrib/project.py
```diff
@@ -1135,11 +1135,12 @@ def groupby(self, key=None, default=None):
             present. The value must be sortable and is only used if not None
             (Default value = None).
 
-        Returns
-        -------
-        Tuple[str, Iterable[Job]]
-            A pair whose first element is the grouped key and whose second element
-            is an iterable of `Job` instances matching the key.
+        Yields
```
bdice, line 1138: Should we use `yield from` instead of `return` for the `itertools.groupby` object? Is it clear that this acts as a generator?

The `itertools.groupby` object has no useful methods beyond ones like `__next__` and `__iter__` that define it as an iterable.

vyasr: I would be fine with that switch rather than documenting that we returning an `itertools.groupby` object.

vyasr:
I'm not sure whether I like documenting multiple return values. I've almost exclusively only ever seen it documented as returning a tuple of the contained objects (or in this case, an `Iterable[Tuple[...]]`). I see that `numpydoc` has this in their documentation, and it's fine if we do it, but it may look different from other docs (although I can't think of many signac APIs that return tuples, so maybe it's not an issue here). I'll admit that I don't really care what decision we make, I just wanted to raise the concern about consistency in case someone else had thoughts.

bdice:
> I'm not sure whether I like documenting multiple return values. I've almost exclusively only ever seen it documented as returning a tuple of the contained objects (or in this case, an `Iterable[Tuple[...]]`).

The example I think of that uses this pattern is `numpy.unique`, which can have multiple return values. https://numpy.org/doc/stable/reference/generated/numpy.unique.html

signac/contrib/project.py
bdice: Changed.

vyasr (approved):
Works for me.

bdice:
@cbkerr feel free to merge if you approve, or @b-butler you can approve/merge if you need to move it forward for the 1.8.0 release.

cbkerr:
I don't have time to review this, especially as I don't fully understand `yield` in python.

PR #834, opened 2022-10-27
signac next (2.0 draft)

bdice:
## Description
This PR merges all the changes for signac 2.0 from the `next` branch to the `master` branch. (I created a new branch `next-rebase` rather than force-pushing over the current `next` branch just in case I made mistakes during the rebase process.)

These changes have all been previously reviewed. @glotzerlab/signac-committers If you have time, please take a quick look to make sure that the changes look reasonable and have been merged correctly.

Unless crucial mistakes have been made in the rebase process, we should try to merge this PR with no additional changes and do follow-up PRs for anything that is left.

**This PR should not be squash-merged.**


signac/common/config.py
```diff
@@ -5,195 +5,107 @@
...
 
-DEFAULT_FILENAME = ".signacrc"
-CONFIG_FILENAMES = [DEFAULT_FILENAME, "signac.rc"]
-HOME = os.path.expanduser("~")
-CONFIG_PATH = [HOME]
-FN_CONFIG = os.path.expanduser("~/.signacrc")
+PROJECT_CONFIG_FN = os.path.join(".signac", "config")
+USER_CONFIG_FN = os.path.expanduser(os.path.join("~", ".signacrc"))
 
+# TODO: Consider making this entire module internal and removing all its
```
bdice, line 18: Future work: It looks like the only public part of the API is the `Config` class. We may be able to separate this, or make `Config` a private class (not sure if that would matter, since it's immutable).

signac/common/validate.py
```diff
@@ -5,53 +5,16 @@
...
-def password(value, *args, **kwargs):  # noqa: D103
-    return value
-
-
 def get_validator():  # noqa: D103
-    return Validator({"mongodb_uri": mongodb_uri, "password": password})
+    return Validator()
 
 
+# TODO: Rename to something internal and uppercase e.g. _CFG.
```
bdice, line 17: Future work: reminder to grep for `TODO`s like this. Several of them would be easy to address.

vyasr:
I know that this rebase was tricky, and the changeset is basically unreviewable at this point, so in lieu of reviewing directly my plan is to redo the rebase myself and then diff my final result with the branch in this PR. That should allow us to identify any issues by looking for discrepancies in our rebases.

bdice:
@vyasr thanks a lot for the independent confirmation! I wasn’t 100% sure how to handle those config APIs, since there were conflicting changes. I’ll take a closer look when I’m at a computer and we can merge your PR if all looks good.

bdice:
Closing in favor of #835.

PR #835, opened 2022-10-30
signac 2.0 changes

vyasr:
## Description
This PR merges all the changes for signac 2.0 from the `next` branch to the `master` branch. (I created a new branch `next-rebase-vyas` rather than force-pushing over the current `next` branch just in case I made mistakes during the rebase process.)

These changes have all been previously reviewed. @glotzerlab/signac-committers If you have time, please take a quick look to make sure that the changes look reasonable and have been merged correctly.

Unless crucial mistakes have been made in the rebase process, we should try to merge this PR with no additional changes and do follow-up PRs for anything that is left.

**This PR should not be squash-merged.**

Note: This PR is a replacement for #834 that fixes a couple of minor errors. Due to the size of #834 my review process was to just redo the rebase myself and compare the result, which allowed me to verify that the results are nearly identical aside from the small errors discussed on that PR.


bdice (approved):

mikemhenry (approved):
LGTM :tada:

signac/common/validate.py
```diff
@@ -5,53 +5,16 @@
...
-        except pymongo.errors.InvalidURI:
-            raise VdtValueError(value)
-    return value
-
-
-def password(value, *args, **kwargs):  # noqa: D103
-    return value
-
-
 def get_validator():  # noqa: D103
```
b-butler, line 13: This function has no reason to exist now it seems. Might as well just call `Validator` unless that somehow depends on the state of the stack when called.

signac/syncutil.py
```diff
@@ -160,22 +111,22 @@ class _FileModifyProxy:
 
     Parameters
     ----------
-    root :
+    root : str, optional
```
b-butler, line 114: should we change this to path?

b-butler (approved):
This is obviously massive. I made a couple comments but nothing blocking. I wasn't able to look through everything.

bdice:
Huge thanks to @vyasr @mikemhenry @b-butler. I'm going to merge this and we can address the comments from #834 and this PR in follow-up work.

signac/syncutil.py
bdice: I vote no because this is not necessarily the Project's root/path. In this instance, it is the workspace: https://github.com/glotzerlab/signac/blob/a924bad99c1f0601f9f7ade16101fc2ec0f5e871/signac/sync.py#L502-L503

signac/common/validate.py
bdice: This does seem like it can be cleaned up. I'll file an issue.

PR #848, opened 2022-11-02
Stop filtering out warnings.

vyasr:
## Description

Removes all filters of warnings during pytests.

## Motivation and Context

These were hardcoded at various points during 1.x development due to ongoing deprecations and other related changes that could not be worked around. They are no longer necessary.

## Checklist:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,25 +33,6 @@ source = signac
 omit = 
 	*/signac/common/configobj/*.py
 
-[tool:pytest]
-filterwarnings = 
-	error::FutureWarning:signac.*
-	error::DeprecationWarning:signac.*
-	ignore:Project names are deprecated:FutureWarning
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -33,25 +33,6 @@ source = signac
 omit = 
 	*/signac/common/configobj/*.py
 
-[tool:pytest]
-filterwarnings = 
-	error::FutureWarning:signac.*
-	error::DeprecationWarning:signac.*
```
bdice, line 39: We should leave these warnings as errors. That's how we know we've removed them all. The only lines that should be deleted are the `ignore:` lines.

vyasr: Whoops thanks I missed that these weren't all ignores in the batch delete.

bdice (approved):
Thanks!

PR #850, opened 2022-11-03
Reenable multithreading tests

vyasr:
## Description

Tests of multithreaded support for synced collections were added as part of the development effort. The tests were generally reliable, but we saw sporadic failures on Windows for unknown reasons, so they were marked as conditionally expected to fail on Windows (but done so incorrectly, see below). The tests were changed to be unconditionally `xfail`ed due to development difficulties in #777. This PR adds them back in with a more target scope for skipping. This change also allows us to enable `xfail_strict` in our testing.

## Motivation and Context

I finally have access to a Windows machine myself, so I looked into these again this morning. I cannot reproduce the failures locally on WSL no matter what I try (with/without pytest-xdist, many repetitions with pytest-repeat, etc), but I _can_ reproduce the failures in normal Windows. The error is consistently a `PermissionError` encountered in the `os.replace` call. There is a [Python bug clearly indicating that Python intentionally avoids documenting such incompatibilities, but that this is a known problem with `os.replace` in multithreaded settings on Windows due to the underlying filesystem instruction](https://bugs.python.org/issue46003). As such, we should not expect this to work on Windows, and the original exclusion Windows-based `skip` prior to #777 is correct, not the current blanket `xfail`.

[Our previous thread justifying the change in testing behavior](https://github.com/glotzerlab/signac/pull/777#discussion_r901175366) identified that there was a bug in our old filtering. We were using `"win" in platform.python_implementation()` as a way to identify whether we were on Windows, and using that to determine whether or not to run the tests. `platform.python_implementation()` indicates what _Python_ implementation we are using (CPython, Jython, PyPy, etc), not the operating system, so this filter was clearly incorrect. However, the other claim in that thread is also obviously false:
> Apparently this was getting WINDOWS = True because of that bug, and hiding failures on other platforms too:

At the time we ran tests on Pypy and CPython, and both `"win" in "CPython"` and `"win" in PyPy` are always `False`, not always `True`. The effect of this error therefore could not have been to hide failures on non-Windows platforms, all it could have done is run tests on Windows that we expected to fail. Therefore, my suspicion is that what was observed in that PR were only Windows-specific errors. There were additional reports of failures in #776, but they appear to have been scattered among various other failures that occurred during the CI transition, so it is unclear whether we ever had reliable examples of failures on other operating systems.

This PR reenables these tests on the platforms where we should expect them to succeed. Given the upcoming release of signac 2.0, if we start seeing failures in these tests again we should reinvestigate our claims of multithreading support. The tests don't affect any behavior that signac relies on, only otherwise unused features of synced collections, so this is mostly relevant for documenting synced collections when we split those into a separate package post 2.0. 

## Checklist:


vyasr:
I'm happy to rerun tests on this PR a few times to see if we can reproduce the failures again before we merge.

signac/_synced_collections/backends/collection_json.py
```diff
@@ -116,6 +117,16 @@ def json_attr_dict_validator(data):
 """
 
 
+class _IsWindows:
+    """A truthy type that is False on Windows and True otherwise."""
```
csadorf, line 121: Shouldn't it be the other way round? "_IsWindows" should be `True` on Windows and `False` otherwise?

signac/_synced_collections/backends/collection_json.py
```diff
@@ -116,6 +117,16 @@ def json_attr_dict_validator(data):
...
+
+    def __bool__(self):
+        if sys.platform.startswith("win32") or sys.platform.startswith("cygin"):
+            return False
+        else:
+            return True
```
csadorf, line 127: Assuming it should be `True` on windows and I think there is a typo:
```suggestion
        return sys.platform.startswith("win32") or sys.platform.startswith("cygwin"):
```

signac/_synced_collections/backends/collection_json.py
```diff
@@ -145,10 +158,17 @@ class JSONCollection(SyncedCollection):
...
+    This class is _not_ thread safe on Windows. It relies on ``os.replace`` for
+    atomic file writes, and that method can fail in multithreaded situations if
+    open handles exist to the destination file withiin the same process on a
+    different thread. See https://bugs.python.org/issue46003 for more
+    information.
     """
 
     _backend = __name__  # type: ignore
-    _supports_threading = True
+    _supports_threading = _IsWindows()  # type: ignore
```
csadorf, line 171:
```suggestion
    _supports_threading = not _IsWindows()  # type: ignore
```

tests/test_synced_collections/test_json_buffered_collection.py
```diff
@@ -23,6 +24,8 @@
 )
 from signac._synced_collections.errors import BufferedError, MetadataError
 
+WINDOWS = sys.platform.startswith("win32") or sys.platform.startswith("cygin")
```
csadorf, line 27: Minor suggestion for a better variable name and fix a typo:
```suggestion
ON_WINDOWS = sys.platform.startswith("win32") or sys.platform.startswith("cywgin")
```

tests/test_synced_collections/test_json_collection.py
```diff
@@ -15,6 +16,8 @@
     JSONList,
 )
 
+WINDOWS = sys.platform.startswith("win32") or sys.platform.startswith("cygin")
```
csadorf, line 19: Minor suggestion for a better variable name and fix a typo:
```suggestion
ON_WINDOWS = sys.platform.startswith("win32") or sys.platform.startswith("cygin")
```

signac/contrib/job.py
```diff
@@ -131,10 +131,16 @@ def _save(self):
...
+        # support. The sequence of operations here is imoprtant since changing
+        # the filename updates the lock id.
+        if type(self)._threading_support_is_active:
+            old_lock_id = self._lock_id
+
         self._filename = job._statepoint_filename
-        type(self)._locks[self._lock_id] = type(self)._locks.pop(old_lock_id)
+
+        if type(self)._threading_support_is_active:
+            type(self)._locks[self._lock_id] = type(self)._locks.pop(old_lock_id)
```
csadorf, line 143: This segment feels like implementation details of the base class are leaking here. Maybe this could be abstracted into a classmethod in the base class? That would also make direct class attribute access via `type(...)` unnecessary. What do you think?

csadorf (changes requested):

signac/_synced_collections/backends/collection_json.py
vyasr: I think I've been spending too much time in compiled languages and so I created `_IsWindows` just to have something that would be very obviously something evaluated at "runtime". In practice I don't know that this adds any extra clarity and just increases boilerplate. What if I just inlined
```
_supports_threading = not ( sys.platform.startswith("win32") or sys.platform.startswith("cygin"))
```
Then your previous two comments also become moot.

signac/contrib/job.py
vyasr: I agree. I've moved this logic into a property setter for `filename` in `JSONCollection`, which is the most appropriate location since that class is where the `filename`<-->`lock` relationship is established.

signac/_synced_collections/backends/collection_json.py
vyasr: See the discussion below.

signac/_synced_collections/backends/collection_json.py
vyasr: See below.

setup.cfg
```diff
@@ -34,6 +34,7 @@ omit =
 	*/signac/common/configobj/*.py
 
 [tool:pytest]
+xfail_strict = True
```
csadorf, line 37: I consider this a rather drastic change in the behavior of the overall test suite. Should we consider to apply this a bit more granular, i.e., use `strict=True` where this is relevant?

signac/_synced_collections/backends/collection_json.py
```diff
@@ -227,6 +251,7 @@ def _lock_id(self):
         "registry",
         # These keys are specific to the JSON backend.
         "_filename",
+        "filename",
```
csadorf, line 254: Is this a sneaky bug fix? A property with the same name existed before, right?

signac/_synced_collections/backends/collection_json.py
csadorf: I think having a dedicated private constant for this makes sense and makes the code more readable and overall more consistent. We could reuse that variable also in the tests, no?

signac/contrib/job.py
csadorf: Great, thanks.

signac/_synced_collections/backends/collection_json.py
vyasr: Yes, the property existed but was unused. I must have originally just added it for completeness. The bug fix is actually needed now since we are making use of the property, otherwise I would not have noticed.

setup.cfg
vyasr: Actually, wanting to enable this was my original goal and what made me remember that these multithreaded tests had been xfailed. IMO setting this can only be a good thing because it forces us to immediately deal with changes that incidentally and unexpectedly fix a bug or change behaviors in other unexpected ways that we only catch because a test suddenly starts xpassing. The value of `strict` passed to the decorator supersedes the value in the file, so I would rather have `xfail_strict=True` configured here and then manually mark with `strict=False` cases where for whatever reason we actually need to allow xfails, for instance because a test is flaky. I would prefer to force developers to jump through hoops in order to enable flaky tests.

bdice: I'm +1 for making the change at a global level, but I would be okay with doing it in a separate PR, or at least adding a separate changelog line to indicate this.

bdice (approved):
Thank you for tracking this down and citing the Python bug. This looks like an appropriate fix to me.

signac/_synced_collections/backends/collection_json.py
vyasr: We could. I wouldn't want to leak a constant like that into our public API, but that's the sort of gray area where I would be OK with importing internals into tests I suppose.

vyasr: Done.

bdice:
@csadorf @vyasr I'll let you two sort out the rest of this PR and merge it. I'm going to focus on a few other items for 2.0.

signac/_synced_collections/backends/collection_json.py
vyasr: I lied, this wasn't necessary before because the property was only gettable, not settable. The `__getattr__` override is only triggered when the attribute already exists (unlike `__setattr__`), so a property without a setter avoids this problem unless it is also accessed directly in the implementation of `__getitem__` (which is really only `_data`).

setup.cfg
csadorf: No need for separate PR IMO (albeit cleaner), but +1 for dedicated changelog entry.

csadorf (approved):
Good to go! :)

bdice:
@vyasr Excluding both `win32` and `cygwin` was correct. I tested it locally and reverted my previous changes. I will automerge this PR.

PR #851, opened 2022-11-06
Remove old tests

vyasr:
## Description

This PR removes some tests that are no longer relevant:
- test_jsondict.py: This test module was written to test the old JSONDict class. When we swapped in the `synced_collections` implementation, we aliased it into the old name used in signac to avoid making a breaking change, which resulted in these tests continuing to run and pass silently. These tests are entirely redundant now and can be removed.
- `host_config_test(s)?.sh` These test files were used to test signac features around managing hosts for MongoDB that have been removed on master now and will no longer exist in signac 2.0.

## Motivation and Context

Cleanup for 2.0.

## Checklist:


vyasr:
At the moment, `signac.buffered` is an alias for `BufferedJSONDict.buffer_backend`, which is tested much more extensively in `test_synced_collections/test_json_buffered_collection.py`. That said, I have left `test_buffered_mode.py` in place at the moment since one could argue that we might eventually introduce buffering at the signac layer that is beyond what is provided by `synced_collections`. I'll leave it up to reviewers to decide whether I should remove `test_buffered_mode.py` as well. My thought is that if we remove the test, we should also consider whether the `signac.buffered` API makes sense or if we should remove that as well and force users to interact directly with the underlying `synced_collections` method going forward. IMO that would be a cleaner solution, except that we probably still want some sort of signac-specific method that insulates users from the choice of resource since we may make that configurable one day. That decision is probably out of scope for this PR, though.

bdice (approved):

PR #854, opened 2022-11-06
Move _collect_by_type to project.py

vyasr:
## Description

Moves the `_collect_by_type` function from schema.py to project.py

## Motivation and Context

Addresses and open TODO.

## Checklist:


signac/contrib/project.py
```diff
@@ -50,6 +51,26 @@
 JOB_ID_REGEX = re.compile(f"[a-f0-9]{{{JOB_ID_LENGTH}}}")
 
 
+def _collect_by_type(values):
```
bdice, line 54: This function is only called once. Can we just define it inline in `detect_schema`?

bdice: Only a short docstring would be needed for an inline function definition.

vyasr: We could. It doesn't use any local state for that function and is well-defined as a standalone so I figured we would save the overhead of redefining it on call, but I can change it if you feel strongly.

vyasr: I mean, it's an internal function so I'd be fine deleting the docstring either way. I left it because I thought you would want to see a docstring 😄

bdice: Nah, the docstring was added just to make pydocstyle happy. Not required for functions defined in functions though, iirc.

bdice: There's probably no meaningful overhead here since we don't call `detect_schema` that many times. I'll move this and then merge this PR.

bdice (approved):

PR #856, opened 2022-11-06
Update v2 migration logic.

bdice:
## Description
This PR makes v1 -> v2 schema migration logic a little more robust. I fixed a bug where the `[y/N]` prompt would disappear if the user pressed backspace. I also switched to using `os.path.join` because, unlike in the cases where we use `os.sep.join` internally, we don't need to sacrifice the checks that `os.path.join` does for speed. We typically put explanatory comments anywhere we use `os.sep.join` to be explicit about the performance needs, which are not relevant for the migration script.

(This does not require a changelog entry because the corresponding migration code has not yet been released.)

## Checklist:


signac/contrib/migration/v1_to_v2.py
```diff
@@ -19,6 +19,7 @@
 from signac.common import configobj
 from signac.common.config import _get_project_config_fn
 from signac.contrib.project import Project
+from signac.contrib.utility import _mkdir_p
```
bdice, line 22:
```suggestion
```

signac/contrib/migration/v1_to_v2.py
```diff
@@ -66,25 +65,27 @@ def _migrate_v1_to_v2(root_directory):
...
+        doc = BufferedJSONAttrDict(filename=fn_doc, write_concern=True)
+        doc["signac_project_name"] = cfg["project"]
     del cfg["project"]
     cfg.write()
 
     # Move signac.rc to .signac/config
     v1_fn = os.path.join(root_directory, "signac.rc")
     v2_fn = _get_project_config_fn(root_directory)
-    os.mkdir(os.path.dirname(v2_fn))
+    _mkdir_p(os.path.dirname(v2_fn))
```
bdice, line 80:
```suggestion
    os.mkdir(os.path.dirname(v2_fn))
```

bdice (commented):
I was considering using our `_mkdir_p` utility so that creating `.signac` would not raise if the directory already existed. However, I changed my mind. Raising if `.signac` exists is probably a good safeguard.

signac/contrib/migration/v1_to_v2.py
```diff
@@ -79,12 +81,10 @@ def _migrate_v1_to_v2(root_directory):
...
-        ".signac_shell_history": os.sep.join((".signac", "shell_history")),
-        ".signac_sp_cache.json.gz": os.sep.join(
-            (".signac", "statepoint_cache.json.gz")
-        ),
+        ".signac_shell_history": os.path.join(".signac", "shell_history"),
+        ".signac_sp_cache.json.gz": os.path.join(".signac", "statepoint_cache.json.gz"),
     }
     for src, dst in files_to_move.items():
-        src = os.sep.join((root_directory, src))
+        src = os.path.join(root_directory, src)
```
vyasr, line 88: I'm fine making the switch, but to be clear we also don't benefit from the extra checks in this case since we control the inputs.

signac/common/config.py
```diff
@@ -52,8 +52,8 @@ def _raise_if_older_schema(root):
         raise IncompatibleSchemaVersion(
             "The signac schema version used by this project is "
             f"{schema_version}, but signac {__version__} requires "
-            f"schema version {SCHEMA_VERSION}. Try running python -m "
-            "signac migrate."
+            f"schema version {SCHEMA_VERSION}. Try running 'python -m "
+            "signac migrate'."
```
vyasr, line 56: I'm going to undo this change because it implies a consistency in our formatting of this stuff that isn't there right now (for example [here's one message](https://github.com/glotzerlab/signac/blob/master/signac/__main__.py#L1285), and [here's another](https://github.com/glotzerlab/signac/blob/master/signac/contrib/project.py#L849)). I'm fine with standardizing the code snippets in our messages across the code base, but we should do so consistently in a separate PR.

vyasr, line 56:
```suggestion
            f"schema version {SCHEMA_VERSION}. Try running python -m "
            "signac migrate."
```

vyasr (approved):

signac/contrib/migration/v1_to_v2.py
bdice: Right. It's just about not using a "performance hack" (which we previously decided needed a code comment for explanation in every instance it was used) where a more common function can do the job correctly.

signac/common/config.py
bdice: I just wanted to match this one, in the aim of standardization: https://github.com/glotzerlab/signac/blob/4079d0b1e1d5a85f0b3f54f95fe600eaf9838fc7/signac/contrib/project.py#L235

These two are effectively the same message, so it's weird to write the same message in two different ways even if the rest of the codebase has other messages that are styled differently.

bdice: @vyasr Would you be willing to make a PR with that standardization? The change you reverted was my attempt at moving in that direction. 🙃

PR #857, opened 2022-11-06
Clean up test files

bdice:
## Description
This PR cleans up some unused global variables from test files and makes variable naming more consistent.

One noteworthy change: we need to rename the function `testdata` because pytest is interpreting it as a test based on its name. It is just a helper function that returns a generated random string. This was throwing a warning:

```
tests/test_job.py::testdata
  /opt/hostedtoolcache/Python/3.10.8/x64/lib/python3.10/site-packages/_pytest/python.py:199: PytestReturnNotNoneWarning: Expected None, but tests/test_job.py::testdata returned '61aed30d-f239-4bae-aa5c-24d122db9017', which will be an error in a future version of pytest.  Did you mean to use `assert` instead of `return`?
    warnings.warn(
```

## Checklist:

### DIFF
--- a/tests/test_job.py
+++ b/tests/test_job.py
@@ -2,7 +2,6 @@
 # All rights reserved.
 # This software is licensed under the BSD 3-Clause License.
 import copy
-import io
 import json
 import logging
 import os
@@ -52,12 +51,7 @@ BUILTINS_HASH = "7a80b58db53bbc544fc27fcaaba2ce44"
 NESTED_HASH = "bd6f5828f4410b665bffcec46abeb8f3"
 
 
-def config_from_cfg(cfg):
-    cfile = io.StringIO("\n".join(cfg))
-    return signac.common.config._read_config_file(cfile)
-
-
-def testdata():
+def generate_test_data():
     return str(uuid.uuid4())
 
 
@@ -625,7 +619,7 @@ class TestJobOpenAndClosing(TestJobBase):
 class TestJobDocument(TestJobBase):
     def test_get_set(self):
         key = "get_set"
-        d = testdata()
+        d = generate_test_data()
         job = self.open_job(test_token)
         assert not bool(job.document)
         assert len(job.document) == 0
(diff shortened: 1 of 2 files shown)


tests/test_project.py
```diff
@@ -111,11 +86,7 @@ def test_workspace_directory(self):
         assert os.path.join(self._tmp_pr, "workspace") == self.project.workspace
 
     def test_config_modification(self):
-        # In-memory modification of the project configuration is
-        # deprecated as of 1.3, and will be removed in version 2.0.
-        # This unit test should reflect that change beginning 2.0,
-        # and check that the project configuration is immutable.
-        assert _CURRENT_VERSION < _VERSION_2
```
bdice, line 118: This test was going to fail when we bumped the version to 2.0.0. I fixed it here in advance of the actual version bump.

vyasr: For reference, this was intended to guard against releasing 2.0 with code that was previously removed in #608. This comment lingered on past when the offending code was removed, which is why it's safe to remove now with no additional changes.

vyasr (changes requested):
The current implementation of `generate_test_data` doesn't add much value. In almost every test that is using it I think we would be better off hardcoding values. None of the tests are doing anything that relies on the randomness/uniqueness of a uuid, and in some places we're doing odd things like asserting that the generated values are not identical. Let's get rid of this entirely, or if we really like having at least one default input for the tests that don't need multiple, make this a fixture.

bdice:
@vyasr That sounds like a good plan. I was also troubled by the usage patterns of the `generate_test_data` function but wasn't sure if a thorough change should be in the scope of this PR. I just wanted to fix the warning in this PR that came from treating `def testdata()` as a test. I'll change this PR accordingly.

tests/test_job.py
```diff
@@ -52,15 +51,6 @@ def builtins_dict():
...
 
-def config_from_cfg(cfg):
-    cfile = io.StringIO("\n".join(cfg))
-    return signac.common.config._read_config_file(cfile)
-
-
-def testdata():
-    return str(uuid.uuid4())
-
-
```
csadorf, line 63: Yes, this would be better served by [text()](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.text).  😁

tests/test_project.py
```diff
@@ -67,30 +66,6 @@
 WINDOWS = sys.platform == "win32"
```
csadorf, line 66: Should this be aligned with https://github.com/glotzerlab/signac/pull/850/files#diff-2536a92baaf2e0ef0a296bc63eb34b07ed102cf82c7f6d50f13a68b862d138c6R31 ?

bdice: *sigh* Unfortunately no. We're really testing for two different functionalities. The multithreaded failures occur on both Windows and Cygwin, but the linked view failures occur only on Windows and _not_ Cygwin.

bdice: The right solution is probably to refactor both checks to describe exactly what they're checking, and not just "Windows." It's really two different sets of functionality.

csadorf (approved):

vyasr (approved):

PR #860, opened 2022-11-06
Refactor _Config class.

bdice:
## Description
Refactors `_Config` to be private and have a minimal API. Also makes a few helper functions private.

## Checklist:

### DIFF
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -693,7 +693,7 @@ def main_config_show(args):
     if not isinstance(cfg, Section):
         print(cfg)
     else:
-        for line in config.Config(cfg).write():
+        for line in config._Config(cfg).write():
             print(line)
 
 
--- a/signac/__main__.py
+++ b/signac/__main__.py
@@ -133,15 +133,15 @@ def _open_job_by_id(project, job_id):
         )
 
 
-def find_with_filter_or_none(args):
+def _find_with_filter_or_none(args):
     """Return a filtered subset of jobs or None."""
     if args.job_id or args.filter:
-        return find_with_filter(args)
+        return _find_with_filter(args)
     else:
         return None
 
 
-def find_with_filter(args):
+def _find_with_filter(args):
     """Return a filtered subset of jobs."""
     if getattr(args, "job_id", None):
         if args.filter:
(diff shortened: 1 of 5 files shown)


signac/__main__.py
```diff
@@ -718,6 +697,27 @@ def main_config_show(args):
             print(line)
 
 
+def _verify_config(cfg, preserve_errors=True):
```
bdice, line 700: Aside from the name change from `verify_config` to `_verify_config`, this is a pure move, so that this function is next to `main_config_verify`.

signac/common/config.py
```diff
@@ -97,10 +97,18 @@ def _locate_config_dir(search_path):
             search_path = up
 
 
+class _Config(ConfigObj):
```
bdice, line 100: Changes here:
- renamed `Config` to `_Config`
- removed the arbitrary `*args, **kwargs` to limit the API to only the argument we use (`preserve_errors`)
- moved the class definition above where it is used in `_read_config_file` and `_load_config`

signac/__main__.py
```diff
@@ -133,15 +133,15 @@ def _open_job_by_id(project, job_id):
         )
 
 
-def find_with_filter_or_none(args):
+def _find_with_filter_or_none(args):
```
vyasr, line 136: I'm fine with making all of these very explicitly internal, but I also don't think it's critical in a module named `__main__.py` that is clearly not part of the public API.

signac/common/config.py
vyasr: 'encoding' is a kwarg for the parent constructor, and from looking at the source it doesn't look like setting the class member variable does what we want because each instance has the attribute overridden on construction anyway. At this point I think removing is fine since that means that we haven't been specifying the encoding historically anyway.

vyasr (approved):
LGTM.

signac/__main__.py
bdice: I agree — the other helpers in the main module already began with an underscore so I figured it was a helpful change for consistency. There was no difference in intended visibility so we should just make all of them private.

signac/common/config.py
bdice: Great, that was the same thing I saw with the parent constructor not using the class member.

PR #871, opened 2022-12-06
Treat all warnings as errors

vyasr:
## Description

This PR changes our test suite to treat all warnings as errors.

## Motivation and Context

Warnings are generally something that we want to react to. We currently have a warning-free test suite, so it's a good time to enable this starting with our clean slate so that we can catch any issues that arise, e.g. in our dependencies.

## Checklist:


setup.cfg
```diff
@@ -35,9 +35,7 @@ omit =
 
 [tool:pytest]
 xfail_strict = True
-filterwarnings = 
-	error::FutureWarning:signac.*
-	error::DeprecationWarning:signac.*
+filterwarnings = error
```
bdice, line 38: There are lots of warnings outside of our control from other packages, stuff like deprecated NumPy usage in `tables` that can take a long while to get fixed. I would prefer to restrict this warnings-as-errors to only the signac package. I think this is the right syntax:
```suggestion
filterwarnings = error:::signac.*
```

bdice (changes requested):

setup.cfg
vyasr: We make very limited use of packages like numpy and pandas that have long deprecation timescales due to the size of their user base. In the few cases that we do, if such an issue arises it is a minimal burden to add a `pytest.warns` context. Moreover, while I agree that it can be annoying to deal with those deprecations, ignoring external warnings can make us miss when real issues arise, such as if a dependency deprecates functionality that we truly rely on. We are currently at 0 warnings. I would prefer to make this a blanket filter for all warnings for now, then revisit the discussion if we hit a point where we suddenly have to deal with a large number of irrelevant warnings. It is a lot less work to make the filter less stringent in the future than it would be to use the less strict filter now; we could try to switch later only to discover that a large number of warnings have crept in that we never dealt with. Let's deal with the "too many warnings out of our control" case if and when it arises, rather than having a looser test suite now in anticipation of future pain that may never come.

bdice (approved):

PR #873, opened 2022-12-06
Move calc_id into job.py

vyasr:
## Description

Move hashing.calc_id into the job module. The parameter is also renamed from `spec` to `statepoint` to more accurately reflect its usage.

## Motivation and Context

The id computed by `hashing.calc_id` is not a generic hash id. It is very specifically tied to the project-job-statepoint relationship that forms the core data model in signac. This move helps clarify that. Additionally, if we ever move to support more general file layouts, this move ensures that the calculated id is tightly coupled to the Job class.

## Checklist:


signac/job.py
```diff
@@ -18,15 +19,38 @@
...
 from ._utility import _mkdir_p
 from .errors import DestinationExistsError, JobsCorruptedError
 from .h5store import H5StoreManager
-from .hashing import calc_id
 from .sync import sync_jobs
 
 logger = logging.getLogger(__name__)
 
 
+def calc_id(statepoint):
```
tommy-waltmann, line 31: Would this work better as a method of the job class?

bdice: A free function is my preference here. Most hash functions are written as free functions, and exceptions that I can think of (like parts of Python's `hashlib`) are related to persisting object state for hashing large inputs.

bdice (approved):
I kind of liked the shared hashing functionality being in its own module, but I don't feel strongly. I accept this.

tommy-waltmann (approved):

Issue #876, opened 2022-12-08
Decide which modules should define __all__


vyasr:
While refactoring I noticed that we define `__all__` in some modules that are not `__init__.py` files representing a package. Is that something that we want to do or care about very much? I would be fine relying solely on underscore prefixes in all modules that are not packages and only implementing `__all__` for package files, but I don't feel strongly about this. I just wanted to raise the question for consistency's sake.

bdice:
In theory we should define `__all__` for every public API in the namespace that we expose it in. So maybe this is only needed for package `__init__.py` files, specifically like the root and `migration` directory.

PR #877, opened 2022-12-08
Makes the config module internal

### LINKED ISSUE #874: Make config module internal
The config.py module no longer contains any public APIs. The only publicly visible items are the constants `PROJECT_CONFIG_FN` and `USER_CONFIG_FN`. Should we change this module to `_config`? If so, do we want to keep those two variables public so that users can easily and programmatically see where to expect signac's config files to be stored? To do so I would suggest importing those into `signac/__init__.py` and adding them to `__all__` in that package.

vyasr:
## Description

This PR moves signac.config to signac._config, reflecting that all its members are private. To simplify its usage, all the functions that it contains have had their leading underscores removed since they are now redundant.

## Motivation and Context

Resolves #874 

## Checklist:


signac/project.py
```diff
@@ -20,17 +20,17 @@
 from tempfile import TemporaryDirectory
 from threading import RLock
 
+from ._config import (
+    Config,
+    get_project_config_fn,
+    load_config,
+    locate_config_dir,
+    raise_if_older_schema,
+    read_config_file,
```
b-butler, line 29: We are now expose these without underscores then to the users though `signac.project`. Is the idea here that most users interact with signac only through the top module?

b-butler (commented):
I just have one comment about our import in `signac/project.py`.

signac/project.py
vyasr: That is a good point, and somewhat related to #876. This issue isn't unique. There are many places where we have modules marked as internal. Should we also require that their contents are always prefixed with an underscore? I don't really mind that, so I can revert https://github.com/glotzerlab/signac/pull/877/commits/f94a2a7f1e77b73f17f3df708b4efad81f2c2c83 if you think so.

vyasr: Reverted.

b-butler (approved):
:+1:

bdice (approved):

PR #878, opened 2022-12-18
Move to a fully pyproject.toml based build

vyasr:
## Description

This PR removes setup.py and setup.cfg entirely, migrating all project and build configuration information into pyproject.toml. In the process, all linter configs have also been moved into pyproject.toml. The exception is flake8, which does not (and will not) support pyproject.toml, so the flake8 configuration is now stored in the .flake8 file, which is specific to this linter. Additionally, bump2version also does not support pyproject.toml (although unlike flake8 [the proposal has not been entirely rejected, so it may eventually](https://github.com/c4urself/bump2version/issues/42)), so that configuration has also been moved to a project-specific .bumpversion file.

## Motivation and Context

Various changes to Python packaging over the last 6 or 7 years have moved towards more static packaging and towards storing data in a backend-agnostic format. These changes allow these of setuptools alternatives (like flit) as well as more reproducible builds based on build isolation into virtual environments that provide all necessary build dependencies. Direct invocation of setup.py has been deprecated in the process. The changes in this PR modernize signac's build system for compatibility with these new approaches.

## Checklist:


pyproject.toml
```diff
@@ -1,3 +1,60 @@
+# Copyright (c) 2020 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+[build-system]
+build-backend = "setuptools.build_meta"
+requires = ["setuptools"]
```
vyasr, line 6: setuptools >= 64 is required for editable installs (pip install -e) with pyproject.toml-only builds. I don't think we need to encode that requirement here, just wanted to mention it.

vyasr: OK, this is quite odd. I tried changing our CI to do a non-editable pip install of the package, and I started seeing failures in `test_single_writer_multiple_reader_same_instance`. [Here's one such example](https://github.com/glotzerlab/signac/actions/runs/3726228439/jobs/6320325740). The problem seems isolated to specific combinations in the test matrix, so I suspect that the issue is in some invalid assumption of the test. It's completely unrelated to this PR's changes, so I just bumped the setuptools requirement to move forward and continue doing an editable install. Just wanted to document the issue here in case someone sees it elsewhere.

doc/installation.rst
```diff
@@ -53,13 +53,18 @@ Consider installing :ref:`optional dependencies <optional_dependencies>`.
 Source Code Installation
 ========================
 
-Alternatively you can clone the `git repository <https://github.com/glotzerlab/signac>`_ and execute the ``setup.py`` script to install the package.
+Alternatively you can clone the `git repository <https://github.com/glotzerlab/signac>`_ and pip install it directly, or install directly from git.
```
klywang, line 56:
```suggestion
Alternatively you can clone the `git repository <https://github.com/glotzerlab/signac>`_ and pip install it directly (Option 1), or install directly from git (Option 2).
```

klywang (approved):

pyproject.toml
```diff
@@ -1,3 +1,59 @@
...
+
+[project.urls]
+Homepage = "https://signac.io"
+Documentation = "https://docs.signac.io"
+Download = "https://pypi.org/project/signac/"
+Source = "https://github.com/glotzerlab/signac"
+Issues = "https://github.com/glotzerlab/signac/issues"
+
+[tool.setuptools]
+packages = { find = { namespaces = false, exclude = ["tests*", "benchmarks*"] } }
```
bdice, line 55: Should this be written in some kind of nested syntax instead of flat?

pyproject.toml
```diff
@@ -19,3 +75,24 @@ force-exclude = '''
...
+[tool.mypy]
+ignore_missing_imports = true
+
+[tool.pytest.ini_options]
+xfail_strict = true
+filterwarnings = "error"
+
+[tool.coverage.run]
+branch = true
+concurrency = [ "thread" , "multiprocessing" ]
```
bdice, line 95: Is all this whitespace a TOML requirement? I prefer formatting like:
```suggestion
concurrency = ["thread", "multiprocessing"]
```

setup.cfg
```diff
@@ -1,48 +0,0 @@
...
-current_version = 1.8.0
-commit = True
-tag = False
-message = Bump up to version {new_version}.
-
-[metadata]
-description-file = README.md
-
-[bdist_wheel]
-python-tag = py3
```
bdice, line 11: Did we lose the python tag specifier?

pyproject.toml
vyasr: In TOML [inline tables are designed to live on a single line](https://toml.io/en/v1.0.0#inline-table). I could reduce the nesting though, something like this should work. I can try it.
```suggestion
[tools.setuptools.packages.find]
namespaces = false
exclude = ["tests*", "benchmarks*"]
```

pyproject.toml
vyasr: No, it is not required.

setup.cfg
vyasr: This is outdated. I found a few different random threads indicating that this section is no longer necessary, but the most authoritative source is the PyPA's wheel documentation and the only usage of this section I found anywhere was around [explicitly declaring that a wheel is universal](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#wheels). I also tested building wheels both with and without this section prior to moving config info from setup.cfg to pyproject.toml, and then I build wheels again after the switch. In all cases the produced wheel had the same tag: `signac-1.8.0-py3-none-any.whl`.

pyproject.toml
```diff
@@ -1,3 +1,61 @@
...
+signac = "signac.__main__:main"
+
+[project.urls]
+Homepage = "https://signac.io"
+Documentation = "https://docs.signac.io"
+Download = "https://pypi.org/project/signac/"
+Source = "https://github.com/glotzerlab/signac"
+Issues = "https://github.com/glotzerlab/signac/issues"
+
+[tool.setuptools]
```
vyasr, line 54:
```suggestion
```

.pre-commit-config.yaml
```diff
@@ -45,6 +43,8 @@ repos:
             ^doc/|
             ^tests/|
           )
+        additional_dependencies:
+          - toml
```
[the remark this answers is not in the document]
bdice: I would add a comment linking to this issue.

bdice (approved):
@vyasr My reasons for not approving earlier were both (1) and (2). I haven't done much with non-setup.py builds and lack deep (and up to date) expertise in this area. I am deferring to your knowledge and I am mostly reviewing this code for perceived completeness. I also don't think I can allocate the time to learn that at the level that I would need at this exact moment (because of holidays and I'm recovering from illness). I trust your judgment and appreciate your responses to my questions, so I'm approving.

I saw you opened a PR to signac-flow. Please consider making the same kinds of changes for signac-dashboard, if you are able. That would be ideal for consistency.

PR #895, opened 2023-02-06
Use Python 3.11 for package releases, enable 3.11 testing with newest dependencies.

### LINKED ISSUE #885: Enable testing for Python 3.11 once wheels exist for h5py
PR #875 adds support for Python 3.11. However, we cannot enable testing with the full set of optional dependencies until wheels are available for `h5py` on Python 3.11. See this issue for tracking: https://github.com/h5py/h5py/issues/2146

Once wheels exist, we can revert the changes from this commit: https://github.com/glotzerlab/signac/pull/875/commits/50fa8c3cf10884e229b4a034d90b41a059a075e3

The action items are:
- add a matrix item for Python 3.11 with newest dependencies
- use Python 3.11 for CI during releases

bdice:
## Description
Resolves #885.

## Checklist:

### DIFF
--- a/.github/workflows/run-pytest.yml
+++ b/.github/workflows/run-pytest.yml
@@ -24,6 +24,7 @@ jobs:
         config: [ {python: '3.8', dependencies: 'newest'},
                   {python: '3.9', dependencies: 'newest'},
                   {python: '3.10', dependencies: 'newest'},
+                  {python: '3.11', dependencies: 'newest'},
                   {python: '3.11', dependencies: 'minimal'},
                   {python: '3.8', dependencies: 'oldest'} ]
     steps:
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -21,7 +21,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
-        python-version: "3.10"
+        python-version: "3.11"
     - name: Install dependencies
       run: |
         # We must explicitly install the requirements so that we can force


.github/workflows/run-pytest.yml
```diff
@@ -24,6 +24,7 @@ jobs:
         config: [ {python: '3.8', dependencies: 'newest'},
                   {python: '3.9', dependencies: 'newest'},
                   {python: '3.10', dependencies: 'newest'},
+                  {python: '3.11', dependencies: 'newest'},
                   {python: '3.11', dependencies: 'minimal'},
```
tommy-waltmann, line 28: Now that you have the ability to test with the newest dependencies on python 3.11, do you still want to test python 3.11 with the minimal dependencies? Is having a "minimal" classification even necessary anymore? If not, much of the logic in this file can likely be simplified.

tommy-waltmann (commented):
I left some questions for consideration moving forward.

.github/workflows/run-pytest.yml
vyasr: We keep one test config running on 'minimal' so that we know the test suite will pass without dependencies that are listed as optional. That ensures that 1) we properly skip tests that require optional dependencies to function, and 2) no hard dependency is accidentally listed as optional. In practice, since signac is pure Python it would probably be safe to just run one of the 4 Python versions as minimal rather than running both minimal and newest for 3.11, I don't have a strong opinion there.

bdice: We have historically run the latest Python version with both “newest” and “minimal” dependencies. I think we should stick to that. Minimal tests run super fast.

bdice:
@glotzerlab/signac-committers Can someone review and approve?

csadorf (approved):
LGTM

PR #901, opened 2023-03-02
Update GitHub Actions and dependency versions.

bdice:
## Description
Updates GitHub Actions and dependency versions.

PRs #898 and #900 were closed because code coverage dropped. However, it appears this was due to a rate limit (either from GitHub or Codecov) preventing some coverage reports from being uploaded. I have created a new secret `CODECOV_TOKEN` for this repository so that it is used instead of the default Codecov "public repo" token which appears to be rate-limited. See https://github.com/codecov/codecov-action/issues/557.

## Checklist:

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -41,14 +41,14 @@ jobs:
       run:
         python -m pip install signac --progress-bar off --no-index -f dist/
     - name: Run MongoDB
-      uses: supercharge/mongodb-github-action@1.7.0
+      uses: supercharge/mongodb-github-action@1.9.0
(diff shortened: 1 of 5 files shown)


bdice:
@csadorf I believe this addresses the code coverage issue. It incorporates the changes from the PRs that were closed due to reduced coverage and a few other small updates.

bdice:
> If we are still hitting a rate limit, then we can limit the upload to happen on only one element of our matrix

We would not get full coverage. Our different OS and dependency configurations have different line coverage.

csadorf (commented):
Thanks a lot for looking into this!! 👍🏻

PR #903, opened 2023-03-22
Remove Zenodo.

bdice:
## Description
This PR removes Zenodo metadata.

## Motivation and Context
We haven't been updating the [Zenodo records for signac](https://zenodo.org/record/2581327) (last updated February 2019). To simplify the repository, we can remove this. Discussed offline with @csadorf @vyasr.

cc: @joaander in case this affects how we track citation statistics. We may also remove the Zenodo reference from https://docs.signac.io/en/latest/acknowledge.html.

## Checklist:


bdice:
It looks like `contributors.yaml` was being used to generate Zenodo metadata. Do we want to keep tracking contributors in the same way, or change that somehow?

joaander:
The siganc developer team is free to track citations however they wish.

vyasr:
> It looks like `contributors.yaml` was being used to generate Zenodo metadata. Do we want to keep tracking contributors in the same way, or change that somehow?

No strong opinion, so I say we just leave it as is until we have a reason to do something else. I think maintaining a simpler text format would also be fine, but no value add in converting now.

pyproject.toml
```diff
@@ -80,7 +80,7 @@ skip_glob = 'signac/_vendor/*'
 
 [tool.pydocstyle]
 convention = "numpy"
-match = "^((?!\\.sync-zenodo-metadata|setup).)*\\.py$"
+match = "^((?!setup).)*\\.py$"
```
vyasr, line 83: Not technically in scope, but since we're modifying this line anyway I think we can get rid of it entirely now. We don't have a setup.py file anymore.

vyasr (changes requested):

changelog.txt
```diff
@@ -70,6 +70,7 @@ Removed
  - The ``Config`` class (#860).
  - The ``DictManager`` class (#858).
  - The ``signac.testing`` module (#863).
+ - Zenodo is no longer updated (#903).
```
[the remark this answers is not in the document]
cbkerr: Doesn't hurt to have it in case someone is looking for changes

CITATION.cff
```diff
@@ -9,7 +9,7 @@ abstract: |
     a well-defined indexable storage layout for data and metadata. This makes it easier to
     operate on large data spaces, streamlines post-processing and analysis and makes data
     collectively accessible.
-doi: 10.5281/zenodo.2581327
+doi: "10.1016/j.commatsci.2018.01.035"
```
vyasr, line 12:
```suggestion
```

pyproject.toml
vyasr, line 83:
```suggestion
```

vyasr (approved):

PR #905, opened 2023-03-29
Warn if old schema on init_project()

cbkerr:
## Description

Adds test and fix for detecting an existing project with an old schema version when running `signac.init_project()`.

Tagging @vyasr as the main author of the migration code.

## Motivation and Context

Since `signac.init_project()` is idempotent, lots of existing scripts probably use it. Till now, running `signac.init_project()` in the directory of a project using the old schema would create the new config files in `.signac` but leave the old `signac.rc` file.

While these are not conflicting configurations when run with signac 2.0, many users have different versions in different environments, so not cleaning up the old config files could result in unexpected and hard to debug behavior if an old version loads the old `signac.rc` file.

Thus, we should also detect if an existing old project exists even when initializing a new project and tell users about migrating schema versions. I use the existing code path for that.

## Checklist:

### DIFF
--- a/signac/_config.py
+++ b/signac/_config.py
@@ -44,8 +44,9 @@ def _raise_if_older_schema(root):
         schema_version = _get_config_schema_version(root, int(SCHEMA_VERSION))
         assert schema_version != int(SCHEMA_VERSION), (
             "Migration schema loader succeeded in loading a config file "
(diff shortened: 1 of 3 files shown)


vyasr (approved):
Seems like a useful change to me.

PR #926, opened 2023-05-29
Check up on restrictions to signac view

cbkerr:
## Description

I had a space in my state point and was surprised that I couldn't make a signac view. This fix worked for me on mac but I'll run the tests on Windows and Linux through GitHub.

## Motivation and Context

The PR that created this restriction was initially a fix involving path separation characters but there was [no reason given for also excluding spaces and asterisks](https://github.com/glotzerlab/signac/pull/233#issuecomment-547587695). I was able to create views in tests with these characters on mac.

I tested this locally with my project that had spaces in the state point after this change, and it worked.

There are [other characters we might want to exclude as well](https://www.informit.com/articles/article.aspx?p=1144082&seqNum=5), like `?`, `[]`, `{}`. These characters might be annoying to escape on the terminal, but a strong use of view is to browse in a file browser.

Perhaps we can emit a warning about special characters being used?

I think if the tests pass in creating views, then we should be good to create views with spaces and asterisks in the state point.

## Checklist:

- [ ] The [package documentation](https://github.com/glotzerlab/signac/tree/main/doc) and [framework documentation](https://docs.signac.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.


signac/linked_view.py
```diff
@@ -64,7 +64,7 @@ def create_linked_view(project, prefix=None, job_ids=None, path=None):
     key_list = [k for job in jobs for k in job.statepoint().keys()]
     value_list = [v for job in jobs for v in job.statepoint().values()]
     item_list = key_list + value_list
-    bad_chars = [os.sep, " ", "*"]
+    bad_chars = [os.sep]
```
b-butler, line 67: We don't need a list any more. It should also be a quicker check to use `==`.

b-butler (changes requested):
Just a couple of quick changes.

vyasr:
Spaces were probably excluded because they don't work on older filesystems, and I'm guessing that one of the supercomputers we originally cared about (maybe the original Stampede, not even Stampede2? Not sure) was an example. I'm fine with removing the restriction and leaving the onus on the user to recognize why a view fails if they have such a filesystem.

bdice:
@cbkerr @b-butler Can you push this through?

b-butler (approved):

PR #970, opened 2024-01-25
Fix parsing of $not expression on command line

cbkerr:
## Description

- fixes bug
- adds test to `test_shell.py`
- I do not fix the testing code in `test_find_command_line_interface.py` that does not properly test the code path but I left a todo for now.

## Motivation and Context

The `$not` expression was broken on the command line. I noticed this while working on https://github.com/glotzerlab/signac-docs/pull/199

signac was inserting `sp.` prefixes before the operator

```bash
$ signac find '{"$not": {"a": 1}}'
Error: "Invalid operator placement 'sp.$not.a'."
```

The function `filterparse._add_prefix` does the inserting, which later triggered an error in `_search_indexer._find_expression`. This code path wasn't tested.

## Checklist:

### DIFF
--- a/signac/filterparse.py
+++ b/signac/filterparse.py
@@ -195,6 +195,8 @@ def parse_filter_arg(args):
 
 def _add_prefix(filter):
     """Add prefix "sp." to a (possibly nested) filter."""
+
+    # Logical operators ($and, $or, $not) should not be prefixed, but their values should.
     for key, value in filter.items():
         if key in ("$and", "$or"):
             if isinstance(value, list) or isinstance(value, tuple):
--- a/signac/filterparse.py
+++ b/signac/filterparse.py
@@ -195,7 +195,6 @@ def parse_filter_arg(args):
 
 def _add_prefix(filter):
     """Add prefix "sp." to a (possibly nested) filter."""
-
     # Logical operators ($and, $or, $not) should not be prefixed, but their values should.
     for key, value in filter.items():
         if key in ("$and", "$or"):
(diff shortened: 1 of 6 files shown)


tests/test_shell.py
```diff
@@ -280,6 +280,14 @@ def test_find(self):
                 == [job.id for job in project.find_jobs({"doc.b": i})][0]
             )
 
+        # ensure that there are no errors due to adding sp and doc prefixes
+        # by testing on all the example complex expressions
+        from test_find_command_line_interface import FILTERS
```
bdice, line 285: Can we define `FILTERS` in some common file, maybe in `conftest.py` as a fixture?

Alternatively, can we move all `import` statements to the top of the file?

signac/filterparse.py
```diff
@@ -203,6 +203,8 @@ def _add_prefix(filter):
                 raise ValueError(
                     "The argument to a logical operator must be a list or a tuple!"
                 )
+        elif key == "$not":
```
bdice, line 206: Maybe add a comment above the `$and`/`$or` handling that says:

```python
# Logical operators ($and, $or, $not) should not be prefixed, but their values should.
```

tests/test_shell.py
cbkerr: I moved the import up, but I like having FILTERS defined near where they are most used.

bdice: Test files should not import from other test files. It is generally a good design principle for test files to be independent of one another.

bdice: Can we refactor that before merging?

bdice (approved):
I'd like to see `test_shell` avoid importing from `test_find_command_line_interface`, but I won't block this PR on that.

PR #975, opened 2024-02-07
Improve performance

joaander:
## Description

Improve the performance of signac for many use-cases, especially when workspaces have with large job counts. Read the complete commit messages for full details on each change.

To summarize:
* Introduce `Job.statepoint_mapping` (edit: `Job.cached_statepoint`)which allows cached, read-only access to the statepoint. `statepoint_mapping` is loaded on demand when the job is not in the cache.
* Use the statepoint cache in additional code paths.
* All `open_job` code paths now lazily load `Job._statepoint`.
* Add `validate_statepoint` argument to `Job.init`. When `False`, `init` checks only that the job directory exists.
* Cache job ids in `JobsCursor` so that `__len__` and `__contains__` are O(1).
* Re-use the results from `listdir` when iterating over jobs in `Project` and bypass the `exists` check on every job as it is opened.

## Motivation and Context

Users would like their scripts to complete quickly. I will post benchmark results in the comments.

## Checklist:

- [ ] The [package documentation](https://github.com/glotzerlab/signac/tree/main/doc) and [framework documentation](https://docs.signac.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.

### DIFF
--- a/signac/job.py
+++ b/signac/job.py
@@ -11,8 +11,8 @@ import os
 import shutil
 from copy import deepcopy
 from threading import RLock
-from typing import FrozenSet
 from types import MappingProxyType
+from typing import FrozenSet
 
 from synced_collections.backends.collection_json import (
     BufferedJSONAttrDict,
@@ -474,7 +474,8 @@ class Job:
                 if self._statepoint_mapping is None:
                     # Load state point data lazily (on access).
                     self._statepoint = _StatePointDict(
-                        jobs=[self], filename=self._statepoint_filename,
+                        jobs=[self],
+                        filename=self._statepoint_filename,
                     )
                     statepoint = self._statepoint.load(self.id)
 
@@ -483,7 +484,9 @@ class Job:
                 else:
                     # Create _StatePointDict lazily with a known statepoint dict.
                     self._statepoint = _StatePointDict(
-                        jobs=[self], filename=self._statepoint_filename, data=self._statepoint_mapping
+                        jobs=[self],
+                        filename=self._statepoint_filename,
+                        data=self._statepoint_mapping,
                     )
 
                 self._statepoint_requires_init = False
--- a/signac/job.py
+++ b/signac/job.py
@@ -297,8 +297,14 @@ class Job:
         else:
             # Only an id was provided. State point will be loaded lazily.
             self._id = id_
-            self._statepoint_mapping = None
             self._statepoint_requires_init = True
+
+            # Fetch the statepoint mapping from the project's cache. Don't load it
+            # from disk on a cache miss (will be loaded on demand).
+            try:
+                self._statepoint_mapping = project._sp_cache[id_]
+            except KeyError:
+                self._statepoint_mapping = None
 
     def _initialize_lazy_properties(self):
         """Initialize all properties that are designed to be loaded lazily."""
@@ -485,6 +491,7 @@ class Job:
 
                     # Update the project's state point cache when loaded lazily
                     self._project._register(self.id, statepoint)
+                    self._statepoint_mapping = statepoint
                 else:
                     # Create _StatePointDict lazily with a known statepoint dict.
                     self._statepoint = _StatePointDict(
--- a/signac/job.py
+++ b/signac/job.py
@@ -12,6 +12,7 @@ import shutil
 from copy import deepcopy
 from threading import RLock
 from typing import FrozenSet
+from types import MappingProxyType
 
 from synced_collections.backends.collection_json import (
     BufferedJSONAttrDict,
@@ -287,12 +288,12 @@ class Job:
             except TypeError:
                 raise KeyTypeError
 
-            self._statepoint_dict = deepcopy(statepoint)
+            self._statepoint_mapping = statepoint
             self._statepoint_requires_init = True
         else:
             # Only an id was provided. State point will be loaded lazily.
             self._id = id_
-            self._statepoint_dict = None
+            self._statepoint_mapping = None
             self._statepoint_requires_init = True
 
     def _initialize_lazy_properties(self):
@@ -331,7 +332,7 @@ class Job:
 
     def __repr__(self):
         return "{}(project={}, statepoint={})".format(
-            self.__class__.__name__, repr(self._project), self.statepoint_dict
+            self.__class__.__name__, repr(self._project), self.statepoint_mapping
         )
 
     @property
@@ -404,10 +405,10 @@ class Job:
         self.statepoint = statepoint
 
     @property
-    def statepoint_dict(self):
-        """Get a copy of the job's statepoint as a dict.
+    def statepoint_mapping(self):
+        """Get a copy of the job's statepoint as a read-only mapping.
 
-        `statepoint_dict` uses the statepoint cache to provide fast access to the
+        `statepoint_mapping` uses the statepoint cache to provide fast access to the
         job's statepoint for reading.
 
         .. note::
@@ -421,13 +422,13 @@ class Job:
 
         Returns
         -------
-        dict
+        Mapping
             Returns the job's state point.
         """
-        if self._statepoint_dict is None:
-            self._statepoint_dict = deepcopy(self._project._get_statepoint(self._id))
+        if self._statepoint_mapping is None:
+            self._statepoint_mapping = self._project._get_statepoint(self._id)
 
-        return self._statepoint_dict
+        return MappingProxyType(self._statepoint_mapping)
 
     @property
     def statepoint(self):
@@ -441,7 +442,7 @@ class Job:
 
         .. tip::
 
-            Use `statepoint_dict` for faster access to read the statepoint.
+            Use `statepoint_mapping` for faster access to read the statepoint.
 
         .. warning::
 
@@ -470,21 +471,19 @@ class Job:
         """
         with self._lock:
             if self._statepoint_requires_init:
-                if self._statepoint_dict is None:
+                if self._statepoint_mapping is None:
                     # Load state point data lazily (on access).
                     self._statepoint = _StatePointDict(
                         jobs=[self], filename=self._statepoint_filename,
                     )
                     statepoint = self._statepoint.load(self.id)
 
-                    self._statepoint_dict_requires_init = False
-
                     # Update the project's state point cache when loaded lazily
                     self._project._register(self.id, statepoint)
                 else:
                     # Create _StatePointDict lazily with a known statepoint dict.
                     self._statepoint = _StatePointDict(
-                        jobs=[self], filename=self._statepoint_filename, data=self._statepoint_dict
+                        jobs=[self], filename=self._statepoint_filename, data=self._statepoint_mapping
                     )
 
                 self._statepoint_requires_init = False
--- a/signac/job.py
+++ b/signac/job.py
@@ -416,20 +416,20 @@ class Job:
 
     @property
     def cached_statepoint(self):
-        """Get a copy of the job's statepoint as a read-only mapping.
+        """Get a copy of the job's state point as a read-only mapping.
 
-        :py:attr:`cached_statepoint` uses the statepoint cache to provide fast access to
-        the job's statepoint for reading.
+        :py:attr:`cached_statepoint` uses the state point cache to provide fast access to
+        the job's state point for reading.
 
         .. note::
 
-            Create and update the statepoint cache by calling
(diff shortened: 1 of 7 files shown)


joaander:
@tcmoore3 @janbridley here are the signac modifications I've been talking about. I will post benchmarks soon.

joaander:
Here are the results for a suite of benchmarks run on cheme-hodges (NVME) with 100,000 jobs in the workspace. For example, the `access_statepoint_mapping` benchmark is:
```python
for job in project:
    job.statepoint_mapping['a']
```

The "no cache" column is run with no statepoint cache file on disk. `cached` is measured after executing `signac update-cache`.

`main`:
| Benchmark                 | no cache (s)   | cached (s)   |
|---------------------------|----------------|--------------|
| iterate_jobs              | 0.365          | 0.379        |
| with_job                  | 5.6            | 5.65         |
| open_job_id               | 0.299          | 0.303        |
| open_job_statepoint       | 1.34           | 1.31         |
| open_job_statepoint_id    | 0.942          | 0.938        |
| access_statepoint_mapping | 3.73           | 3.72         |
| access_statepoint         | 3.55           | 3.62         |
| access_job_document       | 8.08           | 8.03         |
| access_job_stores         | 34.2           | 33.9         |
| find_all                  | 2.07           | 1.36         |
| groupby                   | 2.47           | 1.66         |
| project_len               | 6.6            | 6.62         |
| project_contains          | 0.47           | 0.474        |
| find_jobs_len             | 24.9           | 30.1         |
| find_jobs_contains        | >240           | >240         |

This pull request (7998f9a0b2084f99493312d17ee28391cffce8e4):
| Benchmark                 |   no cache (s) |   cached (s) |
|---------------------------|----------------|--------------|
| iterate_jobs              |         0.158  |       0.16   |
| with_job                  |         0.533  |       0.541  |
| open_job_id               |         0.323  |       0.169  |
| open_job_statepoint       |         0.507  |       0.579  |
| open_job_statepoint_id    |         0.0733 |       0.0762 |
| access_statepoint_mapping |         1.08   |       0.292  |
| access_statepoint         |         3.27   |       3.3    |
| access_job_document       |         3.01   |       2.79   |
| access_job_stores         |        25.5    |      25.6    |
| find_all                  |         1.15   |       0.478  |
| groupby                   |         2.54   |       1.65   |
| project_len               |         6.64   |       6.67   |
| project_contains          |         0.497  |       0.398  |
| find_jobs_len             |         1.06   |       0.391  |
| find_jobs_contains        |         1.17   |       0.527  |

joaander:
Here is the same on Great Lakes scratch (GPFS).

`main`:
| Benchmark                 | no cache (s)   | cached (s)   |
|---------------------------|----------------|--------------|
| iterate_jobs              | 45.5           | 46.6         |
| with_job                  | 136.0          | 128.0        |
| open_job_id               | 49.0           | 50.6         |
| open_job_statepoint       | 1.69           | 1.66         |
| open_job_statepoint_id    | 1.19           | 1.17         |
| access_statepoint_mapping | 117.0          | 114.0        |
| access_statepoint         | 117.0          | 123.0        |
| access_job_document       | 199.0          | 201.0        |
| access_job_stores         | >240           | >240         |
| find_all                  | 120.0          | 2.17         |
| groupby                   | 113.0          | 2.24         |
| project_len               | 9.37           | 9.36         |
| project_contains          | 50.1           | 52.3         |
| find_jobs_len             | 151.0          | 45.1         |
| find_jobs_contains        | >240           | >240         |

This pull request (7998f9a0b2084f99493312d17ee28391cffce8e4):
| Benchmark                 | no cache (s)   | cached (s)   |
|---------------------------|----------------|--------------|
| iterate_jobs              | 0.214          | 0.215        |
| with_job                  | 52.5           | 53.2         |
| open_job_id               | 48.8           | 0.256        |
| open_job_statepoint       | 0.616          | 0.732        |
| open_job_statepoint_id    | 0.0893         | 0.0877       |
| access_statepoint_mapping | 116.0          | 0.403        |
| access_statepoint         | 110.0          | 115.0        |
| access_job_document       | 116.0          | 111.0        |
| access_job_stores         | >240           | >240         |
| find_all                  | 119.0          | 0.764        |
| groupby                   | 115.0          | 2.23         |
| project_len               | 9.42           | 9.41         |
| project_contains          | 50.8           | 50.7         |
| find_jobs_len             | 112.0          | 0.66         |
| find_jobs_contains        | 106.0          | 0.763        |

joaander:
This is ready for review. I suggest waiting to merge until I complete additional testing of this branch with signac-flow.

bdice (commented):
I worked quite a bit on improving signac performance prior to signac 2. At first glance, these optimizations seem fine, but it would be nice to know if there are tradeoffs in guaranteed consistency. The two cases that are very hard to protect with increased levels of caching / reduced validation are:
- parallel access from multiple Python interpreters modifying the same signac project
- files being manually modified on disk by researchers who are unaware of signac's data model

At some point, the filesystem itself is the layer that gives signac atomicity, in the sense of a database transaction. Cutting out the filesystem where possible is important for performance, but may come at a cost to ACID properties (atomicity, consistency, isolation, durability). If you anticipate significant impact to those properties, please share your thoughts.

joaander:
All changes around `statepoint_mapping` and the more aggressive use of the statepoint cache fully retain the existing data model. As noted in one of the commit comments, the only significant change occurs when a user opens a job by statepoint with numpy arrays in the dictionary. The `NumpyConversion` warning is deferred until/if the user access `Job.statepoint`. (They would have received this warning when previously calling Job.`init()`).

Removing the exists checks while iterating and caching ids in `find_jobs` is crucial to allowing faster iteration through the project. Depending on timing, this *may* (or may not) cause different behavior in cases where users remove job directories after starting iteration but before the iteration reaches the removed job. In signac 2.1, the user would receive a `KeyError`: https://github.com/glotzerlab/signac/blob/2e9a23b3c781de5a1a3ca915f9177cdf75e78286/signac/project.py#L545-L547 In this pull request, the user would receive a `Job` instance that knows its id at a minumim. If the statepoint is in the cache, the user will be able to call `job.init()` to restore the missing directory. If the statepoint is not in the cache, then the user will get an error when calling `job.init()`. There are similar race conditions if the user attempts to open the job document or stores after the job is created if the job directory is removed.

Iterating over jobs while at the same time removing them is asking for undefined behavior. I don't see a functional difference between a `KeyError` from *signac* or e.g. a `FileNotExists` error when a user attempts to open `job.fn('some_file')`.

Please be aware that these race conditions existed prior to this pull request. A user could remove the directory after the `self._contains_job_id(id)` check but before the job is accessed (similarly for documents and stores) and get the same `FileNotExists` error as with this pull request.

I don't see how this PR changes durability. signac 2.1.0 already lazily loads the statepoint in many cases. Any invalid modifications to `signac_statepoint.json` would previously have been detected only when it is loaded.

janbridley (approved):
Looks good and seems to perform well! Thanks for this

tests/test_job.py
```diff
@@ -308,7 +308,7 @@ def test_interface_job_identity_change(self):
 
     def test_interface_nested_kws(self):
         with pytest.raises(InvalidKeyError):
-            job = self.open_job({"a.b.c": 0})
+            job = self.open_job({"a.b.c": 0}).statepoint
```
cbkerr, line 311: It's not very user friendly to only warn on statepoint access, but the tradeoff is worth it for the massive speedup

cbkerr (commented):
Great!

Discussing naming offline with Josh

tests/test_job.py
joaander: Users would be notified on `job.init()` before a numpy array is converted to the statepoint written to disk. 

When using a statepoint to open existing job, the id is calculated the same whether a given item is a numpy array or the equivalent list. In this case, the warning seems unnecessary.

signac/project.py
```diff
@@ -712,6 +719,20 @@ def find_jobs(self, filter=None):
 
         See :ref:`signac find <signac-cli-find>` for the command line equivalent.
 
+        .. tip::
```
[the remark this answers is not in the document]
joaander: What do you mean by "automatic"? This tip appears in the API docs: https://signac--975.org.readthedocs.build/projects/core/en/975/api.html#signac.job.Job.statepoint

signac/job.py
```diff
@@ -406,6 +414,32 @@ def update_statepoint(self, update, overwrite=False):
...
+    def statepoint_mapping(self):
+        """Get a copy of the job's statepoint as a read-only mapping.
+
+        `statepoint_mapping` uses the statepoint cache to provide fast access to the
+        job's statepoint for reading.
+
+        .. note::
+
+            Create and update the statepoint cache with ``signac update-cache`` on the
+            command line.
```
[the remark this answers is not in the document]
joaander: We can change this back if you think it important. I did make the change intentionally as `update_cache()` is not something that e.g. should be called in every flow operation or something like that. It should be called once when the user knows that the job workspace has been initialized. It needs to be called later only after jobs are added to the workspace.

The warning is only shown when users enable debug logging. I am considering updating the flow command line to issue a similar warning to nudge users to run `signac update-cache` when needed.

signac/project.py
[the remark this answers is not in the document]
cbkerr: I assumed bc I couldn't find aything for _job_dirs in the current docs

joaander:
This pull request breaks flow `aggregate.groupsof` (and possibly other aggregates). The jobs appear in random orders in the aggregates. I am investigating.

signac/job.py
```diff
@@ -406,6 +414,33 @@ def update_statepoint(self, update, overwrite=False):
         statepoint.update(update)
         self.statepoint = statepoint
 
+    @property
+    def cached_statepoint(self):
```
bdice, line 418: Must we add a new public API in order to provide the performance benefits of this PR? I am unsure if we should permit users to call this, or if it should be only leveraged internally as a private property `job._cached_statepoint`.

signac/project.py
```diff
@@ -1398,7 +1429,7 @@ def _remove_persistent_cache_file(self):
             if error.errno != errno.ENOENT:
                 raise error
 
-    def update_cache(self):
+    def update_cache(self, validate=True):
```
bdice, line 1432: If we can, I would like to keep users from having to think about "should I validate?" in any public APIs. When would we expect this to be called with `validate=False`? I see this is used in a single test below -- is that test needed? I haven't looked closely at what it's covering.

joaander: I added this only for the purposes of that test which assumes that `update_cache` will populate the cache will statepoint dict that does not match the key. I also don't like adding this option - I think the cache should always contain valid id: statepoint mappings.

I could remove this validate argument and also remove the test that it breaks.

joaander: This is the test that fails when `update_cache` validates entries:
https://github.com/glotzerlab/signac/blob/1de715530749018c334f8244a1d4cd25b61ef5a6/tests/test_project.py#L512-L516
The `job.init()` on line 516 fails because `update_cache` reads the corrupted statepoint and raises `JobsCorruptedError`. On `main`, `update_cache` inserts the incorrect statepoint into the cache.

joaander: I found a more localized fix to the test that does not require a public `validate` argument.

signac/job.py
bdice: Thanks for the explanation, that is helpful.

I would be open to changing `statepoint` semantics to be read-only in a future major version. We had discussed this at one point as a possibility for signac 2. Let's file an issue for that proposal.

signac/project.py
bdice: Very helpful, thanks!

bdice (approved):
Thanks for the hard work on this @joaander.

signac/job.py
joaander: Noted in #983.

signac/project.py
```diff
@@ -871,6 +903,9 @@ def _get_statepoint(self, job_id):
         ----------
         job_id : str
             A job id to get the state point for.
+        validate : bool
+            When True, validate that any statepoint read from disk matches the job_id.
```
cbkerr, line 907:
```suggestion
            When True, validate that any state point read from disk matches the job_id.
```

cbkerr (commented):
Adding changes to unify docstrings around state point being two words (https://docs.signac.io/en/latest/glossary.html#term-state-point).

PR #1017, opened 2025-01-28
Support `__getitem__` with `JobsCursor` objects

janbridley:
## Description
Implemented `__getitem__` for JobsCursors, allow for indexing and slicing after calling `project.find_jobs()`

## Motivation and Context
`JobsCursors` are iterable through the `_JobsCursorIterator` API, but slicing and indexing was not supported. This is atypical for Python types and requires users to unpack the cursor to index it.

This implementation returns a single `Job` if an integer index is requested, or a `_JobsCursorIterator` if a slice is requested. New tests have been added to cover the new functionality.

## Checklist:

- [ ] My name is on the [list of contributors](https://github.com/glotzerlab/signac/blob/main/contributors.yaml).

### DIFF
--- a/contributors.yaml
+++ b/contributors.yaml
@@ -142,4 +142,9 @@ contributors:
     family-names: Anderson
     given-names: "Joshua A."
     affiliation: "University of Michigan"
+  -
+    family-names: Bradley
+    given-names: Jenna
+    affiliation: "University of Michigan"
+    orcid: "https://orcid.org/0009-0007-2443-2982"
 ...
--- a/signac/project.py
+++ b/signac/project.py
@@ -1810,6 +1810,26 @@ class JobsCursor:
         # Code duplication here for improved performance.
         return _JobsCursorIterator(self._project, self._ids)
 
+    def __getitem__(self, index):
+        """Get an item or slice of items from the `self._ids` list.
+
+        Parameters
+        ----------
+        index : int | slice
+            The jobs to get.
+
+        Returns
+        -------
+        :class:`~signac.job.Job` | :class:`~signac.project._JobsCursorIterator`
+            Single job or iterator of a slice of jobs.
+        """
+        ids_to_return = self._ids[index]
+
+        if isinstance(ids_to_return, list):
+            return _JobsCursorIterator(self._project, ids_to_return)
+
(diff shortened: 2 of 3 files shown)


janbridley:
Note that docs are failing due to the ReadTheDocs configuration change - that will need to be addressed in a separate PR.

bdice:
This part of the code is usually benchmarked thoroughly before changes are accepted. @janbridley Would you be able to run the benchmark suite and share results? Here is a guide: https://github.com/glotzerlab/signac/blob/main/doc/support.rst#benchmarking

janbridley:
The `asv` suite does not seem to run either on my local machine (osx-arm) or a remote computer (Rocky Linux 8.10 x86). It could be a problem with `micromamba`, although I did try with a uv environment as well. If this is a known behavior please let me know - otherwise I can troubleshoot later this week.

```
· Creating environments
· Discovering benchmarks
·· Uninstalling from virtualenv-py3.12
·· Building 82a9c691 <main> for virtualenv-py3.12....
·· Installing 82a9c691 <main> into virtualenv-py3.12
·· Failed to build the project and import the benchmark suite.
```

bdice:
@janbridley I was able to reproduce the failure you're seeing. I tried `$ asv run main..feature/jobscursor-getitem -v` but the verbose output didn't help. I think this is the same as https://github.com/airspeed-velocity/asv/issues/1453.

bdice:
@janbridley I found that using `asv==0.6.1` works but newer versions do not work. I reported this upstream. https://github.com/airspeed-velocity/asv/issues/1453#issuecomment-2634786573

bdice:
@janbridley I had a bit of trouble interpreting the benchmarks. Were all the numbers present for both before and after? If you can confirm there is no performance change, I will approve.

janbridley:
> @janbridley I had a bit of trouble interpreting the benchmarks. Were all the numbers present for both before and after? If you can confirm there is no performance change, I will approve.

I updated the benchmark results - I'm not sure why the previous output tested the same commit twice. All large benchmarks (N=1000) are within standard deviation of main

bdice (approved):

bdice:
@janbridley Would you be able to fix the Sphinx error here?

PR #1030, opened 2025-04-18
Fix schema on state point that is list of dicts that contain lists

cbkerr:
I had a state point containing "dislocations" specified as a list of dictionaries each containing location and Burgers vectors which failed on schema detection.

I am not sure if this case means that we missed other complex nested state points.

## Description

The fix is to also recursively call `_to_hashable` when creating a `_hashable_dict`.

I added a test for the schema case that fails before my fix. I added a test in `test_job` that doesn't catch anything from before.

## Motivation and Context

This code used to fail with `TypeError: unhashable type: 'list'`

```python
import signac

project = signac.init_project()

minimal = {"a": [{"a": []}]}
job = project.open_job(minimal).init()
print(job, job.sp) # works up to here

schem = project.detect_schema()
```

## Checklist:


signac/_utility.py
```diff
@@ -122,7 +122,8 @@ def _dotted_dict_to_nested_dicts(dotted_dict, delimiter_nested="."):
 
 class _hashable_dict(dict):
     def __hash__(self):
-        return hash(tuple(sorted(self.items())))
+        hashed = {_to_hashable(k): _to_hashable(v) for k, v in self.items()}
```
cbkerr, line 125: I'm not totally sure if we need to hash the key, because we already put restrictions on state point keys

bdice: Python’s dict implementation requires that dict keys must be hashable. We can skip making the keys hashable.

Perhaps this can be a one-liner (may require reformatting):
```python
return hash(tuple(sorted((k, _to_hashable(v)) for k, v in self.items())))
```

tests/test_job.py
```diff
@@ -80,6 +82,12 @@ def nested_dict(self):
         d["g"] = builtins_dict()
         return d
 
+    @classmethod
+    def list_of_dict(self):
```
bdice, line 86: Maybe rename?
```suggestion
    def dict_of_list_of_dict(self):
```

bdice (commented):
Nice catch. I’m surprised we haven’t run across this before. I have a couple suggestions.

changelog.txt
```diff
@@ -7,6 +7,14 @@ The **signac** package follows `semantic versioning <https://semver.org/>`_.
 Version 2
 =========
 
+[2.4.0] -- 2025-xx-xx
+---------------------
+
+Fixed
++++++
+
+ - Schema detection when a state point was a list of dictionaries having lists (#1030).
```
bdice, line 16: A state point must be a dict at the root. Do you mean “a dictionary of lists containing dictionaries (possibly of lists)?”

bdice (approved):
Question on the changelog, otherwise LGTM.

changelog.txt
cbkerr, line 16:
```suggestion
 - Schema detection when a state point value is a list of dictionaries having lists (#1030).
```

bdice (approved):

bdice:
Hmm. I wonder why the automerge isn't working. Is there an unexpected branch protection on `main`? 
image

PR #1121, opened 2026-07-02
Don't trigger neighbor list warnings on ignored constant keys that are actually present

cbkerr:
## Description

Fixes an errant warning that would appear when ignoring a constant parameter.
Neighbor list calculation filtered out constant state point parameters, but then purposefully ignoring such a parameter would trigger a warning that the ignored key was not present. This adds tests for the structure of the neighbor list while fixing this. 

## Motivation and Context

Discovered while working on https://github.com/glotzerlab/signac-dashboard/pull/320

## Checklist:

- [x] The [package documentation](https://github.com/glotzerlab/signac/tree/main/doc) and [framework documentation](https://signac.readthedocs.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.

### DIFF
--- a/signac/project.py
+++ b/signac/project.py
@@ -1725,7 +1725,8 @@ class Project:
         if not isinstance(ignore, list):
             ignore = [ignore]
 
-        sorted_schema = self._flat_schema(exclude_const=True)
+        sorted_schema = self._flat_schema(exclude_const=False)
+
         need_to_ignore = [sorted_schema.pop(ig, _DictPlaceholder) for ig in ignore]
         if any(is_bad_key := [a is _DictPlaceholder for a in need_to_ignore]):
             bad_keys = list(compress(ignore, is_bad_key))
@@ -1736,6 +1737,10 @@ class Project:
             )
             for bad_key in bad_keys:
                 ignore.remove(bad_key)
+        # now take out constant keys
+        const_keys = [key for key,val in sorted_schema.items() if len(val)==1]
+        for key in const_keys:
+            sorted_schema.pop(key)
 
         self.update_cache()
         # pass a copy of cache
--- a/signac/project.py
+++ b/signac/project.py
@@ -1738,7 +1738,7 @@ class Project:
             for bad_key in bad_keys:
                 ignore.remove(bad_key)
         # now take out constant keys
-        const_keys = [key for key,val in sorted_schema.items() if len(val)==1]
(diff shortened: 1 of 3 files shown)


signac/project.py
```diff
@@ -1736,6 +1737,10 @@ def get_neighbors(self, ignore=[]):
             )
             for bad_key in bad_keys:
                 ignore.remove(bad_key)
+        # now take out constant keys
+        const_keys = [key for key, val in sorted_schema.items() if len(val) == 1]
+        for key in const_keys:
+            sorted_schema.pop(key)
```
bdice: Maybe microbenchmark these snippets against each other.

```suggestion
        sorted_schema = {
            key: val
            for key, val in sorted_schema.items()
            if len(val) != 1
        }
```

If you don't like the above, try using a `set` instead of a `list` and use `del` instead of `pop`.
```suggestion
        const_keys = {key for key, val in sorted_schema.items() if len(val) == 1}
        for key in const_keys:
            del sorted_schema[key]
```

tests/test_neighborlist.py
```diff
@@ -87,6 +89,139 @@ def test_neighbors_ignore(self):
...
+    def test_neighbors_no_neighbors(self):
+        job = self.project.open_job({"constant": 1, "constant_2": 2}).init()
+        neighbor_list = self.project.get_neighbors()
+        assert neighbor_list[job.id] == {}
+
+        neighbor_list = self.project.get_neighbors(ignore="constant")
+        assert neighbor_list[job.id] == {}
+
+    def test_neighborlist_structure(self):
+
```
bdice:
```suggestion
```

tests/test_neighborlist.py
```diff
@@ -87,6 +89,139 @@ def test_neighbors_ignore(self):
...
+                    if neighbor_job_id == "47c6f9a7c0d88ab60a9e8de4ed002e6f":
+                        # that's job with b:None
+                        print(f"ignoring {neighbor_value=} and {neighbor_job_id=}")
+                    else:
+                        assert neighbor_value is not None
+                    assert neighbor_job_id
+                    assert key
+
+    def test_neighborlist_structure_with_constant(self):
+
```
bdice:
```suggestion
```

bdice (approved):
Approving with a couple ideas. I don't need to review further.

PR #1122, opened 2026-07-13
Fix update_cache when number of jobs changes

cbkerr:
`Project.update_cache()` doesn't update the cache if jobs were added and then project goes out of scope (eg, CLI use).

## Description

The problem was that `_update_in_memory_cache()` was called too late inside `update_cache`.

I removed side effects of the internal `Project._read_cache()` to make it more clear throughout the code when it was being used to update `Project._sp_cache`. Now this function only returns what's truly in the file sp cache.

~~Since the sp cache is read when updating the cache, it can only grow. This follows the _current_ design where the cache is a superset of jobs in the project to allow for repairing corrupted jobs or accessing recently deleted jobs.~~

~~Therefore, I added the `prune` argument to allow the cache to reduce in size to match the workspace. It defaults to False.~~

## Motivation and Context

Current (incorrect) behavior
```bash
signac init
# Initialized project.

signac job --create '{"a": 0}'
# 9bfd29df07674bc4aa960cf661b5acd2

signac update-cache
# Updating cache...
# Updated cache (size=1).

gunzip -c .signac/statepoint_cache.json.gz
# {"9bfd29df07674bc4aa960cf661b5acd2": {"a": 0}}

signac job --create '{"a": 1}'
# 42b7b4f2921788ea14dac5566e6f06d0

signac update-cache
# Updating cache...
# Cache is up to date.

gunzip -c .signac/statepoint_cache.json.gz
# {"9bfd29df07674bc4aa960cf661b5acd2": {"a": 0}} # WRONG

# have to delete the cache and recreate it to get the update
rm .signac/statepoint_cache.json.gz
signac update-cache
# Updating cache...
# Updated cache (size=2).

gunzip -c .signac/statepoint_cache.json.gz
# {"42b7b4f2921788ea14dac5566e6f06d0": {"a": 1}, "9bfd29df07674bc4aa960cf661b5acd2": {"a": 0}}
```

It's a bit more involved to recreate in python because you have to recreate the project object.
```python
import gzip
import json
import signac
import os

def manual_read_cache_file(project):
    with gzip.open(project.fn(project.FN_CACHE), "rb") as cachefile:
        cache = json.loads(cachefile.read().decode())
    return cache

num_initial = 3
num_total = 6

project = signac.init_project()

# protects against confusion when running this test multiple times
assert len(project) == 0

for a in range(num_initial):
    project.open_job({"a": a}).init()

project.update_cache() # writes the sp cache to disk

file_cache_1 = manual_read_cache_file(project)
assert file_cache_1 == project._sp_cache
assert len(file_cache_1) == num_initial

# add some jobs
for a in range(num_initial, num_total):
    project.open_job({"a": a}).init()

# let session end or make the project object again
del project

project.update_cache()
file_cache_2 = manual_read_cache_file(project)
# these fail:
assert file_cache_2 == project._sp_cache
assert len(file_cache_2) == num_total
```

## This PR
```bash
signac init
# Initialized project.

signac job --create '{"a": 0}'
# 9bfd29df07674bc4aa960cf661b5acd2

signac update-cache
# Updating cache...
# Updated cache (size=1).

gunzip -c .signac/statepoint_cache.json.gz
# {"9bfd29df07674bc4aa960cf661b5acd2": {"a": 0}}

signac job --create '{"a": 1}'
# 42b7b4f2921788ea14dac5566e6f06d0

signac update-cache
# Updating cache...
# Updated cache (size=2).

gunzip -c .signac/statepoint_cache.json.gz
# {"9bfd29df07674bc4aa960cf661b5acd2": {"a": 0}, "42b7b4f2921788ea14dac5566e6f06d0": {"a": 1}}

# remove a job directory manually (or use signac's methods to remove a job)
rm -r workspace/42b7b4f2921788ea14dac5566e6f06d0/

# cache matches jobs in workspace
signac update-cache
# Updating cache...
# Updated cache (size=1).

gunzip -c .signac/statepoint_cache.json.gz
{"9bfd29df07674bc4aa960cf661b5acd2": {"a": 0}}

```

## Checklist:

- [ ] The [package documentation](https://github.com/glotzerlab/signac/tree/main/doc) and [framework documentation](https://signac.readthedocs.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.


signac/project.py
```diff
@@ -1435,21 +1437,38 @@ def _remove_persistent_cache_file(self):
             if error.errno != errno.ENOENT:
                 raise error
 
-    def update_cache(self):
-        """Update the persistent state point cache.
+    # TODO: change name to write_cache to better capture the meaning of this function?
+    def update_cache(self, prune=False):
```
janbridley: Agreed that write_cache is more clear. Could it even be write_disk_cache?

tests/test_project.py
```diff
@@ -2150,6 +2150,77 @@ def test_repr(self):
...
+        del self.project
+        self.project = self.project_class.get_project(path=self._tmp_pr)
+
+        self.project.update_cache()
+        file_cache_2 = self.manual_read_cache_file()
+        project_mem_cache_2 = self.project._sp_cache
+        project_read_cache_2 = self.project._read_cache()
+        assert file_cache_2 == project_mem_cache_2
+        assert file_cache_2 == project_read_cache_2
+        assert len(file_cache_2) == num_total
```
janbridley: Haven't used this myself, but pytest-check mirrors a pattern we use in rust for multiple assertions https://github.com/okken/pytest-check

janbridley (approved):
Approved overall, but I made a few notes on improvements!

signac/project.py
cbkerr: Yes! And perhaps even generalize to write additional objects to cache like a search index. Since this would be breaking I'll leave this note in to consider in v3

tests/test_project.py
cbkerr: I added more pytest subtests, which are built in and achieve a similar thing

cbkerr:
After discussing with Josh, I decided to remove `prune` as an option and always "prune" the cache. This will make it easier in the future to use the state point disk cache as a record of the job directories in the workspace (if we also keep track of the latest mtime along with the cache).

